PR #22, opened 2025-12-02
blacked

ichumuh:


LucaKro:
why not do it for the whole repository?

PR #45, opened 2025-12-04
Hsrb

Narenvasant:
changed urdf to work with the meshes in hsr descriptions and also added joint states for hsr


pycram/src/pycram/robot_descriptions/hsrb_states.py
```diff
@@ -0,0 +1,73 @@
...
+    ],
+    joint_positions=[0.0, 1.5, -1.85, 0.0],
+    state_type=StaticJointState.Park,
+    arm=Arms.LEFT,
+)
+
+left_gripper_open = GripperState(
+    name=PrefixedName("hsrb", "left_gripper_open"),
+    joint_names=["hand_l_proximal_joint", "hand_r_proximal_joint", "hand_motor_joint"],
+    joint_positions=[0.3, 0.3, 0.3],
```
ichumuh: @Tigul how does pycram work interally? cause this should be mimic joints if i'm not mistaken, meaning you only need to set one of them, prob hand_motor_joint.

Tigul: But the whole robot states thing is only an intermediate solution until I merge this into the AbstractRobot annotations

PR #53, opened 2025-12-06
Cleaning

AbdelrhmanBassiouny:
- Removed `HashedValue` completely.
- Refactored `HashedIterable`, renamed it to `ReEnterableLazyIterable`, and moved it to cache_data.py.
- Removed unused methods (`_is_duplicate_output_`, `_projection_`)
- Removed unused classes (`Infer`, `From`)
- Reduced ReEnterableLazyIterable implementation to the bare minimum and confined its use to `_domain_` of `Variable`.

### DIFF
--- a/krrood/src/krrood/entity_query_language/cache_data.py
+++ b/krrood/src/krrood/entity_query_language/cache_data.py
@@ -1,7 +1,5 @@
 from __future__ import annotations
 
-from typing import Generic, Iterable, List
-
 from .utils import T
 
 """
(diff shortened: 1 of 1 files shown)


krrood/src/krrood/entity_query_language/cache_data.py
```diff
@@ -82,3 +86,43 @@ def clear(self):
...
+    materialized_values: List[T] = field(default_factory=list)
+    """
+    The materialized values of the iterable.
+    """
+
+    def set_iterable(self, iterable):
+        """
+        Set the iterable to wrap.
+        """
+        self.iterable = (v for v in iterable)
```
LucaKro: why do we need that?

AbdelrhmanBassiouny: I added a clarification to the method's doc now.

PR #62, opened 2025-12-09
Ported CartesianPositionVelocityLimit & CartesianRotationVelocityLimit

MarvinDufner:
Ported CartesianPositionVelocityLimit and CartesianRotationVelocityLimit to new API. Also added tests. Basically the same thing as #52 split into two classes

### DIFF
--- a/giskardpy/src/giskardpy/motion_statechart/tasks/cartesian_tasks.py
+++ b/giskardpy/src/giskardpy/motion_statechart/tasks/cartesian_tasks.py
@@ -320,10 +320,10 @@ class CartesianPositionVelocityLimit(Task):
     that evaluates whether the current translational speed is within the limit.
 
     .. warning::
-       Strict Cartesian velocity limits typically require many constraints to be
-       added to the optimization problem and can substantially slow down the
-       system. If runtime or responsiveness is important,
-       prefer a larger limit or a softer (lower-weight) constraint
+       Strict Cartesian velocity limits require as many constraints as the prediction
+       horizon size, making the optimization problem more complex. This can impact
+       solve time especially at high control frequencies. If computation time is critical,
+       consider using larger limits or reducing the prediction horizon.
     """
 
     root_link: KinematicStructureEntity = field(kw_only=True)
@@ -336,8 +336,8 @@ class CartesianPositionVelocityLimit(Task):
     tip-frame translational velocity does not exceed this value."""
     weight: float = field(default=DefaultWeights.WEIGHT_ABOVE_CA, kw_only=True)
     """Optimization weight determining how strongly the linear velocity
-    limit is enforced. Larger values increase enforcement priority at the
-    cost of potentially making the optimization problem harder to solve."""
+    limit is enforced. Higher weights give this constraint soft priority
+    over lower weighted constraints when conflicts occur."""
 
     def build(self, context: BuildContext) -> NodeArtifacts:
         artifacts = NodeArtifacts()
@@ -374,10 +374,10 @@ class CartesianRotationVelocityLimit(Task):
     compliance with the defined angular velocity limits.
 
     .. warning::
-       Strict Cartesian velocity limits typically require many constraints to be
-       added to the optimization problem and can substantially slow down the
-       system. If runtime or responsiveness is important,
-       prefer a larger limit or a softer (lower-weight) constraint
+       Strict Cartesian velocity limits require as many constraints as the prediction
+       horizon size, making the optimization problem more complex. This can impact
+       solve time especially at high control frequencies. If computation time is critical,
+       consider using larger limits or reducing the prediction horizon.
     """
 
     root_link: KinematicStructureEntity = field(kw_only=True)
@@ -390,8 +390,8 @@ class CartesianRotationVelocityLimit(Task):
     rotation rate does not exceed this threshold."""
     weight: float = field(default=DefaultWeights.WEIGHT_ABOVE_CA, kw_only=True)
     """Optimization weight determining how strongly the rotational velocity
-    limit is enforced. Larger values increase enforcement priority at the
-    cost of potentially making the optimization problem harder to solve."""
+    limit is enforced. Higher weights give this constraint soft priority
+    over lower weighted constraints when conflicts occur."""
 
     def build(self, context: BuildContext) -> NodeArtifacts:
         artifacts = NodeArtifacts()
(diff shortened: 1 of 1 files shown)


ichumuh (changes requested):

giskardpy/src/giskardpy/motion_statechart/tasks/cartesian_tasks.py
```diff
@@ -306,57 +306,82 @@ def on_start(self, context: ExecutionContext):
...
-        :param weight: default DefaultWeights.WEIGHT_ABOVE_CA
-        :param hard: Turn this into a hard constraint. This make create unsolvable optimization problems
-        """
-        r_P_c = context.world.compose_forward_kinematics_expression(
+    root_link: KinematicStructureEntity = field(kw_only=True)
+    """root link of the kinematic chain"""
+    tip_link: KinematicStructureEntity = field(kw_only=True)
+    """tip link of the kinematic chain"""
+    max_linear_velocity: float = field(default=0.1, kw_only=True)
+    """in m/s"""
```
tomsch420: dont be lazy here. Write concrete sentences

giskardpy/src/giskardpy/motion_statechart/tasks/cartesian_tasks.py
```diff
@@ -306,57 +306,82 @@ def on_start(self, context: ExecutionContext):
...
+    weight: float = field(default=DefaultWeights.WEIGHT_ABOVE_CA, kw_only=True)
```
tomsch420: missing doc

giskardpy/src/giskardpy/motion_statechart/tasks/cartesian_tasks.py
```diff
@@ -306,57 +306,82 @@ def on_start(self, context: ExecutionContext):
...
+
 
 @dataclass
 class CartesianRotationVelocityLimit(Task):
-    root_link: Body = field(kw_only=True)
-    tip_link: Body = field(kw_only=True)
-    weight: float = DefaultWeights.WEIGHT_ABOVE_CA
-    max_velocity: Optional[float] = None
+    """
+    This goal will use put a strict limit on the Cartesian velocity. This will require a lot of constraints, thus
```
tomsch420: looks like there is a common superclass to extract here

tomsch420: the fields are the exact same tho

giskardpy/src/giskardpy/motion_statechart/tasks/cartesian_tasks.py
```diff
@@ -343,20 +358,37 @@ def build(self, context: BuildContext) -> NodeArtifacts:
...
+       system. If runtime or responsiveness is important,
+       prefer a larger limit or a softer (lower-weight) constraint
     """
 
     root_link: KinematicStructureEntity = field(kw_only=True)
-    """root link of the kinematic chain"""
+    """Root link of the kinematic chain. Defines the reference frame from which the tip's motion is measured."""
     tip_link: KinematicStructureEntity = field(kw_only=True)
-    """tip link of the kinematic chain"""
+    """Tip link of the kinematic chain. The translational velocity of this link (expressed in the root link frame) is constrained."""
```
tomsch420: still root link, tip link and weight are duplicated. Where superclass

tomsch420 (changes requested):

giskardpy/src/giskardpy/motion_statechart/tasks/cartesian_tasks.py
ichumuh: oh that's what you mean. yeah translation and rotation velocity share the same fields, but that's more of a coincidence, the implementation is quite different. don't think it saves much code to collect that into a common super class. i would say we keep these two separate, unless more similar cases pop up

ichumuh: implementing the general cart velocity limits as a parallel of the two others makes more sense, this reduced duplication of complex code

giskardpy/src/giskardpy/motion_statechart/tasks/cartesian_tasks.py
```diff
@@ -306,114 +309,171 @@ def on_start(self, context: ExecutionContext):
...
+    This goal enforces a strict cap on the linear speed of the frame defined by
+    the kinematic transform from `root_link` to `tip_link`. Enforcement is performed
+    by adding constraints to the optimizer and by providing an observation expression
+    that evaluates whether the current translational speed is within the limit.
+
+    .. warning::
+       Strict Cartesian velocity limits typically require many constraints to be
+       added to the optimization problem and can substantially slow down the
+       system. If runtime or responsiveness is important,
+       prefer a larger limit or a softer (lower-weight) constraint
```
ichumuh: The last sentence is not true. I'm not talking about the slowdown due to slower motions. 

The optimization problem is getting more complex and therefore needs more time to solve. this may become a problem if Giskard is running at a very high frequency. A "normal" positional constraint, like we use for cartesian position only needs a single constraint in the underlying QP. But a velocity constraint requires as many constraints as our prediction horizon is large. so if the prediction horizon is 15, we need 15 constraints to properly enforce a velocity limit.

giskardpy/src/giskardpy/motion_statechart/tasks/cartesian_tasks.py
```diff
@@ -306,114 +309,171 @@ def on_start(self, context: ExecutionContext):
...
+    optimization weights to prioritize its enforcement in solving problems
+    involving kinematic motion. The task calculates and enforces constraints
+    based on the rotation matrix between the root and tip links, ensuring
+    compliance with the defined angular velocity limits.
+
+    .. warning::
+       Strict Cartesian velocity limits typically require many constraints to be
+       added to the optimization problem and can substantially slow down the
+       system. If runtime or responsiveness is important,
+       prefer a larger limit or a softer (lower-weight) constraint
```
ichumuh: same as above

giskardpy/src/giskardpy/motion_statechart/tasks/cartesian_tasks.py
```diff
@@ -306,114 +309,171 @@ def on_start(self, context: ExecutionContext):
...
+    """Tip link of the kinematic chain. The translational velocity of this link (expressed in the root link frame) is constrained."""
     max_angular_velocity: float = field(default=0.5, kw_only=True)
-    """in rad/s"""
+    """Maximum allowed angular speed. Interpreted in radians per second (rad/s).
+    The enforcement ensures the magnitude of the instantaneous
+    rotation rate does not exceed this threshold."""
     weight: float = field(default=DefaultWeights.WEIGHT_ABOVE_CA, kw_only=True)
+    """Optimization weight determining how strongly the rotational velocity
+    limit is enforced. Larger values increase enforcement priority at the
+    cost of potentially making the optimization problem harder to solve."""
```
ichumuh: same as above

ichumuh (changes requested):

giskardpy/src/giskardpy/motion_statechart/tasks/cartesian_tasks.py
ichumuh: yes

Issue #30, opened 2025-12-02
Match queries don't support complex queries.


AbdelrhmanBassiouny:
Currently one cannot express inequalities, predicates, self referencing, and others using a match query.

This can be resolved if we add a where statement to support that after the match query. So something like this:

```python
authorship = match(Authorship)(author=match(Author)(name="Bass"))
select(authorship).where(authorship.author_number > 3)
```

AbdelrhmanBassiouny:
> what would this look like with the example from [#29](https://github.com/cram2/cognitive_robot_abstract_machine/issues/29)

assuming we are using the proposed solution of #29, and we want a professor who teaches more than 5 course, it will look like this:
```python
    professor = a(
        match(Professor)(
            works_for=match(Organization)(uri="http://www.Department0.University0.edu")
        )
    )
    query = select(
        professor,
        name := professor.name,
        email := professor.person.email_address,
        telephone:= professor.person.telephone
    ).where(count(professor.teaches_courses) > 5)
```

One thing though I don't know where should the quantifier be put, currently it is on the match satement, but now once we add a where to the select, it is inconsistent because now there is unquantified conditions added through the where

AbdelrhmanBassiouny:
and the where cannot be used with the match directly as the professor is not defined yet to be used in the where, so I cannot do that:
```python
professor = a(
        match(Professor)(
            works_for=match(Organization)(uri="http://www.Department0.University0.edu")
        ).where(count(professor.teaches_courses) > 5)
    )
```

LucaKro:
im pretty sure that how we talked about it back when eql started as well, that entity is essentially select, and a/an/the are essentially the quantifiers one or all

AbdelrhmanBassiouny:
yes they are like that true, and I also feel the same as you, the thing is currently entity accepts a Variable and the match currently is not a varaible, it is not evaluatable, it creates the query that is evaluatable.

So for the example here `professor` is not a Selectable, which is the type that entity accepts for selection.

In the current version of match it already returns `entity()` or `set_of()` expression that is why you can do `a(match)`

But with the separation of selection from match (what #29  proposes), it should now only return conditions. Maybe there could be a refactor that allows match expressions to fit as selectables in entity such that one can write:
```python
professor = match(...)
an(entity(professor))
```

LucaKro:
[...]

Issue #29, opened 2025-12-02
New Match Query Structure


AbdelrhmanBassiouny:
Current match queries have the problem of mixing responsibilities. I handles both pattern matching and selection which is unintuitive, and makes the query ugly. In addition it violates single responsibility principle.

Example of finding a Professor that works for an Organization with a given uri:

```python
    professor, name, email, telephone = select(Professor), select(), select(), select()
    query = a(
        professor(
            works_for=match(Organization)(uri="http://www.Department0.University0.edu"),
            name=name,
            person=match()(
                email_address=email,
                telephone=telephone,
            ),
        )
    )
```

The proposed solution should look like this:

```python
    professor = a(
        match(Professor)(
            works_for=match(Organization)(uri="http://www.Department0.University0.edu")
        )
    )
    query = select(
        professor,
        name := professor.name,
        email := professor.person.email_address,
        telephone:= professor.person.telephone
    )
```

sorinar329:
I mean, SPARQL queries also use select ...

PR #67, opened 2025-12-10
[KRROOD] Refactor of Query Matching API  + New Features.

AbdelrhmanBassiouny:
## Update:
Also check #71, as either it or this will be merged

Addresses #29 and #30 
This pull request refactors the EQL (Entity Query Language) pattern matching API, replacing the `match` and `entity_matching` functions with a more unified and flexible `matching` API. The changes update both the implementation and documentation, and revise tests to use the new API. Additional improvements include expanding type support in quantification functions and refining internal symbolic handling.

Adding a domain to a match statement is now like this:

 ```python
matching().from_(domain)
```

Now there is a `select(*variables).where(*conditions)` statement for easier and cleaner selection mechanics with match statements, in addition to the ability to use more sophisticated conditions by making use of the `where()` statement.

Example:
```python
    fixed_connection = the(matching(FixedConnection)(
        parent=matching(Container),
        child=matching(Handle)(name="Handle3"),
    ).from_(world.connections))
    container_and_handle = select(container := fixed_connection.parent,
                                  handle := fixed_connection.child).where(container.size > 1)
```

Using core entity API would be like this:
```python
    fixed_connection = let(FixedConnection, domain=world.connections)
    container_and_handle = the(
        set_of((container := fixed_connection.parent, handle := fixed_connection.child),
               HasType(container, Container),
               HasType(handle, Handle),
               handle.name == "Handle3",
               container.size > 1
               ))
```

## Update:
Now all result processors, distinct, and order_by also work with match.
Updated documentation.
used `from_` instead of `domain_from` as it reads better.


krrood/src/krrood/entity_query_language/predicate.py
```diff
@@ -4,6 +4,7 @@
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
 from functools import wraps, lru_cache
+from typing import Iterable
```
tomsch420: extensions

krrood/src/krrood/entity_query_language/predicate.py
```diff
@@ -67,6 +68,9 @@ def __new__(cls, *args, **kwargs):
         update_cache(instance)
         return instance
 
+    def domain_from(self, domain: Iterable[T]) -> T:
+        ...
```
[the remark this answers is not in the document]
AbdelrhmanBassiouny: oh that's a hack for hints from the IDE, that should not be implemented

krrood/src/krrood/entity_query_language/entity_result_processors.py
```diff
@@ -0,0 +1,123 @@
...
+from .match import Match, SelectableMatchExpression
+from .result_quantification_constraint import (
+    ResultQuantificationConstraint,
+)
+from .symbolic import An, The, SetOf, Entity, Selectable, Max, Min, Sum, Average, Count, ResultProcessor
+from .utils import T
+
+
+def an(
+        entity_: Union[SetOf[T], Entity[T], T, Iterable[T], Type[T], Match[T]],
```
LucaKro: hahaha i still die inside when looking at these types hahaha

Image

krrood/src/krrood/entity_query_language/match.py
```diff
@@ -114,30 +330,34 @@ def _resolve(
         :param parent: The parent match if this is a nested match.
         :return:
         """
-        self._update_fields(variable, parent)
+        if self.resolved:
+            return
+        self.update_fields(variable, parent)
         for attr_name, attr_assigned_value in self.kwargs.items():
```
LucaKro: please avoid abbreviations if possible

LucaKro (changes requested):

AbdelrhmanBassiouny:
Replace with #75

PR #75, opened 2025-12-14
Make entity great again

### LINKED ISSUE #29: New Match Query Structure

AbdelrhmanBassiouny:
## Important:

- This breaks APIs
- This Resolves #29 and #30

## Summary:

- Rename `let` to `variable` as it is clearer that it creates a variable. While `let` gives the impression of assigning a value.
- Replace `entity(variable, *conditions)` with `entity(variable).where(*conditions)` to clearly separate selection and conditions.
- Similarly replace `set_of((variable1, variable2, ...), *conditions)`  with `set_of(variable1, variable2, ...).where(*conditions)`
- Rename `matching`  to `match_var(type, domain)` for the outermost match that takes a domain similar to `variable(type, domain`.
- Inner matches are called `match` and do not take a domain.
- The match functionality and scope have been reduced to just equality constraints, and they are meant to be a fancier `variable` that can take keyword argument constraints for easy and clear filtering.

## Examples:

This is how a query that uses match looks:
```python
fixed_connection = match_variable(FixedConnection, domain=world.connections)(
        parent=match(Container)(name="Container1"),
        child=match(Handle)(name="Handle1")
    )
fixed_connection_query = the(entity(fixed_connection))
```

This is the equivalent query without match:
```python
fixed_connection = variable(FixedConnection, domain=world.connections)
fixed_connection_query = the(
    entity(fixed_connection).where(
        HasType(fixed_connection.parent, Container),
        HasType(fixed_connection.child, Handle),
        fixed_connection.parent.name == "Container1",
        fixed_connection.child.name == "Handle1",
    )
)
```


krrood/doc/eql/cache.md
```diff
@@ -16,7 +16,7 @@ kernelspec:
 
 EQL provides a cache for all symbol classes.
 Whenever python creates an object of a class that inherits from `Symbol`, EQL will cache the instance in the symbol graph.
-When you create a free variable via `let` without a domain, EQL will use the `Symbol Graph` to extract the domain.
+When you create a free variable via `var` without a domain, EQL will use the `Symbol Graph` to extract the domain.
```
LucaKro: call it "variable", not "var". avoid abbreviations where possible

krrood/src/krrood/entity_query_language/rxnode.py
```diff
@@ -82,9 +84,10 @@ def parent(self) -> Optional["RWXNode"]:
     @parent.setter
     def parent(self, value: Optional["RWXNode"]):
         if value is None:
```
LucaKro: avoid nesting if statements by inverting this if statement, then doing
```Python
        # Create edge and set as primary (no need to detach non-primary edges)
        self.add_parent(value)
        self._primary_parent_id = value.id
        return
```
to return early

krrood/src/krrood/entity_query_language/rxnode.py
```diff
@@ -102,7 +105,9 @@ def descendants(self) -> List["RWXNode"]:
 
     @property
     def leaves(self) -> List["RWXNode"]:
```
LucaKro: Why is RWXNode in ""?

krrood/src/krrood/entity_query_language/rxnode.py
```diff
@@ -114,14 +119,25 @@ def root(self) -> "RWXNode":
     def __str__(self):
         return self.name
 
-    def visualize(self, figsize=(35, 30), node_size=7000, font_size=25, spacing_x: float = 4, spacing_y: float = 4,
-                  curve_scale: float = 0.5, layout: str = 'tidy', edge_style: str = 'orthogonal',
-                  label_max_chars_per_line: Optional[int] = 13):
+    def visualize(
+        self,
+        figsize=(35, 30),
```
LucaKro: docstring for params

krrood/src/krrood/entity_query_language/match.py
```diff
@@ -147,135 +216,125 @@ def _update_fields(
...
+    @cached_property
+    def expression(self) -> Union[CanBehaveLikeAVariable[T], T]:
+        """
+        Return the entity expression corresponding to the match query.
+        """
+        if not self.variable:
+            self.resolve()
+        return self.variable
+
+    def _resolve(
```
LucaKro: reduce nesting in this method by inverting the if statement

krrood/src/krrood/entity_query_language/symbolic.py
LucaKro: a lot of uncomplete or missing docstrings here

LucaKro (changes requested):

krrood/doc/eql/cache.md
tomsch420: i agree

krrood/src/krrood/entity_query_language/match.py
```diff
@@ -147,135 +216,125 @@ def _update_fields(
...
     """
 
-    attr_name: str
+    parent: AbstractMatchExpression = field(kw_only=True)
     """
-    The name of the attribute to assign the value to.
+    The parent match expression.
     """
-    variable: CanBehaveLikeAVariable
+    attr_name: str = field(kw_only=True)
```
tomsch420: 🚨  abbrevation spotted 🚨

krrood/src/krrood/entity_query_language/predicate.py
```diff
@@ -67,6 +68,9 @@ def __new__(cls, *args, **kwargs):
         update_cache(instance)
         return instance
 
+    def from_(self, domain: Iterable[T]) -> T:
+        raise NotImplementedError("This method should not be called, this is just for IDE hints")
```
tomsch420: i think its not the IDE itself, its the static type checker "Flow"

krrood/src/krrood/entity_query_language/rxnode.py
```diff
@@ -114,14 +119,25 @@ def root(self) -> "RWXNode":
...
+        layout: str = "tidy",
+        edge_style: str = "orthogonal",
+        label_max_chars_per_line: Optional[int] = 13,
+    ):
         """Render a rooted, top-to-bottom directed graph.
         Delegates to a dedicated visualizer class to keep this method small and reusable.
         """
         if not GraphVisualizer:
-            raise RuntimeError("rustworkx_utils is not installed. Please install it with `pip install rustworkx_utils`")
+            raise RuntimeError(
```
tomsch420: ModuleNotFoundError

tomsch420 (changes requested):

krrood/src/krrood/entity_query_language/predicate.py
LucaKro: hmm i missed that, but also this looks pretty smelly to me...but so does most typing stuff in krrood tbf :sweat_smile:

AbdelrhmanBassiouny: will remove it that is not used anymore

PR #78, opened 2025-12-16
[SemDT] Semantic Annotation Rework

LucaKro:
# Feature: Factory Rethink

This PR refactors the creation of semantic annotations by moving factory logic into mixins and the annotation classes themselves, eliminating the need for a separate `factories.py` module. This makes the creation process more object-oriented and reduces code duplication. Additionally, it updates the Procthor adapters to align with these changes.

The majority of the changes were done in
- semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py 
- semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/semantic_annotations.py

### Motivation

This PR was sparked because SUTURO made me realize that beyond a few cases, semantic annotation factories were not really practical to use, as they require a good understanding of the code to even begin to use them. Furthermore would it require either a refactor or duplicated code to create a factory for a fridge, with the dresser factory already existing, since they are virtually the same in terms of abtract geometry (both are essentially a case and may have drawers and doors).
In addition to that, we needed to be able to create semantic annotations without having complete information. for example initialize a door, without knowing its hinge or handle. this was not possible which caused some awkward situations when we didnt have all the required information.

In this new version, basically all fields of semantic annotations are optional, and instead of having one factory that takes all required arguments for example to build a dresser, there is a factory method called Dresser.create_with_new_body_in_world that creates the minimal geometry required for the dresser to exist, namely, its main body, which is *always* mounted onto the worlds root. if we now perceive a drawer and want to add it to the dresser, we first create the drawer in the world using the same method, and then call Dresser.add_drawer, which then modifies the kinematic structure and mounts the drawer as a child of the dresser body. 
Important to know is that at this point, we dont make any assumptions about the kinematic capabilities yet, the drawer is just mounted using a fixed connection. but when we somehow (and we dont specify how that is determined) the kinematic capabilities, then we can, for example, use Slider.create_with_new_body_in_world to create the joint with the appropriate parameters, and then use Drawer.add_slider to mount the joint between the dresser and the drawer automatically. all of these operations do not change the poses of the bodies in the world, just changes the kinematic structure.

I furthermore believe that this is just the start, these are the create methods, but i think this will also be extended to have update and delete methods, since we need a way to really modify our beliefstate in real time, not just add to it, and through the use of mixins we will be able to do this seamlessly for all current and future semantic annotations, with little extra effort besides the specific implementation.

Lastly, the commit https://github.com/cram2/cognitive_robot_abstract_machine/pull/78/commits/7cc43c15d975e2bcc83569b16902906536c6ffea which adds a HasObjects mixin, so that perceived objects can easily be mounted in or on the specified semantic annotation (for example inside a drawer, or on top of a table), can be seen as a small demonstration of how little effort it requires to add functionality to semantic annotations now. 

### Other approaches tried
We initially considered extending the existing `factories.py` with more configuration options, but this quickly led to "smelly" code with excessive nesting and complexity. The decision to rethink the entire creation pipeline and move to a mixin-based approach proved much more scalable and aligned with the project's design goals.

---

### Semantic Digital Twin (Annotations & Mixins)

- Moved creation logic from `factories.py` to `mixins.py` and `semantic_annotations.py`.
- Introduced `HasRootKinematicStructureEntity`, `HasRootBody`, and `HasRootRegion` as base mixins to standardize how entities are added to the world and kinematic structure.
- Added specialized mixins like `HasCaseAsRootBody`, `HasSupportingSurface`, and `HasApertures` to encapsulate complex construction logic for furniture and structural elements.
- Standardized `create_with_new_body_in_world` methods across various annotation types, ensuring consistent attachment of geometries and connections.

### Adapters (Procthor & ROS)

- Updated `procthor_parser.py` and `procthor_pipelines.py` to utilize the new annotation creation methods, simplifying the parsing of AI2-THOR environments.
- Refined `world_fetcher.py` and other ROS adapters to ensure they correctly handle the updated world model and semantic annotations during state synchronization.

---

- **`semantic_annotations/factories.py`**: This module has been removed. Its functionality is now provided by `create_with_...` methods directly on the annotation classes.
- **`Corpus` class**: Renamed to `Cabinet` for better alignment with standard terminology.
- **`Cabinet` class (old)**: Renamed to `Case` to better describe its role as a container base.

---

- **Factory Removal**: Any code relying on `semantic_digital_twin.semantic_annotations.factories` will break. Migration involves calling the static creation methods on the respective annotation classes.
- **Class Renaming**: References to the old `Corpus` or `Cabinet` classes must be updated to `Cabinet` or `Case` respectively.
- **Test Infrastructure**: The `krrood` test dataset has been moved to a private `_dataset` module, requiring updates to import statements in tests.

---

## Other Changes

- **Semantic Position Descriptions**: Moved these to a dedicated module to support complex spatial reasoning and sampling.
- **Geometry Utilities**: Improved `Scale` and `BoundingBox` classes with new helper methods and better serialization.

---

### DIFF
--- a/test/semantic_digital_twin_test/test_semantic_annotations/test_factories.py
+++ b/test/semantic_digital_twin_test/test_semantic_annotations/test_factories.py
@@ -508,7 +508,7 @@ class TestFactories(unittest.TestCase):
             )
             door.add_handle(handle)
         with self.assertRaises(InvalidHingeActiveAxis):
-            door.calculate_world_T_hinge_based_on_handle(Vector3.X())
+            door.calculate_world_T_hinge_based_on_handle(Vector3(1, 1, 0))
 
     def test_calculate_supporting_surface(self):
         world = World()


giskardpy/src/giskardpy/motion_statechart/ros2_nodes/ros_tasks.py
```diff
@@ -5,7 +5,11 @@
 from dataclasses import dataclass, field
 
 from geometry_msgs.msg import PoseStamped, Pose, Point, Quaternion
-from nav2_msgs.action import NavigateToPose
+
+try:
+    from nav2_msgs.action import NavigateToPose
```
Tigul: I don't think that's necessary

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
```diff
@@ -1,106 +1,658 @@
...
+        :param active_axis: The active axis for the connection.
+        :param connection_multiplier: The multiplier for the connection.
+        :param connection_offset: The offset for the connection.
+        :return: The created semantic annotation instance.
+        """
+        if connection_type is None:
+            raise MissingConnectionType(cls)
+
+        self_instance = cls(name=name, root=kinematic_structure_entity)
+        world_root_T_self = (
```
Tigul: You can write this as 
~~~
world_root_T_self or HomogeneousTransformationMatrix()
~~~

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
```diff
@@ -1,106 +1,658 @@
...
+                    axis=active_axis,
+                    dof_id=dof.id,
+                )
+            elif connection_type == FixedConnection:
+                world_root_C_self = FixedConnection(
+                    parent=world.root,
+                    child=kinematic_structure_entity,
+                    parent_T_connection_expression=world_root_T_self,
+                )
+            else:
```
Tigul: Whats with other types like Connection6dof?

semantic_digital_twin/src/semantic_digital_twin/utils.py
```diff
@@ -193,3 +194,12 @@ def camel_case_split(word: str) -> List[str]:
             start = i
     result.append(word[start:])
     return result
+
+
+class Direction(IntEnum):
```
Tigul: Should be in position_descriptions

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
```diff
@@ -1,106 +1,658 @@
...
+    @classmethod
+    def _create_with_connection_in_world(
+        cls,
+        connection_type: Type[Connection],
+        name: PrefixedName,
+        world: World,
+        kinematic_structure_entity: KinematicStructureEntity,
+        world_root_T_self: Optional[HomogeneousTransformationMatrix] = None,
+        connection_limits: Optional[DegreeOfFreedomLimits] = None,
+        active_axis: Vector3 = Vector3.Z(),
```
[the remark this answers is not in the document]
LucaKro: its just a default value so that this method is LSP compliant

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
LucaKro: the types that i currently included are the ones i know we have usecases for. so far i dont think we will put different connection types there, so for now id say is yagni

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
```diff
@@ -1,106 +1,658 @@
...
+        self.doors.append(door)
+
+
+@dataclass(eq=False)
+class HasLeftRightDoor(HasDoors):
+    """
+    A mixin class for semantic annotations that have a left and a right door.
+    """
+
+    left_door: Optional[Door] = None
```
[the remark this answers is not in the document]
LucaKro: hmm id be fine with that, but back when i talked to @tomsch420 about this he liked this better.

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/semantic_annotations.py
```diff
@@ -1,141 +1,396 @@
...
 
 @dataclass(eq=False)
-class Drawer(SemanticAnnotation):
-    container: Container
-    handle: Handle
+class Drawer(Furniture, HasCaseAsRootBody, HasHandle, HasSlider, HasObjects):
+
+    @property
+    def opening_direction(self) -> Direction:
+        return Direction.Z
```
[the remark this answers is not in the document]
LucaKro: this is for drawer. in what case would a drawer not have its opening direction in z?

giskardpy/src/giskardpy/motion_statechart/ros2_nodes/ros_tasks.py
tomsch420: i agree with luca

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
```diff
@@ -1,106 +1,629 @@
...
+    root: KinematicStructureEntity = field(kw_only=True)
+    """
+    The root kinematic structure entity of the semantic annotation.
+    """
+
+    @classproperty
+    def _parent_connection_type(self) -> Type[Connection]:
+        """
+        The type of connection used to connect the root kinematic structure entity to the world.
+        Override if needed.
```
tomsch420: why do you say Override if needed here? Isnt this true for every method not marked with @final?

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
```diff
@@ -147,19 +670,136 @@ def create_from_entity(
...
+        self_C_supporting_surface = FixedConnection(
+            parent=self.root, child=supporting_surface
+        )
+        with self._world.modify_world():
+            self._world.add_region(supporting_surface)
+            self._world.add_connection(self_C_supporting_surface)
+        self.supporting_surface = supporting_surface
+        return supporting_surface
+
+    def points_on_surface(self, amount: int = 100) -> List[Point3]:
```
tomsch420: i would call this points_on_supporting_surface

semantic_digital_twin/src/semantic_digital_twin/world_description/degree_of_freedom.py
```diff
@@ -81,6 +81,12 @@ def resolve(self) -> float:
         return self.dof._world.state[self.dof.id].jerk
 
 
+@dataclass(eq=False)
+class DegreeOfFreedomLimits:
+    lower_limit: DerivativeMap[float]
```
tomsch420: doc

tomsch420 (changes requested):

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
```diff
@@ -1,106 +1,629 @@
...
+        """
+        Generate default degree of freedom limits for a given connection type.
+
+        :param active_connection_type: The type of connection to generate limits for.
+        :return: The generated degree of freedom limits.
+        """
+        lower_limits = DerivativeMap[float]()
+        upper_limits = DerivativeMap[float]()
+        if active_connection_type == PrismaticConnection:
+            lower_limits.position = -np.inf
```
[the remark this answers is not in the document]
LucaKro: I wanted to do that, but @ichumuh did not like that, because he doesnt want default values for connections in general, but for these factories they make sense i think

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
LucaKro: because any semantic annotation which inherits from this class should think about if fixedconnection is the correct connection here. so far it usually is, but i think its worth noting. but if yo disagree I can also remove it, im not married to that docstring

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
```diff
@@ -1,106 +1,629 @@
...
+    """
+    The apertures of the semantic annotation.
     """
 
-    region: Region = field(kw_only=True)
+    def add_aperture(self, aperture: Aperture):
+        """
+        Cuts a hole in the semantic annotation's body for the given body annotation.
+
+        :param aperture: The aperture to cut a hole for.
```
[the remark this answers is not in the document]
LucaKro: we need some more aperture science for that :smiling_face_with_tear:

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/position_descriptions.py
```diff
@@ -0,0 +1,159 @@
...
+
+from semantic_digital_twin.datastructures.variables import SpatialVariables
+
+
+class IntervalConstants:
+    """
+    Predefined intervals for semantic directions.
+    """
+
+    ZERO_DIRAC = (0, 0, Bound.CLOSED, Bound.CLOSED)
```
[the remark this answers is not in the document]
LucaKro: what would you call it then

test/krrood_test/_dataset/__init__.py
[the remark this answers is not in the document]
LucaKro: reverted

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
LucaKro: well the problem here is not actually this code imo, but that connections are not LSP compliant. thats why i cannot use the same constructor for all of them, and thus need different cases

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
```diff
@@ -1,106 +1,658 @@
...
+    @classproperty
+    def _parent_connection_type(self) -> Type[Connection]:
+        """
+        The type of connection used to connect the root kinematic structure entity to the world.
+        Override if needed.
+        """
+        return FixedConnection
+
+    @classmethod
+    def _create_with_connection_in_world(
```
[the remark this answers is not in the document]
LucaKro: oh i see what you mean, but no i dont think so. the only thing this method does is deal with the connection, so by pulling the connection outside, im removing the purpose of this method, and cause more duplicated code

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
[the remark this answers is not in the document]
LucaKro: also sth we talked about in person. Doubledoor now has a method to calculate this from a viewpoint

semantic_digital_twin/src/semantic_digital_twin/world.py
```diff
@@ -1514,7 +1530,7 @@ def compute_child_kinematic_structure_entities(
             self.kinematic_structure.successors(kinematic_structure_entity.index)
         )
 
-    @lru_cache(maxsize=_LRU_CACHE_SIZE)
+    # @lru_cache(maxsize=_LRU_CACHE_SIZE)
```
tomsch420: im all in for deleting this all together

krrood/src/krrood/adapters/json_serializer.py
```diff
@@ -218,33 +220,78 @@ def to_json(obj: Union[SubclassJSONSerializer, Any]) -> JSON_RETURN_TYPE:
...
+    """
+
+    remove: List[Any] = field(default_factory=list)
+    """
+    The items that have been removed from the attribute.
+    """
+
+    def to_json(self) -> Dict[str, Any]:
+        return {
+            JSON_TYPE_NAME: get_full_class_name(self.__class__),
```
tomsch420: **super().to_json()

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/position_descriptions.py
```diff
@@ -0,0 +1,159 @@
...
+
+
+class HorizontalSemanticDirection(SimpleInterval, SemanticDirection):
+    """
+    Semantic directions for horizontal positioning.
+    """
+
+    FULLY_LEFT = IntervalConstants.ZERO
+    LEFT = IntervalConstants.ZERO_TO_ONE_THIRD
+    CENTER = IntervalConstants.ONE_THIRD_TO_TWO_THIRD
```
sunava: maybe middle?

tomsch420 (changes requested):

krrood/src/krrood/adapters/json_serializer.py
```diff
@@ -155,30 +156,31 @@ def from_json(cls, data: Dict[str, Any], **kwargs) -> Self:
...
-                    if change.get("remove"):
-                        setattr(self, key, None)
-                    if change.get("add"):
-                        setattr(
-                            self,
-                            key,
-                            from_json(change["add"], **kwargs),
-                        )
+        for diff in diffs:
+            if not hasattr(self, diff.attribute_name):
```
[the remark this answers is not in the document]
LucaKro: i guess it should never happen, ill remove it

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
```diff
@@ -59,6 +65,42 @@
...
+    Ensures that any modifications to the attributes of an instance of WorldEntityWithID are properly recorded and any
+    resultant changes are appended to the current model modification block in the world model manager. Keeps track of
+    the pre- and post-modification states of the object to compute the differences and maintain a log of updates.
+    """
+
+    @wraps(func)
+    def wrapper(self: SemanticAnnotation, *args: Any, **kwargs: Any) -> Any:
+
+        object_before_change = to_json(self)
+        result = func(self, *args, **kwargs)
```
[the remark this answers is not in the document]
LucaKro: it works with everything that the to_json method works for, its just that i have tested these two cases. I will tell junie to add a few more cases, but its out of scope to test all possible variations. if its used on a class that doesnt work with to_json, it will throw an error and then we deal with it

semantic_digital_twin/src/semantic_digital_twin/exceptions.py
```diff
@@ -125,6 +128,99 @@ class UsageError(LogicalError):
...
+
+    def __post_init__(self):
+        self.message = (
+            f"The semantic annotation of type {self.missing_semantic_annotation_class.__name__} is required"
+            f" by {self.semantic_annotation_class.__name__}, but is missing."
+        )
+
+
+@dataclass
+class InvalidPlaneDimensions(UsageError):
```
Tigul: This seems to be written for a specific use-case but I don't see why the depth of a plane can't be less than the width

pycram/src/pycram/process_modules/default_process_modules.py
```diff
@@ -219,8 +219,8 @@ def _execute(self, desig: ClosingMotion):
             )
         )[0]
 
-        lower_limit = container_connection.dof.lower_limits.position
-        upper_limit = container_connection.dof.upper_limits.position
+        lower_limit = container_connection.dof.lower.position
+        upper_limit = container_connection.dof.upper.position
```
ichumuh: `.limits.lower`?

semantic_digital_twin/src/semantic_digital_twin/orm/model.py
```diff
@@ -247,8 +249,8 @@ class DegreeOfFreedomMapping(AlternativeMapping[DegreeOfFreedom]):
     def from_domain_object(cls, obj: DegreeOfFreedom):
         return cls(
             name=obj.name,
-            lower_limits=obj.lower_limits.data,
-            upper_limits=obj.upper_limits.data,
+            lower_limits=obj.limits.lower.data,
+            upper_limits=obj.limits.upper.data,
```
ichumuh: i don't think that's right

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
```diff
@@ -1,106 +1,658 @@
...
     @property
     def bodies(self) -> Iterable[Body]:
-        return [self.body]
+        """
+        The bodies that are part of the semantic annotation.
+        """
+        return [self.root]
+
+    @classmethod
+    def create_with_new_body_in_world(
```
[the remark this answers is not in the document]
ichumuh: i don't like wrapper functions which simply pass along all parameters to another function

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
```diff
@@ -1,106 +1,621 @@
...
+
+    def _remove_aperture_geometry_from_parent(self, aperture: Aperture):
+        """
+        Remove the geometry of the aperture from the parent body's collision and visual geometry.
+
+        :param aperture: The aperture whose geometry should be removed.
+        """
+        world = self._world
+        world._forward_kinematic_manager.recompile()
+        world._forward_kinematic_manager.recompute()
```
ichumuh: and why do you need to recompile?

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
```diff
@@ -1,106 +1,621 @@
...
+    def add_hinge(
+        self,
+        hinge: Hinge,
+    ):
+        """
+        Add a hinge to the semantic annotation.
+
+        :param hinge: The hinge to add.
+        """
+        self._attach_parent_entity_in_kinematic_structure(
```
ichumuh: `in` -> `to`?

semantic_digital_twin/src/semantic_digital_twin/world_description/world_modification.py
```diff
@@ -431,3 +501,39 @@ def _from_json(cls, data: Dict[str, Any], **kwargs) -> Self:
...
+
+
+def synchronized_attribute_modification(func):
+    """
+    Decorator to synchronize attribute modifications.
+
+    Ensures that any modifications to the attributes of an instance of WorldEntityWithID are properly recorded and any
+    resultant changes are appended to the current model modification block in the world model manager. Keeps track of
+    the pre- and post-modification states of the object to compute the differences and maintain a log of updates.
+    """
```
ichumuh: i don't like the vibes of this decorator. is this to fix synchronization issues?

semantic_digital_twin/src/semantic_digital_twin/adapters/procthor/procthor_pipelines.py
```diff
@@ -1,84 +1,119 @@
...
-    )
-    container_factory = ContainerFactory(
-        name=PrefixedName(drawer.name.name + "_container", drawer.name.prefix),
-        scale=drawer.collision.as_bounding_box_collection_at_origin(
-            HomogeneousTransformationMatrix(reference_frame=drawer._world.root)
+    drawer_scale = (
+        drawer_body.collision.as_bounding_box_collection_at_origin(
+            HomogeneousTransformationMatrix(reference_frame=drawer_body._world.root)
         )
         .bounding_boxes[0]
```
AbdelrhmanBassiouny: why only the first box? will this give a correct scale if modeled as multiple boxes?

semantic_digital_twin/src/semantic_digital_twin/exceptions.py
```diff
@@ -125,6 +128,99 @@ class UsageError(LogicalError):
...
+        self.message = f"The depth of a plane must be less than its width or height. This doesnt hold for your door with dimensions {self.scale}"
+
+
+@dataclass
+class InvalidAxisError(UsageError):
+    """
+    Raised when an invalid axis is provided.
+    """
+
+    axis: Vector3
```
AbdelrhmanBassiouny: Could there be more context parameters here, as this could mean anything?

semantic_digital_twin/src/semantic_digital_twin/world_description/degree_of_freedom.py
```diff
@@ -81,6 +81,32 @@ def resolve(self) -> float:
...
+    upper: DerivativeMap[float] = field(default=None)
+    """
+    Upper limits of the degree of freedom.
+    """
+
+    def __post_init__(self):
+        self.lower = self.lower or DerivativeMap()
+        self.upper = self.upper or DerivativeMap()
+
+    def __deepcopy__(self, memo):
```
AbdelrhmanBassiouny: why? won't deepcopy already do exactly that?

semantic_digital_twin/src/semantic_digital_twin/pipeline/pipeline.py
```diff
@@ -132,33 +132,33 @@ class BodyFactoryReplace(Step):
     Replace bodies in the world that match a given condition with new structures created by a factory.
     """
 
+    annotation_creator: Callable[[Body, World], HasRootKinematicStructureEntity]
+    """
+    A callable that takes a Body, and creates a new semantic annotation (including a new body) for it, and adds them to the world.
+    """
+
     body_condition: Callable[[Body], bool] = lambda x: bool(
```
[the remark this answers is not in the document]
LucaKro: maybe, but this is out of scope for this pr imo

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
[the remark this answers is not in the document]
LucaKro: talked out this in person, to resolve this i think we need to have a look at connections again

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
LucaKro: in practice, when do you think these could be conflicting? because rn i think the only way if its conflicting, then your definition of a concept is either incorrect or inefficient (as in you put inherit from HasBody and from a class that also inherits from hasbody or sth, which i think is fine to throw an error. also you dont really need to guess. you type annotation.add... and it will tell you all the things you can add. if its add_hinge, then you give it exactly that, a Hinge, nothing more, nothing less. 

I think making a general add method becomes really messy because we not just change the kinematic structure, that would be fine, but we also assign for example the handle attribute of self. and we cannot do that in a general way (at least so far nobody i talked to could think of a good way, so lmk if you have an idea.) so either we have a add_handle method, which calls a generic "change the kinematic structure accordingly" method and then assigns the attribute correctly (which is the current way) or we have the user handle the attribute assignment, which i really dont like because it also causes even more problems with attribute change synchronization.

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/semantic_annotations.py
```diff
@@ -274,25 +272,31 @@ def calculate_world_T_hinge_based_on_handle(
...
+                    symbolic_math.sign(-1 * door_P_handle.y)
+                    if door_P_handle.y != 0
+                    else 1
+                )
                 offset = sign * (scale.y / 2)
                 door_T_hinge = HomogeneousTransformationMatrix.from_xyz_rpy(y=offset)
 
             case _:
-                raise InvalidAxisError(axis=opening_axis)
+                raise InvalidHingeActiveAxis(axis=opening_axis)
```
ichumuh: no x case after all?

PR #82, opened 2025-12-20
Mujoco Sim .dae mesh and freejoint compatability

maltehue:
This PR adds two fixes for MuJoCo integration in `multi_sim.py` and a small fix to `urdf.py` when creating `ParsingError` instances:

### Auto-convert DAE meshes to STL

- Instead of skipping unsupported `.dae` (Collada) meshes, `_parse_geom()` now converts them to STL using trimesh.
- Converted meshes are cached in the  asset folder and reused if the target `.stl` already exists, avoiding redundant conversions.

### Fix nested free joints

- Bodies added via `Connection6DoF` must connect to MuJoCoSim's  newly created `world` root body, not to the root body that was used when creating the world in semDT.
- Previously, these created non-top-level `free` joints, causing `ValueError: free joint can only be used on top level` during `spec.compile()`.
- Now they are moved under the world root, ensuring all free joints remain top-level as MuJoCo requires.

### DIFF
--- a/semantic_digital_twin/src/semantic_digital_twin/adapters/multi_sim.py
+++ b/semantic_digital_twin/src/semantic_digital_twin/adapters/multi_sim.py
@@ -1360,11 +1360,6 @@ class MujocoBuilder(MultiSimBuilder):
             f"Converting Collada mesh to STL for MuJoCo: {original_mesh_file_path}"
         )
         tm = trimesh.load(original_mesh_file_path, force="mesh")
-        if tm.is_empty:
-            logger.warning(
-                f"Failed to load .dae mesh (empty): {original_mesh_file_path}. Skipping."
-            )
-            return
 
         tm.export(stl_file_path)
 


HoangGiang93:
> > I would like to raise a concern regarding DAE support. The DAE (Collada) format is effectively deprecated: the upstream repository has been inactive for over 7 years, and many developers have already moved away from it (see, for example, https://devtalk.blender.org/t/moving-collada-i-o-to-legacy-status/34621).
> > From my experience with multiverse_parser, DAE introduces numerous practical issues that are costly to maintain and debug. While URDF still allows DAE meshes, this is largely a legacy artifact—URDF itself is already showing its age.
> > Given this, the most reasonable approach when encountering a URDF that references DAE meshes is to convert those assets to a modern, well-supported format (e.g., OBJ) and import them through the standard pipeline. Adding native support for DAE would mean deliberately maintaining legacy functionality, comparable to adding new code paths specifically to support ROS Noetic today.
> 
> I agree with you on not supporting outdated software. But looking around at for example UR (https://github.com/UniversalRobots/Universal_Robots_ROS2_Description/tree/jazzy/meshes/ur8long/visual) and Franka (https://github.com/frankarobotics/franka_description/tree/main/meshes/robot_arms/fr3v2.1/visual) they still upload new visual meshes as .dae as recent as 2-3months ago. For us that would mean that we have to maintain separate copies of all those robots or everytime we want to use any of those robots we first have to convert their robot descriptions locally. The latter is similar to what this PR does by converting the .dae to .stl using trimesh.

ok... I understand that, but my concern still holds ;__;

maltehue:
> > > I would like to raise a concern regarding DAE support. The DAE (Collada) format is effectively deprecated: the upstream repository has been inactive for over 7 years, and many developers have already moved away from it (see, for example, https://devtalk.blender.org/t/moving-collada-i-o-to-legacy-status/34621).
> > > From my experience with multiverse_parser, DAE introduces numerous practical issues that are costly to maintain and debug. While URDF still allows DAE meshes, this is largely a legacy artifact—URDF itself is already showing its age.
> > > Given this, the most reasonable approach when encountering a URDF that references DAE meshes is to convert those assets to a modern, well-supported format (e.g., OBJ) and import them through the standard pipeline. Adding native support for DAE would mean deliberately maintaining legacy functionality, comparable to adding new code paths specifically to support ROS Noetic today.
> > 
> > 
> > I agree with you on not supporting outdated software. But looking around at for example UR (https://github.com/UniversalRobots/Universal_Robots_ROS2_Description/tree/jazzy/meshes/ur8long/visual) and Franka (https://github.com/frankarobotics/franka_description/tree/main/meshes/robot_arms/fr3v2.1/visual) they still upload new visual meshes as .dae as recent as 2-3months ago. For us that would mean that we have to maintain separate copies of all those robots or everytime we want to use any of those robots we first have to convert their robot descriptions locally. The latter is similar to what this PR does by converting the .dae to .stl using trimesh.
> 
> ok... I understand that, but my concern still holds ;__;


Yes and its still a valid concern. I had a look at how the mujoco menagerie is handling that and they derived a new mjcf .xml file for the UR10e (same arms are on tracy) https://github.com/google-deepmind/mujoco_menagerie/tree/main/universal_robots_ur10e
and it seems that they need some manual extra steps to have a nice simulation of the robot. So maybe it is worth keeping mjcf files of our robot for simulation? Can there be more benefits than just having the .dae files replaced? Like accurate physics parameters or controller definitions that could then also be maintained in the xml files?

maltehue:
> we should always maintain MJCF (xml) for robot simulation, not URDF... URDF is sooooo deprecated and limited

Are we maintaining them somewhere already? Should we also maintain MJCF (xml) files of objects similar to how it is done here? https://github.com/ARISE-Initiative/robosuite/tree/master/robosuite/models/assets/objects

PR #141, opened 2026-02-01
removed wrong hash function from region

maltehue:
I added the test because i encountered the error when adding regions to the world while having a tf publisher running.
I could also remove it as it does not seem to related. This fix is probably related to hash function rework that i think @Tigul did some time ago?


tomsch420:
is the hash now its uuid?

PR #155, opened 2026-02-04
Bigfixes from suturo session

ichumuh:
- fixed world fetcher to apply modifications right after parsing form json
- fixed some semantic annotation fail
- sets can be json deserialized


giskardpy/src/giskardpy/model/world_config.py
```diff
@@ -179,6 +178,8 @@ def setup_collision_config(self):
         pass
 
     def setup_world(self):
+        from krrood.ormatic.utils import create_engine
```
tomsch420: why?

PR #169, opened 2026-02-09
[EQL] Cleaning

AbdelrhmanBassiouny:
## Builder Pattern:
When writing the query, the user always has the same `Query` object and extends it with further operations like `where`, `grouped_by`, `having`, `distinct`, `limit`, and `ordered_by`. 

In addition, wrapping the `Query` object in a quantifier also records the quantification metadata and still gives back the same `Query` object. 

This makes the query construction centralized and simplifies query optimizations in the future.

Query object wires things together automatically when:

- calling `evaluate` or `tolist`
- using the Query object inside another query or operation (e.g., inside an aggregator or as part of a condition, ..etc.)

## Decoupled rustworkx and visualization from Query classes + No more singleton graph:
- Previously rustworkx was used, having a singleton graph that contains all query graphs ever created in runtime. Now every query is separate and does is garpage collected normally. Each query builds a graph locally usnig builtin implementation instead of Rustworkx.

- Rustworkx use is now limited to building a graph when needed for visualization. This makes eagerly building queries and sub-queries and connecting them together much simpler and faster.

- Visualization-related code is removed from the Query expressions and is now in a single file dedicated to building a graph for visualization, if needed.

## New Classes

- `InstantiatedVariable` as a new class that inherits from `Variable` for the cases when the variable instantiates its domain, like Predicates or Rule inference. Previously, its functionality was covered by `Variable`, so this makes `Variable` much simpler.

- `CanHaveDomainSource` as a super class for all types of variables, that adds an optional Enum field to specify the source type of the Domain of the variable (e.g., EXPLICIT, DEDUCED)

- `ExternallySetVariable` as a superclass for variables that have their values set by another expression or by another part of the application (e.g., conclusions of rules or inference of probabilistic models, etc.)

- `MultiarityExpression` abstract base class for classes with more than two operands. Now every class is either a `UnaryExpression`, a `BinaryExpression`, or a `MultiarityExpression`.

- `PerformsCartesianProduct` as a super class for operations that involve performing cartesian product on the values of its operands. This unifies the implementation and makes the operations that do so more explicit.

- `MultiArityExpressionThatPerformsCartesianProduct` is a subclass of both above classes, as this pattern occurs multiple times in other expressions.

- `DerivedExpression` abstract superclass for classes that are just pass-through operations and derive their value directly from their children.

- `Filter` class as a super class for filter operations like `Where` and `Having`.

- `TruthValueOperator` as abstact base class for classes that require and use the truth value of its children, this makes the children evaluate their truth value only when they are a child of such a class.

- `ExpressionBuilder`, and all its subclasses are new classes that collect metadata about query operations to construct and wire them during query building.

## Reorganization
`symbolic.py` contained most expressions of EQL and was around 3k + lines. Now this file has been split into multiple files and packages.
- `core` sub package "containing base expressions, variables, and domain mappings."
- `operators` sub package "containing logical operators, aggregators, set operations, and others."
- `query` sub package "is the highest level and nearest to user APIs, it has the query builders and the query operations."
- `factories.py` contains all the user APIs; only this needs to be imported for most use cases.
- Moved symbol_graph and `Symbol` class to a separate package outside EQL.

##  New Use Related Features
- Aggregations can be used inside 'ordered_by`.
- `distinct` as a stand-alone method can be used on variables, queries, or aggregators. 
- `limit` now is its own method accessible from the Query object, similar to `where/having,...etc` instead of being an argument for the evaluate method.
- you always have a `Query` object, and you can explicitly build it by calling `.build()` on it, but after that, you cannot call any query modifiers like `where()` or `having()`, but you can call things like `limit` since it affects results onl,y not query structure. Otherwise, it will automatically be called as mentioned above in various situations. The reason for this is to give some control to the user on when to build the query and disallow modifications to it.
- you can do `query.first()` to evaluate and get the first result. This is especially useful for queries quantified with `the` as there's only one result if any, instead of doing `query.tolist()[0]` or `next(query.evaluate())`.
- Added `length()` symbolic function to symbolically represent and evaluate the built-in `len()` method.
- Added `deduced_variable()` factory for variables that have no explicit domain and are inferred using deductive reasoning (e.g., through conclusions of rules)

## Fixes
- `Union` class is not a Logical operator but a set operator; it has no meaning to say a Union is True or False or has True or False values. But it makes sense to say it has certain members inside its set and others not.
- Subqueries made use of the bindings of their parent query, which is wrong; now they don't do that.
- Variable domain generators were not possible to iterate over when is being iterated over somewhere else; that is fixed.

Other minor improvements to API docs and hints.

After this PR is merged, the documentation will be reworked.

### DIFF
--- a/krrood/src/krrood/entity_query_language/core/base_expressions.py
+++ b/krrood/src/krrood/entity_query_language/core/base_expressions.py
@@ -445,10 +445,7 @@ class SymbolicExpression(ABC):
         Enter a context where this symbolic expression is the current parent symbolic expression. This updates the
         current parent symbolic expression, the context stack and returns this expression.
         """
-        expression = self
-        if (expression is self._root_) or (expression._parent_ is self._root_):
-            expression = expression._conditions_root_
-        SymbolicExpression._symbolic_expression_stack_.append(expression)
+        SymbolicExpression._symbolic_expression_stack_.append(self)
         return self
 
     def __exit__(self, *args):
--- a/krrood/src/krrood/entity_query_language/core/mapped_variable.py
+++ b/krrood/src/krrood/entity_query_language/core/mapped_variable.py
@@ -30,7 +30,7 @@ from ..operators.comparator import Comparator
 from ..utils import (
     T,
     merge_args_and_kwargs,
-    convert_args_and_kwargs_into_a_hashable_key,
+    convert_args_and_kwargs_into_hashable_key,
 )
 
 from ...symbol_graph.helpers import get_field_type_endpoint
@@ -80,7 +80,7 @@ class CanBehaveLikeAVariable(Selectable[T], ABC):
         """
         args = (self,) + args
         all_kwargs = merge_args_and_kwargs(type_, args, kwargs, ignore_first=True)
-        return convert_args_and_kwargs_into_a_hashable_key(all_kwargs)
+        return convert_args_and_kwargs_into_hashable_key(all_kwargs)
 
     def __getattr__(self, name: str) -> CanBehaveLikeAVariable[T]:
         # Prevent debugger/private attribute lookups from being interpreted as symbolic attributes
@@ -300,9 +300,7 @@ class MappedVariableCacheItem:
 
     @cached_property
     def hashable_key(self):
-        return (self.type,) + convert_args_and_kwargs_into_a_hashable_key(
-            self.all_kwargs
-        )
+        return (self.type,) + convert_args_and_kwargs_into_hashable_key(self.all_kwargs)
 
     def __hash__(self):
         return hash(self.hashable_key)
--- a/krrood/src/krrood/entity_query_language/core/variable.py
+++ b/krrood/src/krrood/entity_query_language/core/variable.py
@@ -32,13 +32,13 @@ from .base_expressions import (
 from .mapped_variable import CanBehaveLikeAVariable
 from ..cache_data import ReEnterableLazyIterable
 from ..enums import DomainSource
+from ..failures import NoChildToReplace
 from ..operators.set_operations import MultiArityExpressionThatPerformsACartesianProduct
 from ..utils import (
     T,
     is_iterable,
     make_list,
 )
-
 
 DomainType = Iterable[T]
 """
@@ -119,7 +119,7 @@ class Variable(CanHaveDomainSource[T]):
     def _replace_child_field_(
         self, old_child: SymbolicExpression, new_child: SymbolicExpression
     ):
-        raise ValueError(f"class {self.__class__} does not have children")
+        raise NoChildToReplace(self, old_child, new_child)
 
     @cached_property
     def _name_(self) -> str:
--- a/krrood/src/krrood/entity_query_language/operators/logical_quantifiers.py
+++ b/krrood/src/krrood/entity_query_language/operators/logical_quantifiers.py
@@ -7,6 +7,7 @@ that evaluate conditions over the values of a variable.
 
 from __future__ import annotations
 
+import uuid
 from abc import ABC
 from dataclasses import dataclass
 from functools import cached_property
@@ -27,17 +28,9 @@ class QuantifiedConditional(LogicalBinaryOperator, ABC):
     def variable(self):
         return self.left
 
-    @variable.setter
-    def variable(self, value):
-        self.left = value
-
     @property
     def condition(self):
         return self.right
-
-    @condition.setter
-    def condition(self, value):
-        self.right = value
 
 
 @dataclass(eq=False, repr=False)
@@ -48,7 +41,7 @@ class ForAll(QuantifiedConditional):
     """
 
     @cached_property
-    def condition_unique_variable_ids(self) -> List[int]:
+    def condition_unique_variable_ids(self) -> List[uuid.UUID]:
         return [
             v._binding_id_
             for v in self.condition._unique_variables_.difference(
--- a/krrood/src/krrood/entity_query_language/query/builders.py
+++ b/krrood/src/krrood/entity_query_language/query/builders.py
@@ -104,6 +104,8 @@ class FilterBuilder(ExpressionBuilder, ABC):
 
             if isinstance(expr, Aggregator):
                 aggregators.append(expr)
+                # No need to traverse inside aggregators
+                return False
             elif isinstance(expr, Selectable) and not isinstance(expr, Literal):
                 non_aggregators.append(expr)
 
@@ -111,8 +113,8 @@ class FilterBuilder(ExpressionBuilder, ABC):
             if aggregators and non_aggregators:
                 return True
 
-            if isinstance(expr, (Query, Aggregator)):
-                # Subqueries/Aggregator are a boundary, we don't need to traverse inside them.
+            if isinstance(expr, Query):
+                # Subqueries are a boundary, we don't need to traverse inside them.
                 return False
 
             return any(walk(child) for child in expr._children_)
@@ -318,16 +320,17 @@ class GroupedByBuilder(ExpressionBuilder):
     def aggregators_and_non_aggregators_in_ordered_by(
         self,
     ) -> Tuple[List[Aggregator], List[Selectable]]:
-        non_aggregated_variables, aggregators = [], []
-        if self.query._ordered_by_builder_:
-            variable = self.query._ordered_by_builder_.variable
-            if isinstance(variable, Aggregator):
-                aggregators.append(variable)
-                if variable._child_ is not None:
-                    non_aggregated_variables.append(variable._child_)
-            else:
-                non_aggregated_variables.append(variable)
-        return aggregators, non_aggregated_variables
+        if not self.query._ordered_by_builder_:
+            return [], []
+
+        variable = self.query._ordered_by_builder_.variable
+
+        if isinstance(variable, Aggregator):
+            return [variable], (
+                [variable._child_] if variable._child_ is not None else []
+            )
+
+        return [], [variable]
 
     @cached_property
     def aggregators_in_selected_variables(self) -> Tuple[Aggregator, ...]:
@@ -343,6 +346,10 @@ class GroupedByBuilder(ExpressionBuilder):
 
 @dataclass(eq=False)
 class QuantifierBuilder(ExpressionBuilder):
+    """
+    Builds a result quantifier (An/The) of the specified type with the given child and quantification constraint.
+    """
+
     type: Type[ResultQuantifier] = An
     """
     The type of the quantifier to be built.
--- a/krrood/src/krrood/entity_query_language/query_graph.py
+++ b/krrood/src/krrood/entity_query_language/query_graph.py
@@ -134,13 +134,15 @@ class QueryGraph:
         :param parent_node: The parent node of the children to add.
         """
         parent_expression = parent_node.data
+        selected_var_ids = (
+            [v._binding_id_ for v in parent_expression._selected_variables_]
+            if isinstance(parent_expression, Query)
+            else []
+        )
         for child in parent_expression._children_:
             child_node = self.construct_graph(child)
-            if isinstance(parent_expression, Query):
-                if child._binding_id_ in [
-                    v._binding_id_ for v in parent_expression._selected_variables_
-                ]:
-                    child_node.enclosed = True
+            if child._binding_id_ in selected_var_ids:
+                child_node.enclosed = True
             child_node.parent = parent_node
 
 
--- a/krrood/src/krrood/entity_query_language/rules/conclusion_selector.py
+++ b/krrood/src/krrood/entity_query_language/rules/conclusion_selector.py
@@ -53,28 +53,26 @@ class ConclusionSelector(TruthValueOperator, ABC):
         """
         new_condition = chained_logic(AND, *conditions)
 
-        current_conditions_root = (
-            cls._get_current_conditions_root_for_rule_type_from_context_stack()
+        current_context = cls._get_current_context_condition()
+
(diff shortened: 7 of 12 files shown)


krrood/src/krrood/entity_query_language/core/domain_mapping.py
```diff
@@ -0,0 +1,408 @@
...
+        yield from value
+
+    @cached_property
+    def _name_(self):
+        return f"Flatten({self._child_._name_})"
+
+    @property
+    def _is_iterable_(self):
+        """
+        :return: False as Flatten does not preserve the original iterable structure.
```
tomsch420: Thats very confusing to me

krrood/src/krrood/entity_query_language/query/builders.py
```diff
@@ -0,0 +1,391 @@
...
+
+    @cached_property
+    def aggregators_and_non_aggregators_in_conditions(
+        self,
+    ) -> Tuple[Tuple[Aggregator, ...], Tuple[Selectable, ...]]:
+        """
+        :return: A tuple containing the aggregators and non-aggregators in the conditions.
+        """
+        aggregators, non_aggregators = [], []
+        for cond in self.conditions:
```
tomsch420: gives me group_by vibes, so perhaps itertools.groupby?

krrood/src/krrood/entity_query_language/core/base_expressions.py
```diff
@@ -0,0 +1,765 @@
...
+        return next(self.evaluate())
+
+    def evaluate(
+        self,
+    ) -> Iterator[TypingUnion[T, Dict[TypingUnion[T, SymbolicExpression], T]]]:
+        """
+        Evaluate the query and map the results to the correct output data structure.
+        This is the exposed evaluation method for users.
+        """
+        SymbolGraph().remove_dead_instances()
```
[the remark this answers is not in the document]
AbdelrhmanBassiouny: apparently test_memory_leak fails without it.

krrood/src/krrood/entity_query_language/core/variable.py
```diff
@@ -0,0 +1,292 @@
...
+
+
+@dataclass(eq=False, init=False, repr=False)
+class Literal(Variable[T]):
+    """
+    Literals are variables that are not constructed by their type but by their given data.
+    """
+
+    def __init__(
+        self,
```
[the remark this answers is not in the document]
AbdelrhmanBassiouny: Refactored the whole file haha

krrood/src/krrood/entity_query_language/core/base_expressions.py
```diff
@@ -0,0 +1,802 @@
...
+
+    Symbolic expressions form a rooted directed acyclic graph and are evaluated lazily to produce
+    bindings for variables, subject to logical constraints.
+    """
+
+    _id_: int = field(init=False, repr=False, default_factory=lambda: uuid.uuid4().int)
+    """
+    Unique identifier of this node.
+    """
+    _conclusion_: typing.Set[Conclusion] = field(init=False, default_factory=set)
```
LucaKro: `_conclusions_`?

krrood/src/krrood/entity_query_language/core/base_expressions.py
```diff
@@ -0,0 +1,802 @@
...
+    """
+    Set of conclusion expressions attached to this node, these are evaluated when the truth value of this node is true
+    during evaluation.
+    """
+    _symbolic_expression_stack_: ClassVar[List[SymbolicExpression]] = []
+    """
+    The current stack of symbolic expressions that has been entered using the ``with`` statement.
+    """
+    _is_false__: bool = field(init=False, repr=False, default=False)
```
LucaKro: when do you use double trailing underscore, and when just single?

krrood/src/krrood/entity_query_language/core/base_expressions.py
```diff
@@ -0,0 +1,802 @@
...
+        if parent is self._parent__:
+            self._parent_ = None
+
+    def _update_children_(
+        self, *children: SymbolicExpression
+    ) -> Tuple[SymbolicExpression, ...]:
+        """
+        Update multiple children expressions of this symbolic expression.
+        """
+        from .variable import Literal
```
LucaKro: why local import?

krrood/src/krrood/entity_query_language/core/base_expressions.py
```diff
@@ -0,0 +1,802 @@
...
+
+    def update(self, other: Bindings | OperationResult):
+        if isinstance(other, OperationResult):
+            self.bindings.update(other.bindings)
+        else:
+            self.bindings.update(other)
+        return self
+
+    def __hash__(self):
+        return id(self)
```
LucaKro: hmm i dont like that. why not use uuid here as well? 
Id argue with uuid you can be basically 100% certain they wont repeat, with id() you can get very unlucky and get the same memory address. Its still unlikely, but apparently likely enough that we ran into this problem in the beginning of semdt, and these bugs are hell to debug

krrood/src/krrood/entity_query_language/core/base_expressions.py
```diff
@@ -0,0 +1,802 @@
...
+        """
+        Map the result to the correct output data structure for user usage.
+
+        :param result: The result to be mapped.
+        :return: The mapped result.
+        """
+        return result.value
+
+    @property
+    def _original_value_is_iterable_and_this_operation_preserves_that_(self):
```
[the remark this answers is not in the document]
LucaKro: maybe not? idk this is  weird hahaha

LucaKro (changes requested):
okay i dont think i can do more than the first file right now, but well done!

many of my code suggestions are just nitpics, you can choose if you want to implement then, but i had fun trying to optimize :D

Ill see if i can review more tomorrow

krrood/src/krrood/entity_query_language/core/base_expressions.py
```diff
@@ -0,0 +1,802 @@
...
+    Internal attribute used to track the parent symbolic expression of this expression.
+    """
+    _eval_parent_: Optional[SymbolicExpression] = field(
+        default=None, init=False, repr=False
+    )
+    """
+    The current parent symbolic expression of this expression during evaluation. Since a node can have multiple parents,
+    this attribute is used to track the current parent that is being evaluated.
+    """
+    _expression_: Optional[SymbolicExpression] = field(
```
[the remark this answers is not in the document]
AbdelrhmanBassiouny: No, this was because I asked ChatGPT whether I should put Optional when I will definitely set the expression later, and it said that it still should be optional because during its lifetime, it is allowed to be None.

krrood/src/krrood/entity_query_language/core/base_expressions.py
```diff
@@ -0,0 +1,802 @@
...
+    ) -> Iterator[TypingUnion[T, Dict[TypingUnion[T, SymbolicExpression], T]]]:
+        """
+        Evaluate the query and map the results to the correct output data structure.
+        This is the exposed evaluation method for users.
+        """
+        SymbolGraph().remove_dead_instances()
+        results = map(
+            self._process_result_, (res for res in self._evaluate_() if res.is_true)
+        )
+        if self._limit_ is None:
```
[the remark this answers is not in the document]
AbdelrhmanBassiouny: uh nice one

krrood/src/krrood/entity_query_language/core/base_expressions.py
AbdelrhmanBassiouny: That's the least ugly solution that allows me to split the classes into different files. I asked both Juni and ChatGPT about this, and the other solutions they suggested were worse due to having to modify this class from other files, which is imo worse than this.

I think this is a small price to pay to allow for a more organized package. Also apparently many big frameworks do that alot like SQLAlchemy for example.

krrood/src/krrood/entity_query_language/core/base_expressions.py
```diff
@@ -0,0 +1,802 @@
...
+        """
+        return make_set(self._all_variable_instances_)
+
+    @property
+    def _all_variable_instances_(self) -> List[Variable]:
+        """
+        Get the leaf instances of the symbolic expression.
+        This is useful for accessing the leaves of the symbolic expression tree.
+        """
+        from .variable import Variable
```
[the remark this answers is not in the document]
AbdelrhmanBassiouny: same reason :smile:

krrood/src/krrood/entity_query_language/core/base_expressions.py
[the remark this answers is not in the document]
AbdelrhmanBassiouny: removed the method altogether haha

krrood/src/krrood/entity_query_language/core/base_expressions.py
```diff
@@ -0,0 +1,802 @@
...
+
+from ..failures import NoExpressionFoundForGivenID
+from ..utils import make_list, T, make_set, is_iterable
+from ...symbol_graph.symbol_graph import SymbolGraph
+
+if TYPE_CHECKING:
+    from ..rules.conclusion import Conclusion
+    from .variable import Variable
+
+Bindings = Dict[int, Any]
```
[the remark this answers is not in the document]
LucaKro: then the docstring is wrong no?

krrood/src/krrood/entity_query_language/core/base_expressions.py
```diff
@@ -0,0 +1,802 @@
...
+            return
+
+        if value is None and self._parent__ is not None:
+            if self._id_ in [v._id_ for v in self._parent__._children_]:
+                self._parent__._children_.remove(self)
+            self._parents_.remove(self._parent__)
+
+        self._parent__ = value
+
+        if value is not None and value._id_ not in [v._id_ for v in self._parents_]:
```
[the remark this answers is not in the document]
LucaKro: hmm okay

krrood/src/krrood/entity_query_language/core/mapped_variable.py
```diff
@@ -0,0 +1,364 @@
...
+        Generates a hashable key for the given type and arguments.
+
+        :param type_: The type of the mapped variable to generate a key for, e.g., Attribute, Index, etc.
+        :param args: Positional arguments to pass to the MappedVariable constructor.
+        :param kwargs: Keyword arguments to pass to the MappedVariable constructor.
+        :return: The generated hashable key.
+        """
+        args = (self,) + args
+        all_kwargs = merge_args_and_kwargs(type_, args, kwargs, ignore_first=True)
+        return convert_args_and_kwargs_into_a_hashable_key(all_kwargs)
```
LucaKro: remove "a" in my opinion. just "convert_args_and_kwargs_into_hashable_key"

krrood/src/krrood/entity_query_language/rules/conclusion_selector.py
```diff
@@ -7,74 +7,168 @@
...
+
 
 @dataclass(eq=False)
 class ConclusionSelector(TruthValueOperator, ABC):
     """
     Base class for operators that selects the conclusions to pass through from it's operands' conclusions.
     """
 
+    @classmethod
+    def create_and_update_rule_tree(
```
tomsch420: hmm feels weird to create RDRs here if the dependency in the future will be that RDR uses EQL

tomsch420 (changes requested):

krrood/src/krrood/entity_query_language/query/builders.py
AbdelrhmanBassiouny: I see, anyway, I do not actually want to get all possibilities, I just don't want to have an aggregation and a non-aggregation at the same time. I refactored this method now

krrood/src/krrood/entity_query_language/query/builders.py
```diff
@@ -93,21 +96,32 @@ def aggregators_and_non_aggregators_in_conditions(
...
-                if isinstance(var, Aggregator):
-                    aggregators.append(var)
-                elif isinstance(var, MappedVariable) and any(
-                    isinstance(v, Aggregator) for v in var._descendants_
-                ):
-                    aggregators.append(var)
-                elif isinstance(var, Selectable) and not isinstance(var, Literal):
-                    non_aggregators.append(var)
+
+        def walk(expr: SymbolicExpression):
```
tomsch420: hmmm, SQLAlchemy implements stuff like this using a visitor pattern. Is this appropriate here or do we expect to not traverse queries often, so YAGNI?

tomsch420 (approved):

krrood/src/krrood/entity_query_language/query_graph.py
```diff
@@ -0,0 +1,200 @@
...
+        """
+        Adds child nodes to the graph recursively.
+
+        :param parent_node: The parent node of the children to add.
+        """
+        parent_expression = parent_node.data
+        for child in parent_expression._children_:
+            child_node = self.construct_graph(child)
+            if isinstance(parent_expression, Query):
+                if child._binding_id_ in [
```
LucaKro: why not have those in one statement?

krrood/src/krrood/entity_query_language/operators/logical_quantifiers.py
```diff
@@ -0,0 +1,126 @@
...
+    This is the super class of the universal, and existential conditional operators. It is a binary logical operator
+    that has a quantified variable and a condition on the values of that variable.
+    """
+
+    @property
+    def variable(self):
+        return self.left
+
+    @variable.setter
+    def variable(self, value):
```
LucaKro: i dont like that this gives you two ways of setting left and right idk

PR #186, opened 2026-02-16
extending surface sampling of semantic annotations

LucaKro:
Suturo needs a way to get target locations on a supporting surface, which should also be near a similar object.

This serves pretty much the same purpose as the semantic costmap location in pycram. The reason why i didnt didnt include it in one, is that i didnt just want to replace the array based solution, and the existing ProbabilisticSemanticLocation is doing a different and more complex kind of sampling that is worth keeping in my opinion.

I think the CostmapLocations will need some refactor in the future, as they are quite bloated and could be rewritten, and some of the sampling logic can probably move to semdt semantic annotations

### DIFF
--- a/semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
+++ b/semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
@@ -757,7 +757,7 @@ class HasSupportingSurface(HasStorageSpace, ABC):
 
         surface_circuit = self._build_surface_sampler(
             category_of_interest=category_of_interest,
-            object_bloat_and_standard_deviation=largest_xy_object_dimension,
+            object_bloat_and_variance=largest_xy_object_dimension,
         )
 
         if surface_circuit is None:
@@ -771,18 +771,18 @@ class HasSupportingSurface(HasStorageSpace, ABC):
     def _build_surface_sampler(
         self,
         category_of_interest: Optional[Type[SemanticAnnotation]] = None,
-        object_bloat_and_standard_deviation: float = 0.1,
+        object_bloat_and_variance: float = 0.1,
     ):
         """
         Build a probabilistic circuit representing the supporting surface, truncated by the objects on the surface,
         and with Gaussian mixtures around the objects of interest.
 
         :param category_of_interest: The type of object sample points around.
-        :param object_bloat_and_standard_deviation: The amount of bloat to apply to the object events, and the standard
+        :param object_bloat_and_variance: The amount of bloat to apply to the object events, and the standard
             deviation to use for the Gaussian mixtures.
         """
         truncated_event_2d = self._2d_surface_sample_space_excluding_objects(
-            object_bloat_and_standard_deviation
+            object_bloat_and_variance
         )
 
         objects_of_interest = (
@@ -795,7 +795,7 @@ class HasSupportingSurface(HasStorageSpace, ABC):
                 world_P_obj_list=[
                     obj.root.global_pose.to_position() for obj in objects_of_interest
                 ],
-                standard_deviation=object_bloat_and_standard_deviation,
+                variance=object_bloat_and_variance,
                 sample_space=truncated_event_2d,
             )
         else:
@@ -825,14 +825,14 @@ class HasSupportingSurface(HasStorageSpace, ABC):
     def _2d_gaussian_sampler_from_2d_sample_space(
         self,
         world_P_obj_list: List[Point3],
-        standard_deviation: float,
+        variance: float,
         sample_space: Event,
     ) -> Optional[ProbabilisticCircuit]:
         """
         Create a Gaussian mixture model from a list of points, truncated by an event.
 
         :param world_P_obj_list: A list of points representing the positions of the objects to sample around, in the world frame.
-        :param standard_deviation: The standard deviation to use for the Gaussian mixtures.
+        :param variance: The standard deviation to use for the Gaussian mixtures.
         :param sample_space: The event to truncate the Gaussian mixture model with.
 
         :return: A probabilistic circuit representing the Gaussian mixture model truncated by the event, or None if the event has zero measure.
@@ -849,12 +849,12 @@ class HasSupportingSurface(HasStorageSpace, ABC):
             x_p = GaussianDistribution(
                 SpatialVariables.x.value,
                 float(world_P_obj[0]),
-                standard_deviation,
+                variance,
             )
             y_p = GaussianDistribution(
                 SpatialVariables.y.value,
                 float(world_P_obj[1]),
-                standard_deviation,
+                variance,
             )
             p_object_root.add_subcircuit(leaf(x_p, surface_circuit))
             p_object_root.add_subcircuit(leaf(y_p, surface_circuit))


semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
```diff
@@ -570,6 +594,22 @@ def add_object(self, object: HasRootBody):
...
+
+        ..warning:: object_type does not have to be a subclass of HasRootBody, as some semantic concepts, for example
+        Food may not necessarily inherit from HasRootBody, but some objects stored in here may inherit from Food as well
+        as HasRootBody.
+
+        :param object_type: The type of the semantic annotations to return.
+
+        :return: A list of HasRootBody objects of the given type.
+        """
+        return [obj for obj in self.objects if isinstance(obj, object_type)]
```
AbdelrhmanBassiouny: could also be written `variable(object_type, self.objects).tolist()` just saying haha

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
```diff
@@ -665,30 +705,157 @@ def calculate_supporting_surface(
...
+            z_object_dimension = body_to_sample_for.root.combined_mesh.extents[2]
+
+        self_max_z = self.supporting_surface.combined_mesh.bounds[1][2]
         z_coordinate = np.full(
-            (amount, 1), max([b.max_z for b in area_of_table]) + 0.01
+            (amount, 1),
+            self_max_z + (z_object_dimension / 2),
+        )
+
+        surface_circuit = self._build_surface_circuit(
```
AbdelrhmanBassiouny: I get why it's named circuit, but tbh I don't think this name helps compared to something more general about the nature of the returned object that it is a `sampler` or a `probabilisitc_model` so I would prefer `_build_surface_sampler` here as it indicates intent or `_build_surface_probabilistic_model`

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
```diff
@@ -665,30 +705,157 @@ def calculate_supporting_surface(
...
+                world_P_obj_list=[
+                    obj.root.global_pose.to_position() for obj in objects_of_interest
+                ],
+                standard_deviation=object_bloat_and_standard_deviation,
+                event=truncated_event_2d,
+            )
+        else:
+            return uniform_measure_of_event(truncated_event_2d)
+
+    def _2d_event_truncated_by_objects(self, object_event_bloat: float) -> Event:
```
AbdelrhmanBassiouny: Similarly here I also get why it's named event and called truncated (but that's because I am familiar with Tom's stuff) I would actually prefer `_2d_sample_space_of_surface_and_objects`

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
```diff
@@ -665,30 +705,157 @@ def calculate_supporting_surface(
...
+                .bounding_box()
+                .enlarge_all(object_event_bloat)
+                .simple_event.as_composite_set()
+            )
+            object_event_2d = object_event.marginal(SpatialVariables.xy)
+            event_2d = event_2d - object_event_2d
+
+        return event_2d
+
+    def _gaussian_mixture_from_points_truncated_by_event(
```
AbdelrhmanBassiouny: similarly `_2d_gaussian_sampler_from_2d_sample_space`

AbdelrhmanBassiouny (changes requested):
Nice!

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
```diff
@@ -665,30 +705,157 @@ def calculate_supporting_surface(
...
+        self, world_P_obj_list: List[Point3], standard_deviation: float, event: Event
+    ) -> Optional[ProbabilisticCircuit]:
+        """
+        Create a Gaussian mixture model from a list of points, truncated by an event.
+        """
+
+        surface_circuit = ProbabilisticCircuit()
+        surface_circuit_root = SumUnit(probabilistic_circuit=surface_circuit)
+
+        for world_P_obj in world_P_obj_list:
```
tomsch420: I really don't like the name worpd_P_obj_list. I know the guide, but isn't there a semantically more meaningful name available here?

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
```diff
@@ -665,30 +705,157 @@ def calculate_supporting_surface(
...
+                .enlarge_all(object_event_bloat)
+                .simple_event.as_composite_set()
+            )
+            object_event_2d = object_event.marginal(SpatialVariables.xy)
+            event_2d = event_2d - object_event_2d
+
+        return event_2d
+
+    def _gaussian_mixture_from_points_truncated_by_event(
+        self, world_P_obj_list: List[Point3], standard_deviation: float, event: Event
```
tomsch420: Standard deviation is also a very technical name. Why not something like scale? Or spread?

semantic_digital_twin/src/semantic_digital_twin/world_description/geometry.py
```diff
@@ -873,13 +873,16 @@ def enlarge(
...
+    def enlarge_all(self, amount: float) -> Self:
         """
         Enlarge the axis-aligned bounding box in all dimensions by a given amount in-place.
 
         :param amount: The amount to enlarge the bounding box
+
+        :return: The enlarged bounding box (self)
         """
         self.enlarge(amount, amount, amount, amount, amount, amount)
+        return self
```
tomsch420: Hmmm I think returning self on in place methods makes this easier to missuse. If it doesnt return anything its intuitive that's in place

tomsch420 (changes requested):

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
LucaKro: fair :D but i think the list comprehension makes this the clearest

semantic_digital_twin/src/semantic_digital_twin/world_description/geometry.py
LucaKro: hmm fair, ill change it back but i distinctly remember you doing stuff like this before, so im pretty sure we have some instances of this in the codebase

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
```diff
@@ -665,30 +705,157 @@ def calculate_supporting_surface(
...
+
+        largest_xy_object_dimension = 0.1
+        z_object_dimension = 0.0
+        if body_to_sample_for:
+            largest_xy_object_dimension = body_to_sample_for.root.combined_mesh.extents[
+                :2
+            ].max()
+            z_object_dimension = body_to_sample_for.root.combined_mesh.extents[2]
+
+        self_max_z = self.supporting_surface.combined_mesh.bounds[1][2]
```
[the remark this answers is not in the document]
LucaKro: I  put some in shape collection :+1: idk why trymesh does it this way

PR #191, opened 2026-02-16
Add optional synchronous world state publishing with ack protocol

benjaminalt:
Introduce an acknowledgment mechanism to the world state synchronizer so that publishers can optionally block until all subscribers confirm receipt. This is needed for cases where downstream consumers must have an up-to-date world state before proceeding (e.g. motion planning after a state change).

- Add event_id to WorldStateUpdate for tracking publish/ack pairs
- Every Synchronizer now creates an ack topic pub/sub and acknowledges incoming messages, so it can participate as a receiver in the protocol
- Add synchronous flag to SynchronizerOnCallback (default: False); when True, StateSynchronizer.world_callback blocks until all subscribers acknowledge
- Add pytest configuration to pyproject.toml
- Add test_synchronous_state_synchronization test

### DIFF
--- a/test/krrood_test/dataset/ormatic_interface.py
+++ b/test/krrood_test/dataset/ormatic_interface.py
@@ -457,34 +457,6 @@ class GenericClassDAO(
     }
 
 
-class GenericClass_floatDAO(
-    GenericClassDAO,
-    DataAccessObject[test.krrood_test.dataset.example_classes.GenericClass[float]],
-):
-
-    __tablename__ = "GenericClass_floatDAO"
-
-    database_id: Mapped[builtins.int] = mapped_column(
-        ForeignKey(GenericClassDAO.database_id),
-        primary_key=True,
-        use_existing_column=True,
-    )
-
-    value: Mapped[builtins.float] = mapped_column(use_existing_column=True)
-    optional_value: Mapped[typing.Optional[builtins.float]] = mapped_column(
-        use_existing_column=True
-    )
-
-    container: Mapped[typing.List[builtins.float]] = mapped_column(
-        JSON, nullable=False, use_existing_column=True
-    )
-
-    __mapper_args__ = {
-        "polymorphic_identity": "GenericClass_floatDAO",
-        "inherit_condition": database_id == GenericClassDAO.database_id,
-    }
-
-
 class GenericClass_PositionDAO(
     GenericClassDAO,
     DataAccessObject[
@@ -530,6 +502,34 @@ class GenericClass_PositionDAO(
 
     __mapper_args__ = {
         "polymorphic_identity": "GenericClass_PositionDAO",
+        "inherit_condition": database_id == GenericClassDAO.database_id,
+    }
+
+
+class GenericClass_floatDAO(
+    GenericClassDAO,
+    DataAccessObject[test.krrood_test.dataset.example_classes.GenericClass[float]],
+):
+
+    __tablename__ = "GenericClass_floatDAO"
+
+    database_id: Mapped[builtins.int] = mapped_column(
+        ForeignKey(GenericClassDAO.database_id),
+        primary_key=True,
+        use_existing_column=True,
+    )
+
+    value: Mapped[builtins.float] = mapped_column(use_existing_column=True)
+    optional_value: Mapped[typing.Optional[builtins.float]] = mapped_column(
+        use_existing_column=True
+    )
+
+    container: Mapped[typing.List[builtins.float]] = mapped_column(
+        JSON, nullable=False, use_existing_column=True
+    )
+
+    __mapper_args__ = {
+        "polymorphic_identity": "GenericClass_floatDAO",
         "inherit_condition": database_id == GenericClassDAO.database_id,
     }
 
--- a/pycram/scripts/generate_orm.py
+++ b/pycram/scripts/generate_orm.py
@@ -56,6 +56,12 @@ classes, alternative_mappings, type_mappings = get_classes_of_ormatic_interface(
     semantic_digital_twin.orm.ormatic_interface
 )
 
+# filter out test classes that should not be in the production ORM
+classes = [c for c in classes if not c.__module__.startswith("test.")]
+alternative_mappings = [
+    am for am in alternative_mappings if not am.__module__.startswith("test.")
+]
+
 classes = set(classes)
 
 # create of classes that should be mapped
--- a/pycram/src/pycram/datastructures/grasp.py
+++ b/pycram/src/pycram/datastructures/grasp.py
@@ -51,7 +51,9 @@ class GraspDescription:
     The offset between the center of the pose in the grasp sequence
     """
 
-    def _pose_sequence(self, pose: PoseStamped, body: Body = None, reverse: bool = False) -> List[PoseStamped]:
+    def _pose_sequence(
+        self, pose: PoseStamped, body: Body = None, reverse: bool = False
+    ) -> List[PoseStamped]:
         """
         Calculates the pose sequence to grasp something at the pose if the body is given its geometry is also taken into
         account. The pose sequence consists of 3 poses: one in front of the body (taking body geometry into account),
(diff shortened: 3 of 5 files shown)


semantic_digital_twin/src/semantic_digital_twin/adapters/ros/messages.py
```diff
@@ -77,6 +77,11 @@ class WorldStateUpdate(Message):
     Class describing the updates to the free variables of a world state.
     """
 
+    event_id: UUID
+    """
+    UUID of the update (for syn/ack tracking purposes)
```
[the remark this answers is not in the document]
Tigul: That are the correct terms for that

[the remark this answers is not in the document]
LucaKro: ive also thought about this, and I think we should stick to the no abbreviations rule, and even go as far as renaimg "DiffDrive" to "DifferentialDrive". the reason being that once we open the door to domain-specific abbreviations being allowed, we dont really have an argument for not using, for example, abbreviations for QP specific terms inside  giskard and spatial types. I think not using abbreviations is the safe bet here, if you really want you can mention the correct abbreviations in the docstring. I think this is also what @ichumuh did if i remember correctly.

test/semantic_digital_twin_test/test_ros/test_world_synchronizer.py
```diff
@@ -492,6 +463,37 @@ def test_synchronize_6dof(rclpy_node):
...
+    )
+    synchronizer_2 = StateSynchronizer(
+        node=rclpy_node,
+        world=w2,
+    )
+
+    # Allow time for publishers/subscribers to discover each other
+    time.sleep(0.2)
+
+    w1.state.data[0, 0] = 1.0
```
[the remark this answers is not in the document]
benjaminalt: I moved the sync logic up to the Synchronizer base class, so that both world and state can be published synchronously if required. Currently this behavior is off by default. Turning it on was required in the demo malte did for AICOR Solutions, where the synchronization between giskard and pycram kept throwing errors because the worlds were out of sync. Where/how do you think synchronous behaviour should be turned on?

benjaminalt:
Updated PR with syn/ack functionality for all synchronizer types including world, more comprehensive tests and addressed all comments.

LucaKro (changes requested):
There are merge conflicts with the Main, u think it makes sense for you to resolve those before any further review/merge

semantic_digital_twin/src/semantic_digital_twin/adapters/ros/world_synchronizer.py
```diff
@@ -73,40 +110,116 @@ def __post_init__(self):
...
         """
         tracker = WorldEntityWithIDKwargsTracker.from_world(self.world)
-
         msg = self.message_type.from_json(
             json.loads(msg.data), **tracker.create_kwargs()
         )
+
+        self.acknowledge_publisher.publish(
+            std_msgs.msg.String(
+                data=json.dumps(
```
tomsch420: please make a real class in the messages for this instead of a dict

semantic_digital_twin/src/semantic_digital_twin/adapters/ros/world_synchronizer.py
```diff
@@ -73,40 +110,116 @@ def __post_init__(self):
...
+
+            if msg_dic["node_name"] in self._expected_acknowledgments:
+                self._received_acknowledgments.add(msg_dic["node_name"])
+
+            if self._received_acknowledgments == self._expected_acknowledgments:
+                self._acknowledge_condition_variable.notify_all()
+
+    def _snapshot_subscribers(self) -> Set[String]:
+        """
+        Return the node names of all current subscribers to the synchronization topic.
```
tomsch420: proper RST syntax is missing

semantic_digital_twin/src/semantic_digital_twin/adapters/ros/world_synchronizer.py
tomsch420: I think somewhere in the doc of the classes here it should be explicitly warned about delays due to dead synchronous processes. Some hint on how to kill nodes if your process crashes with atexit or something.

semantic_digital_twin/src/semantic_digital_twin/adapters/ros/world_synchronizer.py
```diff
@@ -88,25 +151,104 @@ def meta_data(self) -> MetaData:
...
 
-        msg = self.message_type.from_json(
-            json.loads(msg.data), **tracker.create_kwargs()
+        msg = from_json(json.loads(msg.data), **tracker.create_kwargs())
+
+        acknowledgment = Acknowledgment(
+            publication_event_id=msg.publication_event_id,
+            node_name=self.node.get_name(),
+        )
+        self.acknowledge_publisher.publish(
```
LucaKro: do we want to send an acknowledgement before we know if the metadata and our metadata matches? (aka if we would acknowledge ourselves) 

seems to me like this could cause problems

LucaKro (changes requested):
nice one, I really like this i think!

semantic_digital_twin/src/semantic_digital_twin/adapters/ros/world_synchronizer.py
```diff
@@ -131,6 +273,20 @@ class SynchronizerOnCallback(Synchronizer, Callback, ABC):
     Additionally, ensures that the callback is cleaned up on close.
     """
 
+    synchronous: bool = False
+    """
+    If True, world_callback will block until all subscribers acknowledge receipt of the published message.
+    """
+
+    _skip_next_world_callback: bool = False
```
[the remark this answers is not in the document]
benjaminalt: Sry, I missed that one when merging main. Deleted, as unused

semantic_digital_twin/src/semantic_digital_twin/adapters/ros/world_synchronizer.py
```diff
@@ -88,25 +151,104 @@ def meta_data(self) -> MetaData:
...
+
+    def _snapshot_subscribers(self) -> Set[str]:
+        """
+        Return the node names of all current subscribers to the synchronization topic.
+
+        :return: Set of node name strings for all current subscribers on this synchronizer's topic.
+        """
+        infos = self.node.get_subscriptions_info_by_topic(self.topic_name)
+        expected = set()
+        for info in infos:
```
tomsch420: expected = {info.node_name for info in infos}

test/krrood_test/dataset/ormatic_interface.py
tomsch420: What happened here? 😱

semantic_digital_twin/src/semantic_digital_twin/adapters/ros/world_synchronizer.py
benjaminalt: Excluded the metadata of the message originator from the acknowledgment mechanism (we never acknowledge ourselves). I had built it with self-ack in the beginning for simplicity's sake (saves a few checks and does no harm), but removing self-ack also works of course

semantic_digital_twin/src/semantic_digital_twin/adapters/ros/world_synchronizer.py
```diff
@@ -28,9 +36,26 @@
...
+
+        When ``synchronous=True`` is used on a :class:`SynchronizerOnCallback`, publication
+        blocks until **all** current subscribers acknowledge receipt or a 5-second timeout
+        elapses. If a subscriber process crashes or exits without unsubscribing, the publisher
+        will wait for the full timeout on every synchronous publish because the dead process
+        never acknowledges.
+
+        To mitigate this, always clean up synchronizers when shutting down::
+
+            import atexit
```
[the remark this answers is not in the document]
benjaminalt: done

benjaminalt:
All tests work again.

maltehue (approved):

PR #212, opened 2026-02-27
Fix surface samples

endxrpxl:
The supporting surface was spawned at the origin of its parent body instead of on its surface.
As a result, the sampled points did not match expectations, since the supporting surface they were sampled from was not positioned correctly.

This fix spawns the supporting surface on the surface of its parent body. It also assigns the proper reference frame, as the sampling points are taken from the supporting surface rather than from the root of the parent body.

### DIFF
--- a/semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
+++ b/semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
@@ -702,8 +702,14 @@ class HasSupportingSurface(HasStorageSpace, ABC):
             ),
             points_3d=points_3d,
         )
+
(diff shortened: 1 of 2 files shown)


AbdelrhmanBassiouny (approved):

tomsch420 (changes requested):
Whats the problem with the current implementation and what is this fixing. Provide a description of the pr please

LucaKro:
the semdt tests are failing, there are multiple tests for the sampling that tests different branches. there the reference frame is mismatching after the update. fix the tests the same way we fixed the first one

tomsch420 (changes requested):

tomsch420 (approved):

tomsch420 (approved):

LucaKro (approved):

PR #216, opened 2026-03-03
A GUI for probabilistic models

tomsch420:
A GUI to please micheal. This is vibe coded with junie and QT, so you dont need to review this actually. I will keep working on this when I get new ideas for better visualization but for now this is good enough.
Screenshot from 2026-03-05 16-42-15
Screenshot from 2026-03-05 16-54-45


ichumuh (approved):
good vibes

PR #219, opened 2026-03-05
fixed issue that caused giskard to get slower over time

ichumuh:
- world state trajectory is more efficient 
- giskard doesn't create plot by default


semantic_digital_twin/src/semantic_digital_twin/world_description/world_state.py
```diff
@@ -474,8 +500,8 @@ def values(self) -> Iterator[WorldStateView]:
         :yield: An iterator of `WorldStateView` objects representing the data
```
AbdelrhmanBassiouny: does yield exist in rst?

PR #242, opened 2026-03-19
[PYCRAM] fix pycram examples

AbdElRahmanFarhan:
- Fix some wrong imports inside the examples: 
example of wrong imports:
- from pycram.process_module import simulated_robot  -> from pycram.motion_executor import simulated_robot
- from pycram.datastructures.enums import TorsoState ->  from semantic_digital_twin.datastructures.definitions import TorsoState

- fixed missing reference frame inside PoseStamped
- Removed links inside pr2.srdf inside the self collision because they do not have collision element.

To test I regenerated the tutorials into jupyter notebook and it was working.

### DIFF
--- a/pycram/examples/action_designator.md
+++ b/pycram/examples/action_designator.md
@@ -64,8 +64,6 @@ world = setup_world()
 pr2_view = PR2.from_world(world)
 
 context = Context(world, pr2_view)
-
-
 ```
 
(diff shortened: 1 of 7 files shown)


AbdElRahmanFarhan:
> can this be closed since jonas PR does the same? or what did you discuss

Yes I think it is not required anymore. I will close it

PR #268, opened 2026-04-04
added multi-robot demo

MoAswad:


LucaKro:
please explain what this PR is for

MoAswad:
It is to add a simple multi robot demo for the pr2 and the hsr, where each robot do an action

PR #282, opened 2026-04-15
Partnet Mobility Dataset integration.

tomsch420:


semantic_digital_twin/src/semantic_digital_twin/adapters/partnet_mobility_dataset/generated_semantic_annotations.py
LucaKro, line 1: i dont like that this is a completely distinct branch of semantic annotaitons

semantic_digital_twin/src/semantic_digital_twin/adapters/partnet_mobility_dataset/loader.py
```diff
@@ -0,0 +1,220 @@
...
+    """
+
+    directory: Path = field(
+        default_factory=lambda: Path.home() / "partnet-mobility-dataset"
+    )
+    """
+    The directory where to save the downloaded URDF files into.
+    """
+
+    default_effort_for_limit_tags: float = 100.0
```
LucaKro, line 61: are 100 effort good default values? 😅 seems a bit high

semantic_digital_twin/src/semantic_digital_twin/adapters/partnet_mobility_dataset/loader.py
```diff
@@ -0,0 +1,220 @@
...
+    """
+    The default effort value to use for limit tags in the URDF files.
+    """
+
+    default_velocity_for_limit_tags: float = 100.0
+    """
+    The default velocity value to use for limit tags in the URDF files.
+    """
+
+    def load(self, model_id: int = 179) -> World:
```
LucaKro, line 71: random default

semantic_digital_twin/src/semantic_digital_twin/adapters/partnet_mobility_dataset/semantic_annotations.py
```diff
@@ -0,0 +1,19 @@
...
+
+from typing_extensions import ClassVar, Set
+
+from semantic_digital_twin.world_description.world_entity import (
+    RootedSemanticAnnotation,
+)
+
+
+@dataclass(eq=False)
+class PartNetLabel(RootedSemanticAnnotation):
```
LucaKro: not rooted semantic annotation. HasRootBody. rooted semantic annotation will be removed soon

semantic_digital_twin/requirements.txt
```diff
@@ -32,3 +32,4 @@ psycopg
 plyfile
 huggingface-hub
 vhacdx
+sapien
```
LucaKro: why do you put this here on the one hand, but on the other hand you guard the import in one of the files above

LucaKro (changes requested):

semantic_digital_twin/src/semantic_digital_twin/adapters/partnet_mobility_dataset/loader.py
tomsch420: its a limit. So at most 100 effort. Giang says its fine.

semantic_digital_twin/src/semantic_digital_twin/adapters/partnet_mobility_dataset/loader.py
tomsch420: actually not. The model ids are quite random and this is at least one that works. Unfortunately the ids are not 0-2000

semantic_digital_twin/src/semantic_digital_twin/adapters/partnet_mobility_dataset/semantic_annotations.py
tomsch420: very soon^tm. Since 2 months soon^tm :D

semantic_digital_twin/requirements.txt
tomsch420: its needed for the CI but not needed for the user. Do we have a requirements for the CI?

semantic_digital_twin/src/semantic_digital_twin/adapters/partnet_mobility_dataset/generated_semantic_annotations.py
tomsch420: what do you suggest?

LucaKro: if this is a dataset we are supporting, why not use our semantic annotations in cases where we already have those concepts (chair for example), and extend our datastructures for those that we dont have yet

semantic_digital_twin/src/semantic_digital_twin/adapters/partnet_mobility_dataset/loader.py
LucaKro: this would be 100N and 100Nm respectively...i guess it could be fine since its only in simulation? irl id be cautious to damage motors in the worst case

semantic_digital_twin/src/semantic_digital_twin/adapters/partnet_mobility_dataset/loader.py
LucaKro: :skull:

LucaKro: document that please

semantic_digital_twin/src/semantic_digital_twin/adapters/partnet_mobility_dataset/semantic_annotations.py
LucaKro: doesnt matter. HasRootBody already exists, no reason to use RootedSemanticAnnotation over it

semantic_digital_twin/requirements.txt
LucaKro: i dont think so, maybe you should add it. if the user doesnt need it, it shouldnt be in the requirements

semantic_digital_twin/src/semantic_digital_twin/adapters/partnet_mobility_dataset/loader.py
tomsch420: its still just a max value for control theory. If you put infinity there it would still be fine as the real motor cannot produce it anyways. Its like if i tell you to only run at most 100km/h. It would be dangerous for you to do so but whatever :D

semantic_digital_twin/src/semantic_digital_twin/adapters/partnet_mobility_dataset/generated_semantic_annotations.py
tomsch420: I dont think that the parser should directly produce our annotations. The partnet annotations may mean something else. Actually it is unknown what it means concretely, there is no definition :( Imo, a good solution would be to have a translation from the partnet annotations to the cool semdt annotations. But this will not be this PR and only done when someone actually needs oit.

LucaKro: what do you mean by "may mean something else"? its fine that there is no definition, but looking at the names its often times its pretty clear what is meant. like they have "furniture" and "drawer". this is just duplicated information. and by mapping the labels to the meshes im sure its not hard to find out what is meant by something like RotationBar

tomsch420: i still think its better to have a distinction between the Partnet annotations and the semdt annotations. A converter is imo the better architecture. Anyways, asserting the non partnet annotations is out of scope for this pr. Its like the schema of Sage10k could be directly semdt classes, its just messy to handle then.

tomsch420: Also on a side note here, having the same name in both Partnet and SemDT doesnt imply that we have the same assumptions about those. Hence, there should be a way to get from PartNet label to semdt annotation, but they should not be forced to be the same

LucaKro: For all our semantic annotations that inherit from HasRootBody, the basic assumption that cannot be violated is there is a root body, which holds for partnet as well. all other fields are either optional or computed.
Just because a partnet drawer never has a handle, does not mean that its incompatible with our drawer, which may or may not have a handle. and in my opinion for this to be merged you need to make the argument that there is a clear conflict at any point that we cannot resolve.

For example we can now not do transporting with partnet, because transporting searches for "drawers", and PartNet drawer is not a drawer. so for it to have *any* use to us we already need the conversion layer. and by starting this we commit to always write conversion layers everytime we add a new dataset. and you need to write a conversion for every single concept that may be duplicate in both any 2 branches of semantic annotations.

I think its more practical and more clean to see if we can resolve the conflicts and thus improve the semdt concepts if there are conflcits

LucaKro: okay so we talked about this over lunch, and our disagreement boiled down to us our different perception on if there are likely to be hard constraints violated. I challenged tom to conceptually give me a hard constraint in current semantic annotations that  may reasonably be violated in the partnet dataset. His example was the parent connection of drawer, in which we expect a prismatic connection in the slider, which may not necessarily hold in partnet, thus we would need to check all partnet drawers to confirm that it wont violate it.

I agree then for now the correct way forward is his suggested approach

LucaKro (approved):

ichumuh (approved):

PR #277, opened 2026-04-11
Robot semantic annotation refactor

LucaKro:
# Feature: Robot Semantic Annotation Refactoring

This PR introduces a comprehensive refactoring of the robot semantic annotation architecture, focusing on improved modularity, clearer separation of concerns, and better guidance when creating new AbstractRobots. 
The main changes involve restructuring how robot parts are defined and assembled using factory methods and explicit part aggregation.

### Reasoning

In our push towards using our existing world modification datastructures for merging and copying worlds, it became clear that the existing way of defining robots made it difficult to write them in a way that resulted in a valid, repeatable modification history.
The changes in this PR now enable this by enforcing a clear pattern for how robot parts are created and added to the world, ensuring that all necessary semantic annotations and other properties are added to the world in the correct, repeatable order.

The refactoring provides several practical benefits:
- **Easier robot integration**: New robots can be added by implementing specific capability interfaces
- **Generic AbstractRobot validation**: Added a validation method to check for missing semantic annotations, repeatable modification history, and other common issues, making testing AbstractRobots easier and more robust
- **Better encapsulation**: Robot parts manage their own world registration through factory methods
- **Smaller and more modular methods to implement**: Instead of having to implement one huge `_setup_semantic_annotations()` method, robot developers can implement smaller methods focused on specific parts their robot has. This also makes it less likely the developer forgets to implement something, as they will be guided by the method names and signatures.
- **Enhanced world copying**: Simplified `__deepcopy__` method using model modification blocks

---

### `semantic_digital_twin/robots/robot_parts.py` (New File)

**Created a new module containing all robot part definitions:**

- **`RobotPart`**: Base class for all robot semantic components (arms, sensors, grippers, etc.)
  - Inherits from `HasRootBody` and `AggregatesRobotParts`
  - Provides `create_and_add_to_world()` factory method pattern for safe instantiation
  - Manages automatic world registration and validation
- **Moved classes from `abstract_robot.py`**:
  - `KinematicChain`, `Arm`, `Manipulator`, `Finger`, `ParallelGripper`, `HumanoidGripper`
  - `Sensor`, `FieldOfView`, `Camera`, `Torso`
  - Renamed `Base` to `MobileBase` for clarity
- **Updated all robot parts** to use factory method pattern:
  - Automatic world registration and part aggregation

### `semantic_digital_twin/robots/abstract_robot.py` (Refactored)

**Transformed from semantic annotation container to robot capability interface:**

[...]

PR #287, opened 2026-04-20
Training Environments for PyCRAM.

tomsch420:
- Training Environments
- Sage10k Environment Upgrades
- Sage10k Plans


pycram/src/pycram/robot_plans/actions/core/misc.py
```diff
@@ -76,3 +86,56 @@ def validate(
...
+                    MoveManipulatorAction(
+                        self.target_pose,
+                        self.manipulator,
+                        allow_gripper_collision=False,
+                    ),
+                ]
+            )
+        ).perform()
+
+    def validate_postcondition(self, result: Optional[Any] = None):
```
Tigul: Thats not how post conditions are done

pycram/src/pycram/robot_plans/motions/misc.py
```diff
@@ -23,3 +26,36 @@ def perform(self):
     @property
     def _motion_chart(self):
         pass
+
+
+@dataclass
+class MoveManipulatorMotion(BaseMotion):
```
Tigul, line 32: I think having MoveToolCenterPoint is enough

pycram/src/pycram/robot_plans/training_environment.py
```diff
@@ -0,0 +1,196 @@
...
+    def generate_episodes(self, number_of_actions: int = 10):
+        """
+        Generate episodes until `number_of_actions` have been executed.
+
+        :param number_of_actions: The number of action executions that should be generated.
+        """
+
+        remaining_actions = number_of_actions
+
+        while remaining_actions > 0:
```
Tigul, line 103: that sounds like a for loop

pycram/src/pycram/robot_plans/training_environment.py
```diff
@@ -0,0 +1,196 @@
...
+
+        with simulated_robot:
+            try:
+                plan.perform()
+            except (PlanFailure, StopIteration):
+                ...
+            finally:
+                self.executed_plans.append(plan)
+
+        number_of_executed_variants = len(plan.root.children)
```
Tigul, line 132: plan.root.children are not necessarily all actions

pycram/src/pycram/robot_plans/training_environment.py
```diff
@@ -0,0 +1,196 @@
...
+
+@dataclass
+class MoveToReachTrainingEnvironment(TrainingEnvironment):
+    """
+    Training environment for MoveToReach actions in an empty world using the PR2.
+    """
+
+    action_type = MoveToReach
+
+    def setup_world(self, **kwargs) -> World:
```
Tigul, line 155: I think it would be better to just take a world as input and then define factories for creating the worlds, otherwise there will be a lot of duplicated code

pycram/src/pycram/robot_plans/training_environment.py
```diff
@@ -0,0 +1,196 @@
...
+                parent=localization_body,
+                child=old_root,
+                world=world_with_urdf,
+            )
+            world_with_urdf.add_connection(c_root_bf)
+            c_root_bf.has_hardware_interface = True
+
+        return world_with_urdf
+
+    def setup_plan(self, limit: int = 10, **kwargs) -> UnderspecifiedNode:
```
Tigul, line 182: Does this create enough variety in the data when only executing one plan in one world setup?

Tigul (changes requested):

pycram/src/pycram/robot_plans/motions/misc.py
tomsch420: The signature is different. I made it such that it works without the Arms Enum

pycram/src/pycram/robot_plans/training_environment.py
tomsch420: No, each episode generates differently many steps and i need to keep track of that

pycram/src/pycram/robot_plans/training_environment.py
tomsch420: yeah but its the variations of the action that is trained

pycram/src/pycram/robot_plans/training_environment.py
tomsch420: I agree. However, for that i need to be able to deepcopy that and that is not fully supported until #277 is merged. Also, where would the duplicated code be? This is currently only duplicated with the tests and i dont think that would be a problem.

pycram/src/pycram/robot_plans/training_environment.py
tomsch420: For this specific training environment yes. For more complicated stuff i think that it could be a nice thing to have multiple trainingenvironments for one action are combine them in one class. But i cannot design this top down, so the best way will be discovered when I get there.

PR #273, opened 2026-04-08
[Patterns] Subclass safe generic

AbdelrhmanBassiouny:
This pull request refactors and improves the handling of type resolution and exception management in the class diagrams module. The main changes include moving and enhancing exception classes, introducing a more robust system for resolving generic type hints, and updating usages throughout the codebase to leverage these improvements.

**Exception management improvements:**

* Exception classes related to class diagram errors (`ClassIsUnMappedInClassDiagram`, `MissingContainedTypeOfContainer`, and `CouldNotResolveType`) were moved from `failures.py` to a new `exceptions.py` file, rewritten as dataclasses, and enhanced to provide better error messages and integration with the codebase. [[1]](diffhunk://#diff-ab529afa9e7d5d154e3e685e56afec4c41b633681bd1f28dfb507416615cc73cR1-R68) [[2]](diffhunk://#diff-687b1dedac61b862315d8be08b674a9fc08011289d2e0e26140b79160d7e464aL1-L30)
* All imports and usages of these exception classes were updated to use the new `exceptions.py` location. [[1]](diffhunk://#diff-7a622951955bc9708335498583cd579f63b42114b53f40df61525c95f579958dL39-R44) [[2]](diffhunk://#diff-451b43533a0dce93fda45a201fc661850e147eec6b5c2b74033a124c310a91d8L28-R28) [[3]](diffhunk://#diff-7cab6021303d2c5ab9adb41c611910558a5f4481cadb0ed62d23a8136ed544f7L18-R18)

**Type resolution and type hint handling:**

* The previous `resolve_type` function was removed from `class_diagram.py` and replaced with a new, more robust implementation in `utils.py`. The new system introduces a `TypeHintResolutionResult` dataclass to encapsulate resolution results and provides utility functions for resolving and retrieving type hints with substitutions. [[1]](diffhunk://#diff-7a622951955bc9708335498583cd579f63b42114b53f40df61525c95f579958dL791-L854) [[2]](diffhunk://#diff-acce2ca65146213fcafd1d79414d8694e290b1a39ee00da471efe15724a7e790R92-R194)
[...]

PR #290, opened 2026-04-21
Robot annotation rethink

LucaKro:
This PR introduces a significant architectural shift in how robots and their semantic annotations are handled within the framework. The core motivation was  to improve consistency and expressiveness of robot definition (representation problem), while also making it easier to add new robots adhering to implicit assumptions of the framework, like that semantic annotations depending on each other need to be added to the world in a specific order in order to be valid for synchronization. (Implementation Problem).

# Representation Problem

This PR underwent multiple iterations, notably https://github.com/cram2/cognitive_robot_abstract_machine/pull/277 which was ultimately rejected because some of the choices developers implementing a new robot needs to do seemed arbitrary (for example why there was a HasArms mixin, but no HasEndeffector mixin). 

Since robots can be so versatile in their structure (there are robots with mobile bases, humanoid robots, robots that are basically just two arms etc), I needed to decide between two cases:
1. the first case was to basically give every part of the robot all possible fields that type of robot part could have, no matter if that is actually the case or not, which would result in not so useful type hints
For example to be able to represent Tracy (table with 2 ur10 arms) PR2 (humanoid robot with mobile omnibase) and ICUB3 (bipedal humanoid robot), I would need the definitions like this:
```python
class AbstractRobot(HasMobileBase, HasArms, HasTorso): ...

class MobileBase(HasTorso): ...

class Torso(HasArms): ...
```
so one would have an arms attribute on the AbstractRobot level as well as the Torso level, which either always need to be kept up to date with each other, which is just duplicated information with all its problems, or one is kept at None and the only the one relevant to the robot is updated (Tracy would only update tracy.arms, and PR2 would only update PR2.mobile_base.torso.arms) which is super unintuitive.

2. The second case is for every part to only have the fields that actually relevant to that robot part. For example Tracy has Tracy.arms, but not Tracy.mobile_base, but PR2 doesnt have PR2.arms but does have PR2.mobile_base.torso.arms. This to me made the most sense, as it gives us the ability to accurately describe the structure of any robot, without any duplicate information.
The problem now was, that the static type checkers had no way of providing type hints past the first level, so one could do PR2.mobile_base, but then have no access to any typehints for mobile_base, which is horrible for usability.

due to the existance of generics in python, i judged that the shortcomings of the second case would be easier to solve than dealing with all the duplicate and unnecessary information of the first case. Another way to solve this would be to override the fields for every new class. The reason why i decided against this will be discussed in the implementation problem section.

Case2 obviously requires us to implement a new class for every single robot part, so even for every single finger. But we decided that this is fine, as since every link in a robot has a unique name, we need to describe every finger individually anyways *in some way*, so it may as well be classes.

So at this point, we have solved the representation problem, now the next issue was to solve details of the implementation and making sure its easy to extend without violating any of the implicit assumptions of the framework.

# Implementation Problem

There were several issues apparent in the original implementation when it came to adding new robots:
1. the datastructures are really complex, so it is a steep learning curve to know what to implement and which methods to override etc. While this is not unusual with datastructures, adding a robot is, based on my experience, something that happens so infrequently that it is quite a big ask to spend a considerable amount of time studying the complex robot annotation datastructures, just to add a robot once, and then potentially never again.
2. Even if this hurdle was taken, and a new robot was implemented, it was possible to create a new robot annotation that was valid on the surface, but was not possible to be synchronized because the order in which semantic annotations were added to the world was invalid, which is not trivial to detect. So one needed to know about two rather niche and complex concepts in semdt, in order to implement a valid robot annotation.
The practical consequence was that I ended up either adding each new robot myself, or fixed the ones others added that were not able to be synchronized. Needless to say, those a big red flags that need to be resolved.

## Solving the Synchronization Problems

There is no easy fix for solving the synchronization issues beyond adding the semantic annotations in the correct order. Someone needs to ensure this is done correctly: either the developer implementing the new robot, or the developer writing the underlying datastructure (me). As the former (as previously established) did not work, I decided that I need to automate the process in the background, so that the user implementing the new robot does not even have to think about that.

**This conclusion also formed the basis for many of my later decisions: Add as many guardrails as possible,  such that at basically any step of the way, the IDE will tell the future developer adding a robot exactly what to implement, reduce the code they need to write, and to automate everything else in the background.**

## Flattening the Learning Curve

As mentioned earlier, there are two ways of solving the type hint problem:
1. Overriding fields for subclasses with the correct types.
2. Python Generics

in addition to that, i was looking for a way automatically initialize the fields with the correct classes, without the developer needing to do so explicitly.

Luckily, both options actually provided a way:
1. For the fields solution, I Initially thought just using the default_factory of fields would be enough, but I struggled with finding a nice way to initialize lists with all the necessary components inside (Tracy.arms should be initialized with an instance of TracyLeftArm and TracyRightArm inside). After some research, i found that typing.Annotated could potentially provide ways of giving fields metadata to initialize complex fields like that.
2. For the Generics solution I remembered that my coworker @AbdelrhmanBassiouny had previously added a way of updating the generic types of fields to their concrete types for subclasses, in a feature called SubclassSafeGeneric introduced in https://github.com/cram2/cognitive_robot_abstract_machine/pull/273. I extended it to support multiple generic bases as well as TypeVarTuple in later PRs https://github.com/cram2/cognitive_robot_abstract_machine/pull/299 https://github.com/cram2/cognitive_robot_abstract_machine/pull/333 https://github.com/cram2/cognitive_robot_abstract_machine/pull/343.
Using this feature, I would be able to iterate though all fields of the class, which are updated to the correct type interally, and then initialize them there.

Both approaches have their pros and cons.
1. the pros are that overriding the fields explicitly makes it easy to inspect the fields offline, without needing to go off into subclasses and infer the fields there. also its arguably the most python native way of doing this. 
1. The cons are that the user needs to write quite a lot of boilerplate code to override all the fields, and has no way of easily confirming that all fields were overridden (missing a field or property would mean we will miss the type hints for that field. If we then also want to automatic initialization we would require typing.Annotated, which is something most developers most likely have not used, making each fields more complicated to override than usual. One could argue that its mostly copy and paste from other robot implementations, but also that means that we just add a whole bunch of boilerplate code.

2. the pros are that the user literally only needs to pass on the types to the mixin superclasses, and all typehints are guaranteed to be updated correctly and all classes are instantiated correctly in the background, without the user needing to write any boilerplate code.
2. the cons are that most users havent worked with generics (although i would argue they are easier to use in this cases than typing.Annotated), and that our work in SubclassSafeGeneric is a bit black magic kind of stuff, so it is a bit less python native than the other option.

So in short, the decision here lies between 1. More closer to native python but more boilerplate + typing.Annotated vs 2. Basically no boilerplate but less python native and more black magic

I ended up deciding for 2. As i think SubclassSafeGeneric, while the way we do it is blackmagic-ish, is something that should be supported by python natively, so I wouldnt label the approach itself as an bad, and the upside of being able to abstract basically everything away in the background is too good to ignore.

# Conclusion

Now that everything we could is abstracted away, there are 3 simple rules one needs to follow when implementing a new robot.

1. Implement a new class for every instance of a concept from robot_parts.py you can identify in your robot
2. For every class of a concept identify the concepts that are direct children of it, and add them to the corresponding robot part mixin  as a generic base in the bases of your class. For example for the PR2RightArm, PR2RightGripper is the direct child concept, so it inherits from HasEndEffector[PR2RightGripper]. While the gripper fingers are also children of the arm, they are not direct children of the arm...they are direct children of the gripper, and so on. 
3. For every class, implement all abstract methods and abstract properties that python requires you to. If you think one of the abstract methods doesnt apply, for example that there are no hardware interfaces for a robot part, just implement it as "pass".

Anything else is done in the background, and one can verify the validity of the new robot by creating a world with the robot inside, and doing robot.validate(), and it will confirm that all fields are at least *plausibly* filled, and that the robot can be synchronized without issues


semantic_digital_twin/src/semantic_digital_twin/robots/robot_parts.py
```diff
@@ -0,0 +1,654 @@
...
+)
+
+if TYPE_CHECKING:
+    from semantic_digital_twin.world import World
+
+logger = logging.getLogger("semantic_digital_twin")
+
+
+@dataclass(eq=False)
+class HasRobotParts(ABC):
```
ichumuh, line 84: doc where?

semantic_digital_twin/src/semantic_digital_twin/robots/robot_parts.py
```diff
@@ -0,0 +1,654 @@
...
+                if part:
+                    current_list.append(part)
+                    # Recursive call for nested robot parts
+                    part.setup_robot_part_semantic_annotations()
+
+
+@dataclass(eq=False)
+class AbstractRobotPart(HasRootBody, HasRobotParts, ABC):
+    """
+    Abstract base class for all robot parts.
```
ichumuh, line 196: bisschen useless doc str hier ne?

semantic_digital_twin/src/semantic_digital_twin/robots/robot_parts.py
```diff
@@ -0,0 +1,654 @@
...
+        return dofs
+
+    def validate(self) -> bool:
+        """
+        Validates the robot semantic annotation.
+            The validation process includes:
+            1. Printing out missing fields of any robot part, so that the user can check if they are intentionally left blank.
+            2. Deepcopy the resulting world to ensure that all parts of the robot are initialized in the correct order
+            3. Assert that the copied world is the same as the original world
+            4. Assert that the robot semantic annotation has a default camera.
```
ichumuh: oh fuck doc str is noch alt

ichumuh (changes requested):

LucaKro:
garmi needs to be ported, will do that next

giskardpy/README.md
```diff
@@ -1,6 +1,6 @@
 # Giskardpy
 Giskardpy is an open source library for implementing motion control frameworks.
-It uses constraint and optimization based task space control to control the whole body of mobile manipulators.
+It uses constraint and optimization based task space control to control the whole body of mobile end_effectors.
```
ichumuh: 👀

krrood/src/krrood/adapters/json_serializer.py
```diff
@@ -53,6 +53,11 @@
 """
 
 
+# TODO: Manage to have JSONData as a proper type alias for the JSON typehints, while keeping everything ormaticable @tomsch420
```
ichumuh, line 56: 👀

krrood/src/krrood/adapters/json_serializer.py
```diff
@@ -305,12 +314,10 @@ def _compute_attribute_diff(
...
         if original_value == new_value:
             return None
-        return JSONAttributeDiff(
-            attribute_name=key, added_values=[from_json(new_value, **kwargs)]
-        )
+        return JSONAttributeDiff(attribute_name=key, added_values=[new_value])
 
-    add = [from_json(x, **kwargs) for x in new_value if x not in original_value]
-    remove = [from_json(x, **kwargs) for x in original_value if x not in new_value]
+    add = [x for x in new_value if x not in original_value]
```
ichumuh: xylophone?

krrood/src/krrood/ormatic/custom_types.py
```diff
@@ -59,6 +60,27 @@ def process_result_value(
...
+    """
+    Type decorator for JSONData that stores JSON without automatic deserialization.
+
+    Unlike regular JSON columns which use the engine's custom json_deserializer
+    (that calls from_json()), this type keeps the data as raw JSON dictionaries/lists.
+    This is necessary for fields that should be deserialized later in application code.
+    """
+
+    impl = types.String
+    cache_ok = True
```
ichumuh, line 73: doc where?

giskardpy/README.md
tomsch420: no

krrood/src/krrood/ormatic/custom_types.py
```diff
@@ -59,6 +60,27 @@ def process_result_value(
...
+class JSONDataType(TypeDecorator):
+    """
+    Type decorator for JSONData that stores JSON without automatic deserialization.
+
+    Unlike regular JSON columns which use the engine's custom json_deserializer
+    (that calls from_json()), this type keeps the data as raw JSON dictionaries/lists.
+    This is necessary for fields that should be deserialized later in application code.
+    """
+
+    impl = types.String
```
ichumuh, line 72: implant?

probabilistic_model/requirements.txt
```diff
@@ -11,7 +11,7 @@ N2G
 python-igraph
 rustworkx>=0.16.0
 scikit-learn>=1.6.1
-jax>=0.6.0
+jax<=0.9.2 # fixed until PM can deal with jax 0.10.0+
```
ichumuh: patrick mania?

tomsch420: thats already fixed

semantic_digital_twin/src/semantic_digital_twin/robots/armar7.py
```diff
@@ -1,362 +1,598 @@
...
-            # tip=self._world.get_body_by_name("Index L Tip_link"),
-            tip=self._world.get_body_by_name("Index L Tip_link"),
-            _world=self._world,
+
+@dataclass(eq=False)
+class Armar7LeftGripper(
+    EndEffector,
+    HasFingers[
+        Union[
+            Armar7LeftThumb,
```
tomsch420: superclass time?

semantic_digital_twin/src/semantic_digital_twin/robots/armar7.py
```diff
@@ -1,362 +1,598 @@
...
 
-        self.add_arm(right_arm)
 
-        # Create camera and neck
-        camera = Camera(
-            name=PrefixedName("AzureKinect_RGB_link", prefix=self.name.name),
-            root=self._world.get_body_by_name("AzureKinect_RGB_link"),
-            forward_facing_axis=Vector3(0, 0, 1),
+@dataclass(eq=False)
+class AzureKinectRGB(Camera):
```
tomsch420, line 430: how is this different than an ordinary Kinect?

pycram/src/pycram/plans/failures.py
```diff
@@ -111,13 +111,13 @@ def __post_init__(self):
 
 
 @dataclass
-class ManipulatorDidNotReachTarget(PlanFailure):
+class EndEffectorDidNotReachTarget(PlanFailure):
```
ichumuh, line 114: doc where?

semantic_digital_twin/src/semantic_digital_twin/robots/robot_part_mixins.py
```diff
@@ -0,0 +1,203 @@
...
+    """
+
+    @property
+    def thumb(self) -> GenericFinger:
+        [thumb] = [finger for finger in self.fingers if finger.is_thumb]
+        return thumb
+
+
+@dataclass(eq=False)
+class HasTwoFingers(
```
tomsch420, line 100: can you put some reasoning somwhere on why this is the best choice to have this nested generic stuff?

pycram/src/pycram/robot_plans/motions/base.py
```diff
@@ -10,7 +10,7 @@
 
 from giskardpy.motion_statechart.graph_node import Task
 from pycram.plans.designator import Designator
-from semantic_digital_twin.robots.abstract_robot import AbstractRobot
+from semantic_digital_twin.robots.robot_parts import AbstractRobot
 from ...alternative_motion_mapping import AlternativeMotion
```
ichumuh: abs

semantic_digital_twin/src/semantic_digital_twin/robots/justin.py
```diff
@@ -1,383 +1,486 @@
...
+                robot_root, "right_3middle4"
+            ),
+        )
+
+
+@dataclass(eq=False)
+class JustinRightRingFinger(Finger):
+
+    @property
+    def is_thumb(self) -> bool:
```
ichumuh: finger shouldn't know that it is a thumb

semantic_digital_twin/src/semantic_digital_twin/robots/robot_parts.py
```diff
@@ -0,0 +1,738 @@
...
+                # Requirement: Each semantic annotation type should only be initialized once
+                if any(type(item) is concrete_type for item in self._robot_parts):
+                    continue
+
+                part = (
+                    concrete_type.setup_default_configuration_in_world_below_robot_root(
+                        self.root
+                    )
+                )
+                if part:
```
ichumuh: das if is da voll unnötig

semantic_digital_twin/src/semantic_digital_twin/robots/robot_parts.py
```diff
@@ -0,0 +1,738 @@
...
+
+    @property
+    def degrees_of_freedom_with_hardware_interface(self) -> List[DegreeOfFreedom]:
+        """
+        The number of degrees of freedom of the robot, which is the sum of the degrees of freedom of all its end_effectors.
+        """
+        dofs = []
+        for connection in self.connections:
+            dofs.extend(connection.controlled_dofs)
+        return dofs
```
ichumuh: i think you will get dofs twice if you have mimic joints

semantic_digital_twin/src/semantic_digital_twin/robots/robot_parts.py
```diff
@@ -0,0 +1,738 @@
...
+        return True
+
+    def _setup_velocity_limits(self):
+        """
+        Sets up velocity limits for 1-DOF connections in the robot.
+        """
+        vel_limits = defaultdict(
+            lambda: 1.0,
+        )
+        self.tighten_dof_velocity_limits_of_1dof_connections(new_limits=vel_limits)
```
ichumuh: use proportional scaling by default

semantic_digital_twin/src/semantic_digital_twin/robots/robot_parts.py
```diff
@@ -0,0 +1,738 @@
...
+
+    @property
+    def all_arms(self) -> list[Arm]:
+        return [p for p in self._robot_parts if isinstance(p, Arm)]
+
+    @property
+    def all_sensors(self) -> list[Sensor]:
+        return [p for p in self._robot_parts if isinstance(p, Sensor)]
+
+    def get_torso(self):
```
ichumuh, line 753: inconsistent with the properties above

semantic_digital_twin/src/semantic_digital_twin/world_description/world_entity.py
```diff
@@ -580,102 +611,98 @@ def class_name_tokens(cls) -> Set[str]:
...
-            if world is not None:
-                entities.update(
-                    world.get_kinematic_structure_entities_of_branch(entity)
-                )
+    def _kinematic_structure_entities(
+        self, visited: Set[int]
+    ) -> list[KinematicStructureEntity]:
+        """
+        Returns the kinematic structure entities that are part of this semantic annotation.
+         This is done by iterating over all fields of the semantic annotation and checking if they are kinematic
```
ichumuh: indentation

test/giskardpy_test/test_ros2_stuff/test_integration_hsr.py
```diff
@@ -1,3 +1,4 @@
+import time
```
ichumuh: sus

ichumuh (changes requested):

krrood/src/krrood/adapters/json_serializer.py
LucaKro: as written in the PR, tom told me that he will deal with this in another PR

krrood/src/krrood/ormatic/custom_types.py
LucaKro: doc in superclass

krrood/src/krrood/ormatic/custom_types.py
LucaKro: @ sqlalchemy

probabilistic_model/requirements.txt
```diff
@@ -11,7 +11,7 @@ N2G
 python-igraph
 rustworkx>=0.16.0
 scikit-learn>=1.6.1
-jax>=0.6.0
+jax<=0.9.2 # fixed until `probabilistic model` can deal with jax 0.10.0+
```
tomsch420: fixed already

semantic_digital_twin/src/semantic_digital_twin/robots/robot_parts.py
```diff
@@ -0,0 +1,727 @@
...
+                part = field_type.setup_default_configuration_in_world_below_robot_root(
+                    self.root
+                )
+                setattr(self, field_name, part)
+                # Recursive call to trigger initialization for child part's fields
+                part.setup_robot_part_semantic_annotations()
+
+    def _initialize_list_field(self, field_name: str, item_type: Any):
+        """
+        Helper to initialize all parts matching item_type and append them to a list field.
```
tomsch420, line 160: thats weird. Union means any of those, but you say all of those.

semantic_digital_twin/src/semantic_digital_twin/robots/robot_parts.py
```diff
@@ -0,0 +1,727 @@
...
+        Returns the connections of the kinematic chain.
+        This is a list of connections between the bodies in the kinematic chain
+        """
+        if self.root == self.tip:
+            return [self.root.parent_connection]
+        return self._world.compute_chain_of_connections(self.root, self.tip)
+
+
+@dataclass
+class FieldOfView:
```
tomsch420: looks like it doesnt belong here

semantic_digital_twin/src/semantic_digital_twin/world_description/connections.py
```diff
@@ -575,7 +576,14 @@ def copy_for_world(self, world: World) -> Connection6DoF:
 
 
 @dataclass(eq=False)
-class OmniDrive(ActiveConnection, HasUpdateState):
+class Drive(ActiveConnection, HasUpdateState, ABC):
```
tomsch420: i dont know a better name but when i read Drive i do not immedeatly think of the thing that makes a robot move

tomsch420 (changes requested):
Add links to the robots of the robot description if you can. If i see the robot, its way easier to get what kind of robot this is instead of trying to comprehend the description

semantic_digital_twin/src/semantic_digital_twin/robots/armar7.py
LucaKro: no, because i want the most accurate type hints, and for a superclass i would need to either lose type hints, or add some type hints that are incorrect (if thumb was the only finger that has sensors, either the superclass cannot inherit from hassensors so the typehint is incorrect, or all fingers would inherit from hassensors, which would also be incorrect

semantic_digital_twin/src/semantic_digital_twin/robots/armar7.py
LucaKro: I dont know, i dont know the hardware well enough, but the body name also has azure kinect inside, while other kinects have different names.

I think one could go over all hardware in the robot parts again and make sure everything has the correct specs, but i think thats out of scope for this PR

semantic_digital_twin/src/semantic_digital_twin/world_description/connections.py
LucaKro: renamed to WheeledDrive

semantic_digital_twin/src/semantic_digital_twin/robots/armar7.py
tomsch420: ok

semantic_digital_twin/src/semantic_digital_twin/robots/armar7.py
tomsch420: i see

semantic_digital_twin/src/semantic_digital_twin/robots/mmp_dresden.py
```diff
@@ -0,0 +1,244 @@
...
+                robot_root, "arm_0_gripper_robotiq_85_left_knuckle_link"
+            ),
+            tip=robot_root._world.get_body_in_branch_by_name(
+                robot_root, "arm_0_gripper_robotiq_85_left_finger_tip_link"
+            ),
+        )
+
+
+@dataclass(eq=False)
+class MMPDresdenIndexFinger(Finger):
```
sunava, line 64: Is the Robot really called Dresden?

sunava (approved):

semantic_digital_twin/src/semantic_digital_twin/robots/mmp_dresden.py
LucaKro: yess, at least thats what i was told iirc

semantic_digital_twin/src/semantic_digital_twin/robots/robot_parts.py
LucaKro: Removed union stuff and substituted it with TypeVarTuple, which fits a lot better

semantic_digital_twin/src/semantic_digital_twin/robots/armar7.py
```diff
@@ -1,362 +1,557 @@
...
-                    left_gripper_joints,
-                    [0] * len(left_gripper_joints),
-                )
+    @classmethod
+    def setup_default_configuration_in_world_below_robot_root(
+        cls, robot_root: KinematicStructureEntity
+    ) -> Self:
+        return cls(
+            root=robot_root._world.get_body_in_branch_by_name(
+                robot_root, "Hand L Palm_link"
```
AbdelrhmanBassiouny, line 108: I see you have a naming convention, maybe to ensure consistency, and make it harder for the user to write it wrong, you can make a template string that is used here as a fucntion that you pass the changeable values to, so for example:

```python
def construct_name(parent_link_name: str, child_link_name, direction: Optional[Enum] = None):
    if direction:
        return f"{parent_link_name} {direction} {child_link_name}"
    else:
        return f"{parent_link_name} {child_link_name}"
```

semantic_digital_twin/src/semantic_digital_twin/robots/robot_part_mixins.py
```diff
@@ -0,0 +1,298 @@
...
+
+    @property
+    def arm(self) -> TGenericArm:
+        [arm] = self.arms
+        return arm
+
+
+@dataclass(eq=False)
+class HasLeftRightArm(
+    Generic[TGenericLeftArm, TGenericRightArm],
```
AbdelrhmanBassiouny: Why this Generic base when there's already HasArms?

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
```diff
@@ -515,6 +527,19 @@ def add_slider(
         )
         self.slider = slider
 
+    def _kinematic_structure_entities(
```
AbdelrhmanBassiouny, line 530: This function is exactly the same as the one in HasHinge; just replace `slider` with `hinge.` Could that duplication be avoided?

test/pycram_test/test_designator/test_multi_robot_action_designator.py
```diff
@@ -79,13 +80,14 @@
...
     scope="session",
     params=[
-        pytest.param(
-            "garmi",
-            marks=pytest.mark.skipif(
-                Garmi is None,
-                reason="GARMI semantic annotation not installed",
-            ),
-        ),
+        # Garmi commented out until we get access to the robot description in CI
```
AbdelrhmanBassiouny: maybe mark with TODO

AbdelrhmanBassiouny (commented):
Very nice work, just a few comments.

semantic_digital_twin/src/semantic_digital_twin/robots/armar7.py
LucaKro: those are not my names, they are taken from the urdf, i dont have any influence on that :smiling_face_with_tear:

semantic_digital_twin/src/semantic_digital_twin/robots/robot_part_mixins.py
LucaKro: because hasarms has a typevartuple so it accepts an unspecified amount of generics, but for hasleftrightarm we want exactly two, and i need the concrete typing for the left and right arm

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
LucaKro: hmm yea maybe, but im unsure rn how id abstract the attribute access (so self.slider or self.hinge), without having the user needing to hardcode it. I think this will be solved once we find a shared superclass for the two, which i think will happen. Unless this is important to you, id like to say this is out of scope for this PR

semantic_digital_twin/src/semantic_digital_twin/robots/armar7.py
AbdelrhmanBassiouny: I see

semantic_digital_twin/src/semantic_digital_twin/robots/robot_part_mixins.py
AbdelrhmanBassiouny: Hmm, idk feels duplicate information, but I guess you have tried it out.

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
AbdelrhmanBassiouny: no if it was important I would have selected reequest changes instead of just comment haha.

semantic_digital_twin/src/semantic_digital_twin/robots/robot_part_mixins.py
LucaKro: fixed

semantic_digital_twin/src/semantic_digital_twin/robots/robot_part_mixins.py
LucaKro: @tomsch420 I updated the PR description with a detailed account of my reasoning. Please review it (if possible @sunava @ichumuh @AbdelrhmanBassiouny as well).
Let me know if you think i should document this somewhere else, for example in the codebase or the documentation page if you think this is valuable

AbdelrhmanBassiouny (approved):

semantic_digital_twin/src/semantic_digital_twin/robots/robot_part_mixins.py
tomsch420: document it in the Abstract Robot class also on how to do it and why to do it this way

tomsch420 (approved):
Will there be a docpage for how to add a robot in the user guide?

semantic_digital_twin/src/semantic_digital_twin/robots/robot_part_mixins.py
LucaKro: done

LucaKro:
> Will there be a docpage for how to add a robot in the user guide?

yes there will be, its in the works. do you want me to add this in this PR, or can it be another one so this can be merged?

PR #296, opened 2026-04-27
Update requirements.txt

Sanic:
test/pycram_test/test_memory_leak.py and test/conftest.py use objgraph but it's not in requirements.txt yet

### DIFF
--- a/pycram/requirements.txt
+++ b/pycram/requirements.txt
@@ -7,6 +7,7 @@ typing_extensions>=4.10.0
 probabilistic_model>=7.1.0 # just until semantic world is integrated
 random_events>=4.1.18
 numpy
+objgraph
 pint>=0.21.1
 transforms3d
 python-box


Tigul:
We only need objgraph for these special tests and not really for the project itself. So I would not put them in the regular requirements but requirements-test instead

Sanic:
@Tigul Executing (e.g.) robokudo tests will automatically load the general test/conftest.py where objgraph is imported and autoused. So other tests are affected as well by default.

What do you mean with requirements-test?
A new, separate requirements file?
Or should it be added to https://github.com/cram2/cognitive_robot_abstract_machine/blob/main/.github/docker/setup_ros_workspace.sh ?

In the CI, objgraph is currently installed due to the Dockerfile. Therefore, dev/desktop installs miss the objgraph package with our install instructions.

tomsch420 (approved):

sunava (approved):

PR #297, opened 2026-04-27
Body Motion Problem and Giskard Pouring Task via Differential Equations

maltehue:
## **Body Motion Problem (BMP) Implementation for Container Manipulation & Pouring**

This PR introduces an implementation of the Body Motion Problem framework (https://arxiv.org/abs/2602.06572) for manipulating articulated containers and pouring liquids.

### **Core Framework**

- **BMP Types** (`pycram/body_motion_problem/types.py`): Foundational data structures for Effects, TaskRequests, PhysicsModels, and Motions
- **BMP Predicates** (`pycram/body_motion_problem/predicates.py`): Three independent checks for manipulation success:
  - **SatisfiesRequest**: Semantic correctness of the intended outcome
  - **Causes**: Causal sufficiency that motion physically produces the effect
  - **CanPerform**: Embodiment feasibility for robot execution

### **Container Manipulation Domain**

- Physics model (`RunMSCModel`) that executes MotionStatecharts to simulate container opening/closing
- Domain-specific effects (`OpenedEffect`, `ClosedEffect`) for articulated containers
- Container feasibility predicate (`ContainerCanPerform`) that verifies robot can approach and follow handle trajectories with collision avoidance

### **Pouring Domain**

- Differential equations framework:
  - `DifferentialEquation` base class
  - `ArticulatedPouringEquation`: 2D rectangular cup model coupling tilt to fill-level dynamics
  - `InflowEquation`: Receiver container fill from external sources
- **Giskard pouring task** (`PouringTask`): QP-based tilt control driven by fill-level setpoints using differential equations
- Physics model (`PouringMSCModel`) that integrates fill-level ODE and records trajectories
- Pouring feasibility predicate (`PouringCanPerform`) verifying robot can execute cup tilt trajectories

### **Semantic Annotations**

- `HasContainerGeometry` mixin: Interior dimension extraction from collision geometry
- `HasFillLevel` mixin: Virtual `LiquidConnection` DOF encoding fill level in [0, 1], with integrated ODE solvers and pouring equation support
- `LiquidConnection`: New connection type for virtual fill-level DOFs with automatic state updates

### **Additional Changes**

- Updated QP adapter to handle both scalar and array-based constraint bounds/slack limits
- Test suite covering predicate semantics, physics simulation, EQL queries, and robot integration


giskardpy/src/giskardpy/motion_statechart/tasks/pouring.py
```diff
@@ -0,0 +1,91 @@
...
+from semantic_digital_twin.world_description.world_entity import Body
+
+
+@dataclass(eq=False, repr=False)
+class PouringTask(Task):
+    """
+    Motion Statechart task for controlling the tilt and fill level of a container.
+    """
+
+    fill_equation: PouringEquation
```
ichumuh: why does this have to be a parameter? isn't it always the same equation?

giskardpy/src/giskardpy/motion_statechart/tasks/pouring.py
```diff
@@ -0,0 +1,91 @@
...
+    """Pouring ODE coupling tilt to the fill-level DOF."""
+
+    fill_connection: LiquidConnection
+    """Virtual DOF whose position encodes fill level in [0, 1]."""
```
ichumuh: can't this be inferred from the container body?

giskardpy/src/giskardpy/motion_statechart/tasks/pouring.py
```diff
@@ -0,0 +1,91 @@
...
+
+    root_link: Body = field(kw_only=True)
+    """Root of the kinematic chain used to derive the cup tilt expression."""
+
+    tip_link: Body = field(kw_only=True)
+    """Tip of the kinematic chain (the cup body)."""
+
+    goal_value: float
+    """Target fill level to achieve."""
```
ichumuh: whats the unit?

semantic_digital_twin/src/semantic_digital_twin/physics/pouring_equations.py
```diff
@@ -0,0 +1,141 @@
...
+    """
+    Symbolic tilt angle of a cup about the vertical axis given its FK transform.
+
+    Uses the z-component of the cup's local up axis in the root frame:
+    θ = acos(R_zz).
+
+    :param root_T_cup: Symbolic FK expression from root to cup frame.
+    :return: Symbolic tilt angle in radians ∈ [-π, π].
+    """
+    cup_z_in_root = root_T_cup.to_rotation_matrix() @ Vector3(0, 0, 1)
```
ichumuh: Vector3.Z()

semantic_digital_twin/src/semantic_digital_twin/physics/pouring_equations.py
```diff
@@ -0,0 +1,141 @@
...
+    return sm.acos(sm.limit(cup_z_in_root.z, -3.14, 3.14))
```
ichumuh: you sure about that? acos is only defined for -1 and 1. there is sm.safe_acos

giskardpy/src/giskardpy/motion_statechart/tasks/pouring.py
```diff
@@ -0,0 +1,91 @@
...
+    def on_tick(
+        self, context: MotionStatechartContext
+    ) -> Optional[ObservationStateValues]:
+        """
+        Checks if the goal fill level has been reached and that the outflow is zero.
+
+        :param context: The runtime context.
+        :return: The observation state.
+        """
+        fill = float(self.fill_connection.position)
```
ichumuh: pretty sure you can define that as a symbolic expression and use artifacts.observation_expression instead

giskardpy/src/giskardpy/motion_statechart/exceptions.py
```diff
@@ -24,6 +24,7 @@ def __post_init__(self):
         self.message = f"Violated collision constraints: \n"
         for collision, threshold in zip(self.violated_collisions, self.thresholds):
             self.message += f"{str(collision.body_a.name), str(collision.body_b.name)}: {collision.distance} < {threshold}\n"
+        super().__post_init__()
```
ichumuh: please add that to the other expections too, which set self.message but don't call the super

pycram/src/pycram/body_motion_problem/container_manipulation/physics.py
```diff
@@ -0,0 +1,71 @@
...
+
+@dataclass
+class RunMSCModel(PhysicsModel):
+    """
+    Physics model that executes a MotionStatechart against a World.
+
+    The MotionStatechart must be fully parameterized before being passed in.
+    This model compiles it against the provided World, steps it tick-by-tick
+    while recording the actuator position after each tick, and resets the World
+    state before returning.
```
ichumuh: motion statecharts already have a trajectory recording method

pycram/src/pycram/body_motion_problem/container_manipulation/physics.py
```diff
@@ -0,0 +1,71 @@
...
+        :param world: The world to simulate in (state is reset before returning).
+        :return: (trajectory, achieved) where trajectory is the list of actuator positions
+                 recorded at each tick, and achieved indicates whether effect.is_achieved().
+        """
+        if self.msc is None:
+            raise MissingMotionStatechartError(
+                "RunMSCModel requires a MotionStatechart instance to run."
+            )
+        trajectory: List[float] = []
+        achieved = self._run_msc(
```
ichumuh: feels like duplication with the normal executor?

pycram/src/pycram/body_motion_problem/container_manipulation/predicates.py
```diff
@@ -0,0 +1,180 @@
...
+            target = self._resolve_target()
+            trajectory = self._compute_body_trajectory(target)
+        return target, trajectory
+
+    def _resolve_target(self) -> Any:
+        """
+        Resolve the handle body from the motion model or via EQL query.
+        """
+        if self.motion.motion_model:
+            return self.motion.motion_model.msc.nodes[0].tip_link
```
ichumuh: this check goes too deep for my taste

ichumuh: and [0] is fragile

pycram/src/pycram/body_motion_problem/container_manipulation/predicates.py
```diff
@@ -0,0 +1,180 @@
...
+                )
+            ).evaluate()
+        )[0].handle
+
+    def _compute_body_trajectory(self, target: Any) -> list:
+        """
+        Convert the actuator-space trajectory to a sequence of handle poses in world space.
+        """
+        handle_trajectory = []
+        reasoning_world = deepcopy(target._world)
```
ichumuh: i hope this doesn't cause memory leaks. like if the predicate stays alive forever due to some eql reasons

pycram/src/pycram/body_motion_problem/container_manipulation/predicates.py
```diff
@@ -0,0 +1,180 @@
...
+        reasoning_body = reasoning_world.get_body_by_name(target.name)
+        actuator_dof_id = reasoning_world.get_degree_of_freedom_by_name(
+            self.motion.actuator.name.name
+        ).id
```
ichumuh: doesn't the actuator know its dof?

pycram/src/pycram/body_motion_problem/container_manipulation/predicates.py
```diff
@@ -0,0 +1,180 @@
...
+
+        for position in self.motion.trajectory:
+            reasoning_world.state[actuator_dof_id].position = position
```
ichumuh: is trajectory always only 1 dof?

pycram/src/pycram/body_motion_problem/container_manipulation/predicates.py
```diff
@@ -0,0 +1,180 @@
...
+            reasoning_world.notify_state_change()
+            handle_trajectory.append(reasoning_body.global_pose)
+        return handle_trajectory
+
+    def _build_collision_rules(self, gripper: Any, target: Any) -> list:
```
ichumuh: any?

pycram/src/pycram/body_motion_problem/container_manipulation/predicates.py
```diff
@@ -0,0 +1,180 @@
...
+        handle_bodies = [target] if isinstance(target, Body) else list(target.bodies)
+        # Avoiding collision with the gripper and the whole apartment makes execution way faster than with only the handle
+        # Future improvement: avoid collision with the gripper apartment parts not related to the gripper.
+        return [
+            AvoidExternalCollisions(robot=self.robot),
```
ichumuh: don't think you need this rule, should be part of the default rules for the robot

pycram/src/pycram/body_motion_problem/container_manipulation/predicates.py
```diff
@@ -0,0 +1,180 @@
...
+            AllowCollisionBetweenGroups(
+                body_group_a=[b for b in gripper.bodies if b.has_collision()],
+                body_group_b=[
+                    b
+                    for b in self.robot._world.bodies
+                    if "apartment" in str(b.name) and b.has_collision()
```
ichumuh: hard coded str

pycram/src/pycram/body_motion_problem/container_manipulation/predicates.py
```diff
@@ -0,0 +1,180 @@
...
+                ],
+            ),
+        ]
+
+    def _build_msc(
+        self, root: Any, gripper: Any, target: Any, trajectory: list
```
ichumuh: any

pycram/src/pycram/body_motion_problem/container_manipulation/predicates.py
```diff
@@ -0,0 +1,180 @@
...
+        point = Pointing(
+            root_link=root,
+            tip_link=self.robot.root,
+            pointing_axis=pointing_axis,
+            goal_point=goal_point,
+            threshold=0.2,
+        )
+        msc.add_node(point)
+
+        approach_sequence = self._build_cartesian_waypoint_sequence(
```
ichumuh: sounds like this method should be a msc template instead?

pycram/src/pycram/body_motion_problem/container_manipulation/predicates.py
```diff
@@ -0,0 +1,180 @@
...
+                reference_frame=gripper.tool_frame, child_frame=gripper.tool_frame
+            ),
+        )
+        msc.add_node(keep_relation)
+
+        approach_sequence.start_condition = point.observation_variable
+        full_sequence.start_condition = approach_sequence.observation_variable
+        keep_relation.start_condition = approach_sequence.observation_variable
+        approach_sequence.end_condition = approach_sequence.observation_variable
+        point.end_condition = point.observation_variable
```
ichumuh: its been a looong time since i had to set the conditions manually. can't you achieve the same with sequences and parallels? less fehleranfällig

pycram/src/pycram/body_motion_problem/pouring/physics.py
```diff
@@ -0,0 +1,129 @@
...
+from semantic_digital_twin.world import World
+from semantic_digital_twin.world_description.connections import (
+    Connection,
+    LiquidConnection,
+)
+from semantic_digital_twin.world_description.world_entity import Body
+
+
+@dataclass
+class PouringMSCModel(PhysicsModel):
```
ichumuh: imagine the same comments as with the other physicsmodel

pycram/src/pycram/body_motion_problem/pouring/predicates.py
```diff
@@ -0,0 +1,75 @@
...
+from semantic_digital_twin.collision_checking.collision_rules import (
+    AllowCollisionBetweenGroups,
+    AvoidExternalCollisions,
+)
+from pycram.body_motion_problem.predicates import CanPerform
+from semantic_digital_twin.world_description.world_entity import Body
+
+
+@dataclass
+class PouringCanPerform(CanPerform):
```
ichumuh: looks like a lot of duplicated code

pycram/src/pycram/body_motion_problem/predicates.py
```diff
@@ -0,0 +1,284 @@
...
+@dataclass
+class Causes(Predicate):
+    """
+    Causal sufficiency predicate.
+
+    Checks whether a motion trajectory physically produces the desired world-state
+    change under the scoped physics model. The check replays the trajectory in a
+    sandboxed copy of the world and tests whether the effect is achieved at the end.
+
+    If no trajectory is available but a physics model is attached to the motion,
```
ichumuh: what is the purpose for all these trajectories?

pycram/src/pycram/body_motion_problem/predicates.py
```diff
@@ -0,0 +1,284 @@
...
+
+        :param poses: Ordered list of goal poses for the tip link.
+        :param root_link: Root link for the Cartesian goal.
+        :param tip_link: Tip link (e.g. gripper tool frame) to move.
+        :param name_prefix: Prefix for individual waypoint node names.
+        :param sequence_name: Name of the resulting Sequence node.
+        :param threshold: Position tolerance for each waypoint.
+        :return: A Sequence node ready to be added to a MotionStatechart.
+        """
+        waypoints = [
```
ichumuh: yeah this should be a template

pycram/src/pycram/body_motion_problem/types.py
```diff
@@ -0,0 +1,225 @@
...
+    intent.
+    """
+
+    target_object: SemanticAnnotation
+    """The object being affected."""
+
+    property_getter: Callable[[SemanticAnnotation], float]
+    """A callable that reads the relevant property from the target object."""
+
+    goal_value: float
```
ichumuh: this looks a lot like a MSC monitor or maybe also the events from segmind

pycram/src/pycram/body_motion_problem/types.py
```diff
@@ -0,0 +1,225 @@
...
+        :param world: The world to simulate in (state is reset after simulation).
+        :return: (trajectory, achieved) where trajectory is the recorded actuator
+                 positions and achieved indicates whether the effect was satisfied.
+        """
+
+    def build_secondary_trajectories(
+        self, effect: Effect
+    ) -> List[Tuple[Connection, List[float]]]:
+        """
+        Return secondary actuator trajectories recorded by the most recent run().
```
ichumuh: what does secondary mean?

pycram/src/pycram/body_motion_problem/types.py
```diff
@@ -0,0 +1,225 @@
...
+    """Task type identifier (e.g., 'open', 'close', 'pour', 'grasp')."""
+
+    name: str
+    """Name identifying the task or target object."""
+
+    goal: Callable[[Effect], bool] = field(compare=False)
+    """Predicate that checks whether an Effect satisfies this task."""
+
+
+class PhysicsModel(ABC):
```
ichumuh: should this be an executor?

pycram/src/pycram/body_motion_problem/types.py
```diff
@@ -0,0 +1,225 @@
...
+    motion_model: Optional[PhysicsModel] = field(default=None)
+    """Optional physics model used to generate the trajectory when it is empty."""
+
+    secondary_trajectories: List[Tuple[Connection, List[float]]] = field(
+        default_factory=list
+    )
+    """
+    Additional coupled actuator trajectories replayed alongside the primary.
+
+    Each entry is ``(connection, positions)`` where ``positions`` is parallel
```
ichumuh: if you need to write doc strings like this, you should probably write classes for this

semantic_digital_twin/src/semantic_digital_twin/collision_checking/collision_groups.py
```diff
@@ -128,6 +135,7 @@ def update_collision_groups(self, world: World):
             group for group in self.collision_groups if len(group.bodies) > 0
         ]
 
+    @lru_cache
```
ichumuh: you need to use @Tigul new memoize, cause this lru cache will likely cause memory leaks

semantic_digital_twin/src/semantic_digital_twin/physics/equations/pouring_equations.py
```diff
@@ -0,0 +1,141 @@
...
+        self, tilt_expression: Scalar, fill_expression: Scalar
+    ) -> Scalar:
+        """
+        :param tilt_expression: Symbolic tilt angle θ in radians.
+        :param fill_expression: Symbolic fill level in [0, 1].
+        :return: Symbolic d(fill_normalized)/dt as a CasADi expression.
+        """
+        A = self.container_geometry.height
+        r = self.container_geometry.half_width
+        h_sym = fill_expression * A
```
ichumuh: better variable names

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
```diff
@@ -1028,3 +1040,119 @@ def _create_container_event(cls, scale: Scale, wall_thickness: float) -> Event:
...
+    """Mixin for semantic annotations whose collision geometry defines a pourable container interior."""
+
+    @property
+    def container_geometry(self) -> ContainerGeometry:
+        """Interior dimensions derived from the body's collision geometry."""
+        return ContainerGeometry.from_body(self.root)
+
+
+@dataclass(eq=False)
+class HasFillLevel(HasContainerGeometry):
```
ichumuh: @LucaKro has to look at this

ichumuh (changes requested):

test/krrood_test/dataset/ormatic_interface.py
```diff
@@ -677,6 +681,56 @@ class GenericClassAssociationDAO(
     )
 
 
+class HomogeneousTransformationMatrixMappingDAO(
```
LucaKro: this seems very sus, why does krrood depend on semdt in an way?

semantic_digital_twin/src/semantic_digital_twin/world_description/geometry.py
```diff
@@ -872,6 +873,32 @@ def _from_json(cls, data: Dict[str, Any], **kwargs) -> Self:
         )
 
 
+@dataclass
+class ContainerGeometry:
```
LucaKro: imo just add the following properties to ShapeCollection:

```python
    @property
    def depth(self) -> float:
        """
        :return: The depth of the bounding box containing all shapes in this collection.
        """
        return self.scale.x
    
    @property
    def width(self) -> float:
        """
        :return: The width of the bounding box containing all shapes in this collection.
        """
        return self.scale.y
    
    @property
    def height(self) -> float:
        """
        :return: The height of the bounding box containing all shapes in this collection.
        """
        return self.scale.z
```

semantic_digital_twin/src/semantic_digital_twin/world_description/connections.py
```diff
@@ -1145,3 +1151,58 @@ def copy_for_world(self, world: World) -> DifferentialDrive:
...
+    Translating DOF representing the fill level of a container.
+
+    Integrates :attr:`outflow_equation` and :attr:`inflow_equation` each tick.
+    Either may be ``None``; the net velocity is their sum.
+    """
+
+    outflow_equation: Optional[PouringEquation] = field(
+        default=None, kw_only=True, init=False
+    )
+    """ODE governing how liquid leaves this container (e.g. tilting to pour)."""
```
LucaKro: what does the Oxford Dictionary of English have to do with this? no abbreviations 😆

giskardpy/src/giskardpy/motion_statechart/tasks/pouring.py
```diff
@@ -0,0 +1,91 @@
...
+    """Root of the kinematic chain used to derive the cup tilt expression."""
+
+    tip_link: Body = field(kw_only=True)
+    """Tip of the kinematic chain (the cup body)."""
+
+    goal_value: float
+    """Target fill level to achieve."""
+
+    tolerance: float
+    """Acceptance band around goal_value."""
```
LucaKro: maybe use "threshold" instead of "band", as thats the term we use for similar concepts everywhere else

giskardpy/src/giskardpy/qp/adapters/qp_adapter.py
```diff
@@ -762,7 +772,12 @@ def eq_derivative_constraint_bounds(
                 derivative
             ):
                 if t < self.control_horizon:
```
LucaKro: invert and early continue

pycram/src/pycram/body_motion_problem/container_manipulation/physics.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+class MissingMotionStatechartError(Exception):
+    """
+    Raised when RunMSCModel is asked to run without a configured MotionStatechart.
+    """
+
+
+@dataclass
+class RunMSCModel(PhysicsModel):
```
LucaKro: i think this should be in giskard, and and the physics model should be an abstract concept in semdt

pycram/src/pycram/body_motion_problem/container_manipulation/physics.py
```diff
@@ -0,0 +1,71 @@
...
+    msc: MotionStatechart
+    """The fully parameterized (but uncompiled) MotionStatechart to execute."""
+
+    actuator: Connection
+    """The connection whose position is recorded as the trajectory."""
+
+    timeout: int = 500
+    """Maximum number of ticks before aborting the rollout."""
+
+    def run(self, effect: Effect, world: World) -> Tuple[Optional[List[float]], bool]:
```
LucaKro: common datastructure for "Tuple[Optional[List[float]], bool]"

pycram/src/pycram/body_motion_problem/container_manipulation/predicates.py
```diff
@@ -0,0 +1,180 @@
...
+                            HasType(drawer, Drawer),
+                            drawer.root.parent_connection == self.motion.actuator,
+                        ),
+                        and_(
+                            HasType(drawer, Door),
+                            drawer.root.parent_connection == self.motion.actuator,
+                        ),
+                    )
+                )
+            ).evaluate()
```
LucaKro: i think instead of list(query.evaluate()) you can just to query.tolist()

pycram/src/pycram/body_motion_problem/container_manipulation/predicates.py
```diff
@@ -0,0 +1,180 @@
...
+        )[0].handle
+
+    def _compute_body_trajectory(self, target: Any) -> list:
+        """
+        Convert the actuator-space trajectory to a sequence of handle poses in world space.
+        """
+        handle_trajectory = []
+        reasoning_world = deepcopy(target._world)
+        reasoning_body = reasoning_world.get_body_by_name(target.name)
```
LucaKro: please use id for this

pycram/src/pycram/body_motion_problem/container_manipulation/predicates.py
```diff
@@ -0,0 +1,180 @@
...
+        actuator_dof_id = reasoning_world.get_degree_of_freedom_by_name(
```
LucaKro: self.motion.actuator.id should be the same

pycram/src/pycram/body_motion_problem/container_manipulation/predicates.py
LucaKro: yea consider some other way of alloweing those collisions, like only getting bodies which are not downstream from a robot root

pycram/src/pycram/body_motion_problem/pouring/effects.py
```diff
@@ -0,0 +1,42 @@
...
+if TYPE_CHECKING:
+    from semantic_digital_twin.semantic_annotations.mixins import HasFillLevel
+
+
+@dataclass(eq=False, kw_only=True)
+class PouringEffect(MonotoneDecreasingEffect):
+    """
+    Effect achieved when the fill level of a container drops to or below ``goal_value``.
+
+    The target object must be a :class:`~semantic_digital_twin.physics.pouring_equations.HasFillLevel`.
```
LucaKro: docstring wrong

pycram/src/pycram/body_motion_problem/predicates.py
```diff
@@ -0,0 +1,284 @@
...
+    sandboxed copy of the world and tests whether the effect is achieved at the end.
+
+    If no trajectory is available but a physics model is attached to the motion,
+    the model is first used to generate a trajectory from the current world state.
+
+    Returns ``False`` if the effect is already achieved before the motion, treating
+    it as a no-op rather than a success.
+    """
+
+    effect: Effect
```
LucaKro: docs

pycram/src/pycram/body_motion_problem/predicates.py
LucaKro, line 1: I think this also belongs in semdt

pycram/src/pycram/body_motion_problem/types.py
LucaKro: i dont think it should be. this should be an abstract concept in semdt, which can also be implemented for different backends. 
that way this also looses its giskard dependency, and giskard can just implement its own "physicsmodel" with giskard specific stuff

pycram/src/pycram/body_motion_problem/types.py
LucaKro: 👆
typehints like this:
"List[Tuple[Connection, List[float]]]"

are also hints that there may be a class you need to write. check out https://testing.googleblog.com/2017/11/obsessed-with-primitives.html

pycram/src/pycram/body_motion_problem/types.py
```diff
@@ -0,0 +1,225 @@
...
+    to ``trajectory``.
+    """
+
+    dt: Optional[float] = field(default=None)
```
LucaKro: what does delirium tremens (severe alcohol withdrawal) to do with this? no abbreviations

semantic_digital_twin/src/semantic_digital_twin/physics/equations/differential_equation.py
LucaKro, line 1: consider a "physics/equations" package, where all those lie, with modules "pouring" etc, because its a bit weird imo that there are multipe "...equation.py" modules depending on each other are in the "physics" package.

semantic_digital_twin/src/semantic_digital_twin/physics/pouring_equations.py
```diff
@@ -0,0 +1,141 @@
...
+    """
+    Abstract ODE for pouring-domain fill-level dynamics.
+
+    Owns the outflow rate constant ``k``.
+    Concrete subclasses implement :meth:`symbolic_velocity`.
+
+    :param k: Outflow rate constant.
+    """
+
+    k: float = field(default=1.0, kw_only=True)
```
LucaKro: no abbreviatations.

semantic_digital_twin/src/semantic_digital_twin/physics/equations/pouring_equations.py
LucaKro, line 1: no abbreviations

semantic_digital_twin/src/semantic_digital_twin/physics/pouring_equations.py
LucaKro: please adhere to our naming style guide

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
```diff
@@ -1028,3 +1040,119 @@ def _create_container_event(cls, scale: Scale, wall_thickness: float) -> Event:
         container_event = outer_box.as_composite_set() - inner_box.as_composite_set()
 
         return container_event
+
+
+@dataclass(eq=False)
+class HasContainerGeometry(HasRootBody, ABC):
```
LucaKro: remove, as ContainerGeometry shouldnt exist anymore after my other commend

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
```diff
@@ -1028,3 +1040,119 @@ def _create_container_event(cls, scale: Scale, wall_thickness: float) -> Event:
...
+        :param k: Outflow rate constant for the articulated pouring equation.
+        """
+        phantom = Body(name=PrefixedName(f"{self.root.name.name}_fill_level_phantom"))
+        with world.modify_world():
+            world.add_body(phantom)
+            connection = LiquidConnection.create_with_dofs(
+                world=world,
+                parent=self.root,
+                child=phantom,
+                axis=Vector3(0, 0, 1),
```
LucaKro: Vector3.Z()

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
LucaKro: I think it should inherit from HasRootBody, to maintain access to the properties needed after the HasContainerGeometry removal.

and then i think the method adding the fill level should be an additional method that needs to be called explicitly, similar to `HasSupportingSurface.calculate_supporting_surface`

overriding create_with_new_body_in_world like this will cause problems as it sneaks new behavior in. create_with_new_body_in_world is supposed to be a factory method that just creates the geometry with the correct parentconnection in the world. All other features, like supporting surfaces, or in this case fill level need to be done separately.

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
```diff
@@ -1028,3 +1040,119 @@ def _create_container_event(cls, scale: Scale, wall_thickness: float) -> Event:
...
+    position encodes fill in the range ``[0, 1]``. Use :meth:`create_with_new_body_in_world`
+    for the standard construction path. When the body is constructed manually and placed
+    in the world before creating the semantic annotation, call :meth:`initialize_fill_level`
+    explicitly after the collision geometry has been set.
+    """
+
+    fill_connection: LiquidConnection = field(init=False)
+    """The virtual connection whose position encodes fill level in [0, 1]."""
+
+    _fill_equation: Optional[PouringEquation] = field(default=None, init=False)
```
LucaKro: are those attributes you want to be synchronizes? Id imagine so right? in that case they can neither be init=False not private attributes.

Instead use a normal attribute, and add a setter method "add_fill_equation(self, fill_equation)" that does the same things as the property setter does right now, and decorate that method with    "@synchronized_attribute_modification"

semantic_digital_twin/src/semantic_digital_twin/world_description/connections.py
```diff
@@ -1145,3 +1151,58 @@ def copy_for_world(self, world: World) -> DifferentialDrive:
...
+
+    inflow_equation: Optional[InflowEquation] = field(
```
LucaKro, line 1411: i think this wont be synchronized, not sure if this is intended. maybe it is, since its *always* set via the semantic annotation?

LucaKro (changes requested):
very cool. as we discussed, i think a lot of stuff in here that is currently in pycram should move to semdt, while still avoiding the giskard dependencies. the most reasonable thing here is probably to define abstract interfaces of, for example the "physicsmodel" in semdt, and have giskard implement the giskard version of it.

later when motion state charts are abstracted away from giskard (@ichumuh 👀 ) one could consider if they make sense to use in semdt for this, but im not sure about that yet

giskardpy/src/giskardpy/motion_statechart/tasks/pouring.py
maltehue: You could also define a different function. THe one i use currently is derived from a 2D simplified model

giskardpy/src/giskardpy/motion_statechart/tasks/pouring.py
maltehue: percentage

pycram/src/pycram/body_motion_problem/container_manipulation/physics.py
maltehue: ill check to reuse the MSC history for that

pycram/src/pycram/body_motion_problem/container_manipulation/physics.py
maltehue: you mean the pycram motion executor? i can try to use that actually

pycram/src/pycram/body_motion_problem/container_manipulation/physics.py
```diff
@@ -0,0 +1,71 @@
...
+
+from pycram.body_motion_problem.types import (
+    Effect,
+    PhysicsModel,
+)
+from semantic_digital_twin.world import World
+from semantic_digital_twin.world_description.world_entity import Connection
+
+
+class MissingMotionStatechartError(Exception):
```
Tigul: Move to exceptions.py

pycram/src/pycram/body_motion_problem/container_manipulation/predicates.py
```diff
@@ -0,0 +1,183 @@
...
+                body_group_a=[b for b in gripper.bodies if b.has_collision()],
+                body_group_b=[
+                    b
+                    for b in self.robot._world.bodies
+                    if "apartment" in str(b.name) and b.has_collision()
+                ],
+            ),
+        ]
+
+    def _build_msc(
```
Tigul: This feels to low-level for PyCRAM, why not use the actions here?

pycram/src/pycram/body_motion_problem/predicates.py
```diff
@@ -0,0 +1,252 @@
...
+    the robot's manipulators in turn.
+
+    Returns ``True`` as soon as any gripper successfully completes the trajectory.
+    Returns ``False`` if no gripper can complete it within the timeout.
+
+    Subclass and implement the abstract methods to define embodiment feasibility
+    for a specific manipulation domain.
+    """
+
+    motion: Motion
```
Tigul: doc

pycram/src/pycram/body_motion_problem/types.py
```diff
@@ -0,0 +1,225 @@
...
+                executor.tick()
+                on_tick()
+                if msc.is_end_motion():
+                    break
+            achieved = effect.is_achieved()
+        return achieved
+
+
+@dataclass(eq=True)
+class Motion:
```
Tigul: More motions?? There is already a motion definition in PyCRAM

semantic_digital_twin/src/semantic_digital_twin/physics/equations/differential_equation.py
Tigul: Shouldn't we move this to the symbolic math?

Tigul (changes requested):

giskardpy/src/giskardpy/motion_statechart/tasks/pouring.py
AbdelrhmanBassiouny: do you mean eql symbolic expression ? haha

giskardpy/src/giskardpy/motion_statechart/tasks/pouring.py
```diff
@@ -0,0 +1,114 @@
...
+    ) -> Optional[ObservationStateValues]:
+        """
+        Checks if the goal fill level has been reached and that the outflow is zero.
+
+        :param context: The runtime context.
+        :return: The observation state.
+        """
+        fill = float(self.fill_connection.position)
+        outflow = float(self.fill_vel_ode.evaluate()[0])
+        print(f"Fill level: {fill}, Outflow: {outflow}")
```
AbdelrhmanBassiouny: use logging instead of print

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
```diff
@@ -950,3 +957,100 @@ def _create_container_event(cls, scale: Scale, wall_thickness: float) -> Event:
         container_event = outer_box.as_composite_set() - inner_box.as_composite_set()
 
         return container_event
+
+
+@dataclass(eq=False)
+class HasFillLevel(HasRootBody):
```
AbdelrhmanBassiouny: Why do you need to inherit from HasRootBody here? I think it could work without this inheritance, since this is meant to be a mixin right? so you make other HasRootBody classes inherit from it like a Cup for example.

semantic_digital_twin/src/semantic_digital_twin/world_description/connections.py
```diff
@@ -1083,3 +1091,125 @@ def copy_for_world(self, world: World) -> DifferentialDrive:
...
+
+    @property
+    def tilt_expression(self) -> Optional[Scalar]:
+        """Symbolic tilt angle used by :attr:`outflow_equation` during physics integration."""
+        root_T_child = self._world.compose_forward_kinematics_expression(
+            self._world.root, self.child
+        )
+        return tilt_expression_from_fk(root_T_child)
+
+    def update_state(self, dt: float):
```
AbdelrhmanBassiouny, line 1505: dt -> step_time or time_elapsed_since_previous_step

test/coraplex_test/test_body_motion_problem.py
```diff
@@ -0,0 +1,764 @@
...
+                max_position = drawer.root.parent_connection.active_dofs[
+                    0
+                ].limits.upper.position
+                drawer.root.parent_connection.position = max_position
+
+        task_sym = variable(TaskRequest, domain=[open_task])
+        effect_sym = variable(Effect, domain=effects)
+        motion_sym = variable(Motion, domain=motions)
+
+        query = an(
```
AbdelrhmanBassiouny, line 903: very nice queries you have here tbh

AbdelrhmanBassiouny: excited to see how that will work with verbalization

AbdelrhmanBassiouny (changes requested):
Very nice and clean 💪

PR #300, opened 2026-04-30
Big location refactor

Tigul:
This PR refactors the PyCRAM Locations. Now they consists of a generation back-end and a list of Validators which ensure the generated pose fits the contratints of the pose. 
This refactor makes it easier to merge locations, to create new ones and overall more versatile.


pycram/demos/pycram_bullet_world_demo/demo.py
```diff
@@ -70,7 +70,7 @@
     node = None
 
 pr2 = PR2.from_world(world)
-context = Context(world=world, robot=pr2)
+context = Context(world=world, robot=pr2, _debug=True, ros_node=node)
```
LucaKro: debug is still true

pycram/examples/location_designator.md
```diff
@@ -94,16 +96,19 @@ with simulated_robot:
...
-location_description = CostmapLocation(target=world.get_body_by_name("milk.stl").global_pose, reachable=True, reachable_arm=Arms.LEFT, context=context)
+location = reachability_location(world.get_body_by_name("milk.stl"), context=context, arm=Arms.LEFT)
 
-plan = execute_single(NavigateAction(next(iter(location_description))), context=context)
+plan = execute_single(NavigateAction(next(iter(location))), context=context)
 
 with simulated_robot:
     plan.perform()
+
+with world.modify_world():
```
LucaKro: i dont think we need a modify world block here

LucaKro: same below for the other code examples

pycram/examples/location_designator.md
```diff
@@ -138,9 +148,12 @@ already have a milk spawned in you world you can ignore the following cell.
 
 ```python
 
-location_description = CostmapLocation(target=Pose(Point3.from_iterable([-1, 0, 1.2]), reference_frame=world.root), visible=True, context=context)
+location = visibility_location(Pose(Point3.from_iterable([-1, 0, 1.2]), reference_frame=world.root), context=context)
 
-for i, pose in enumerate(location_description):
+
+# location_description = CostmapLocation(target=Pose(Point3.from_iterable([-1, 0, 1.2]), reference_frame=world.root), visible=True, context=context)
```
LucaKro: commented out code

pycram/examples/location_designator.md
```diff
@@ -156,8 +169,18 @@ At the moment this location designator only works in the apartment environment,
...
+            root=world.get_body_by_name("cabinet10_drawer_middle"),
+            handle=Handle(root=world.get_body_by_name("handle_cab10_m")),
+        )
+    )
+
+location = accessing_location(world.get_semantic_annotations_by_type(Drawer)[0], context=context, arm=Arms.LEFT)
 
-access_location = AccessingLocation(world.get_body_by_name("handle_cab10_t"), arm=Arms.LEFT, context=context)
-print(next(iter(access_location)))
+print(next(iter(location)))
```
LucaKro, line 180: remove print

pycram/examples/location_designator.md
```diff
@@ -156,8 +169,18 @@ At the moment this location designator only works in the apartment environment,
...
 ```python
-from pycram.locations.locations import AccessingLocation
+from pycram.locations.factories import accessing_location
+from semantic_digital_twin.semantic_annotations.semantic_annotations import Drawer, Handle
+
+with world.modify_world():
+    world.add_semantic_annotation(
+        drawer := Drawer(
+            root=world.get_body_by_name("cabinet10_drawer_middle"),
+            handle=Handle(root=world.get_body_by_name("handle_cab10_m")),
```
LucaKro, line 174: this is sus, if someone did it like this the handle would not be added to the world, and if its added afterwards then copying wont work anymore.

pycram/src/pycram/locations/backends.py
```diff
@@ -0,0 +1,215 @@
...
+        base_bb = self.robot.base.bounding_box
+
+        occupancy_map = OccupancyCostmap(
+            resolution=0.02,
+            height=200,
+            width=200,
+            world=self.world,
+            robot_view=self.robot,
+            origin=ground_pose,
+            distance_to_obstacle=(base_bb.width / 2 + base_bb.depth / 2) / 2 + 0.5,
```
LucaKro: magic numbers

pycram/src/pycram/locations/base.py
```diff
@@ -0,0 +1,154 @@
...
+logger = logging.getLogger("pycram")
+
+
+@dataclass
+class Location(Iterable[Pose]):
+    """
+    Base for all locations that can be used to plan movements.
+    """
+
+    context: Optional[Context]
```
LucaKro, line 32: why optional? I think a location without a world doesnt really make sense

pycram/src/pycram/locations/base.py
```diff
@@ -0,0 +1,154 @@
...
+    """
+    The context where the location can extract information about the robot and world from.
+    """
+
+    target_pose: Pose
+    """
+    The target pose for the location.
+    """
+
+    generator: PoseGeneratorBackend = field(default=None)
```
LucaKro: i dont like default None when its not actually optional

pycram/src/pycram/locations/base.py
```diff
@@ -0,0 +1,154 @@
...
+    """
+    Backend that generates pose candidates.
+    """
+    validators: List[PoseValidator] = field(default=None)
```
LucaKro: default None for a list doesnt make sense

pycram/src/pycram/locations/base.py
```diff
@@ -0,0 +1,154 @@
...
+        :return: A new location that is the merge of this location and the other location.
+        """
+        return Location(
+            self.context,
+            self.target_pose,
+            self.generator.merge(other.generator),
+            self.validators + other.validators,
+        )
+
+    def __and__(self, other: Location) -> Location:
```
LucaKro, line 115: do we need this? I feel like we should just use merge, we have merge everywhere in our codebase and this provides two ways of doing the same thing (both of which need to be maintained in the future) without any upsides

pycram/src/pycram/locations/base.py
```diff
@@ -0,0 +1,154 @@
...
+    The world in which the pose candidate should be validated.
+    """
+
+    robot: AbstractRobot
+    """
+    The robot that should be used to validate the pose candidate.
+    """
+
+    @abstractmethod
+    def __call__(self, *args, **kwargs) -> bool:
```
LucaKro, line 160: do we really want to implement call for that instead of some other method?

pycram/src/pycram/locations/costmaps.py
```diff
@@ -750,7 +747,13 @@ def _generate_map(self):
         # the locations in itself is consistent and just needs to be flipped to fit the world coordinate system
         map = np.flip(map, axis=0)
         map = np.flip(map)
-        self.map = map
+
+        # Invert the map
+        inv_map = np.zeros(map.shape)
+        inv_map[map == 0] = 1
+        inv_map[map != 0] = 0
```
LucaKro, line 791: this doesnt just invert the map, if there are floats inside, is that deliberate?

pycram/src/pycram/locations/factories.py
```diff
@@ -0,0 +1,255 @@
...
+from semantic_digital_twin.semantic_annotations.semantic_annotations import (
+    Cabinet,
+    Drawer,
+)
+from semantic_digital_twin.spatial_types.spatial_types import Pose
+from semantic_digital_twin.world import World
+from semantic_digital_twin.world_description.world_entity import Body
+
+
+def _make_default_occupancy_costmap(context: Context, target: Pose) -> OccupancyCostmap:
```
LucaKro: why not have this be a classmethod of occupancycostmap? i dont see the benefit of having a factory like this for it

pycram/src/pycram/locations/factories.py
```diff
@@ -0,0 +1,255 @@
...
+    manipulators = (
+        [manipulator] if not isinstance(manipulator, list_like_classes) else manipulator
+    )
+    objs = set()
+    for man in manipulators:
+        objs.update(
+            test_world.get_kinematic_structure_entities_of_branch(man.tool_frame)
+        )
+        objs.remove(man.tool_frame)
+    return objs.pop() if objs else None
```
LucaKro, line 49: idk about using sets here and about not just returning a list of objects. I feel like this could be part of the abstract robot or arm/manipulator sem annotation

pycram/src/pycram/locations/factories.py
```diff
@@ -0,0 +1,255 @@
...
+        ApproachDirection.FRONT,
+        VerticalAlignment.NoAlignment,
+        man,
+    )
+    costmap = _make_default_occupancy_costmap(context, target_pose) & RingCostmap(
+        resolution=0.02,
+        width=200,
+        height=200,
+        std=15,
+        distance=0.6,  # That needs to be replaced with an estimate of the reachability space of the robot arms
```
LucaKro: in this PR? so many magic numbers 🥲

pycram/src/pycram/locations/locations.py
LucaKro, line 1: there is still a GiskardLocation in here? how does this differ from a giskardlocationbackend?

pycram/src/pycram/locations/pose_validator.py
```diff
@@ -0,0 +1,171 @@
...
+
+@dataclass
+class VisibilityValidator(PoseValidator):
+
+    target_pose: Pose = field(default=None)
+
+    target_body: Body = field(default=None)
+
+    def __call__(self, *args, **kwargs) -> bool:
+        assert self.target_pose or self.target_body
```
LucaKro: please throw proper errors instead of using asserts, as in certain configurations of python asserts are just ignored for performance reasons. if that is fine, it means the asserts are not necessary and can be removed, and if thats not fine it breaks the code

pycram/src/pycram/locations/pose_validator.py
```diff
@@ -0,0 +1,171 @@
...
+        return self.validate_body() if self.target_body else self.validate_pose()
+
+    def validate_pose(self) -> bool:
+        gen_body = Body(
+            name=PrefixedName("vist_test_obj", "pycram"),
+            collision=ShapeCollection([Box(scale=Scale(0.1, 0.1, 0.1))]),
+        )
+        with self.world.modify_world():
+            self.world.add_connection(
+                Connection6DoF.create_with_dofs(
```
LucaKro: create a fixed connection, its cheaper and there is no benefit for a 6dof connection here

pycram/src/pycram/locations/pose_validator.py
```diff
@@ -0,0 +1,171 @@
...
+
+    def validate_body(self) -> bool:
+        return self._ray_test(self.target_body)
+
+    def _ray_test(self, target_body: Body) -> bool:
+        r_t = self.world.ray_tracer
+        camera = self.robot.get_default_camera()
+        ray = r_t.ray_test(
+            camera.bodies[0].global_transform.to_position().to_np()[:3],
+            target_body.global_transform.to_position().to_np()[:3],
```
LucaKro: i dont think you need to_np here, i think the slicing works without

pycram/src/pycram/locations/pose_validator.py
```diff
@@ -0,0 +1,171 @@
...
+    KinematicStructureEntity,
+)
+
+logger = logging.getLogger("pycram")
+
+
+@dataclass
+class VisibilityValidator(PoseValidator):
+
+    target_pose: Pose = field(default=None)
```
LucaKro, line 47: docs, same for the others

pycram/src/pycram/robot_plans/actions/composite/transporting.py
```diff
@@ -319,88 +319,3 @@ def validate(
     ):
         # The validation will be done in each of the core action perform methods so no need to validate here.
```
LucaKro: idk i i feel like you could successfully validate pick and place, and still fail transporting because of some failurehandling stuff or sth no? I know thats not the case yet, but the comment makes it seem like its categorically impossible, maybe adjust that

LucaKro: the current validation could just be  that it checks that pick and place both have been successfully validated

pycram/src/pycram/motion_executor.py
```diff
@@ -98,7 +98,7 @@ def _execute_for_simulation(self):
         executor.compile(self.motion_state_chart)
         # execute the motion state chart until it is done
         counter = 0
-        while counter < 2000:
+        while counter < len(self.motions) * 1500:
```
LucaKro, line 101: does self.motions include the nodes for collision avoidance? if yes then thats not good, those are a lot of nodes, i think it makes more sense to filter out all collision avoidance nodes. kinda the same for parallel, but we dont use that yes anyways, so maybe just leave a comment for that

pycram/src/pycram/utils.py
```diff
@@ -44,6 +48,30 @@
     from pycram.view_manager import CameraDescription
 
 
+def collision_check(robot: AbstractRobot, world: World) -> List[ClosestPoints]:
```
LucaKro: i think this should be a method of world

pycram/src/pycram/view_manager.py
```diff
@@ -17,11 +17,12 @@ class ViewManager:
...
     @symbolic_function
-    def get_end_effector_view(
-        arm: Arms, robot_view: AbstractRobot
-    ) -> Optional[Manipulator]:
+    def get_end_effector_view(arm: Arms, robot_view: AbstractRobot) -> Manipulator:
         arm = ViewManager.get_arm_view(arm, robot_view)
-        return arm.manipulator
+        man = arm.manipulator
+        if not man:
+            raise ValueError(f"The robot view {robot_view} has no manipulator.")
```
LucaKro: dedicated error

semantic_digital_twin/src/semantic_digital_twin/reasoning/predicates.py
```diff
@@ -578,9 +579,7 @@ def is_place_occupied(
     region_box_shape = box.as_shape()  # returns a Box centered at the region
     region_mesh = region_box_shape.mesh.copy()
     # region_mesh.apply_transform(region_box_shape.origin.to_np())
```
LucaKro: commented out code

LucaKro (changes requested):
nice one!

pycram/examples/location_designator.md
Tigul: No

pycram/src/pycram/locations/base.py
Tigul: I think @tomsch420 did this initially

pycram/src/pycram/locations/base.py
Tigul: Yes because this is a Predicate

pycram/src/pycram/locations/costmaps.py
Tigul: Yes, because floats do not have much info if a pose is good or not

pycram/src/pycram/locations/factories.py
Tigul: I know but I also don't know how to define this better

pycram/src/pycram/locations/base.py
Tigul: For documentation: It is possible to create a location without a Context, if the location is not yet added to a plan. Therefore the context is optional since otherwise this would cause problems with the ORM

pycram/src/pycram/motion_executor.py
Tigul: No they are not included there, the motion_executor would add them afterwards

pycram/src/pycram/locations/backends.py
```diff
@@ -0,0 +1,215 @@
...
+        reachability_map = occupancy_map + gaussian_map
+        reachability_map.number_of_samples = 5
+
+        return reachability_map
+
+    def setup_giskard_executor(
+        self,
+        pose_sequence: List[Pose],
+        world: World,
+        robot_view: AbstractRobot,
```
AbdelrhmanBassiouny: I prefer just `robot` here as it's clearer, if you want it like that then robot_semantic_annotation is more accurate.

pycram/src/pycram/locations/base.py
AbdelrhmanBassiouny: you can actually remove it and just add ABC in the bases, since it is already abstract method in Predicate class.

pycram/src/pycram/locations/factories.py
```diff
@@ -0,0 +1,231 @@
...
+    objs = set()
+    for man in manipulators:
+        objs.update(
+            test_world.get_kinematic_structure_entities_of_branch(man.tool_frame)
+        )
+        objs.remove(man.tool_frame)
+    return objs.pop() if objs else None
+
+
+def occupancy_location(target_pose: Pose, context: Context) -> Location:
```
AbdelrhmanBassiouny, line 52: The Location factories can be better put in classes such that they are more explicit and declarative, one can point to what is a ReachabilityLocation and that it is different from a VisibilityLocation. Similar to how you the costmaps already as classes.

pycram/src/pycram/locations/pose_validator.py
AbdelrhmanBassiouny: ☝️

pycram/src/pycram/locations/pose_validator.py
```diff
@@ -0,0 +1,204 @@
...
+    target_body: Body = field(default=None)
+    """
+    Body for which visibility should be checked
+    """
+
+    def __call__(self, *args, **kwargs) -> bool:
+        assert self.target_pose or self.target_body
+        return self.validate_body() if self.target_body else self.validate_pose()
+
+    def validate_pose(self) -> bool:
```
AbdelrhmanBassiouny, line 62: where doc? This is not a simple method.

pycram/src/pycram/locations/pose_validator.py
```diff
@@ -0,0 +1,204 @@
...
+            with self.world.modify_world():
+                self.world.remove_connection(gen_body.parent_connection)
+                self.world.remove_kinematic_structure_entity(gen_body)
+
+        return result
+
+    def validate_body(self) -> bool:
+        return self._ray_test(self.target_body)
+
+    def _ray_test(self, target_body: Body) -> bool:
```
AbdelrhmanBassiouny, line 94: doc

pycram/src/pycram/locations/pose_validator.py
```diff
@@ -0,0 +1,204 @@
...
+            multiple_hits=True,
+        )
+
+        hit_bodies = [b for b in ray[2] if not b in self.robot.bodies]
+
+        return hit_bodies[0] == target_body if len(hit_bodies) > 0 else False
+
+
+@dataclass
+class ReachabilityValidator(PoseValidator):
```
AbdelrhmanBassiouny: I prefer the names of predicates to be like a question, such that they read naturally in queries, e.g. IsReachableBy

pycram/src/pycram/locations/pose_validator.py
```diff
@@ -0,0 +1,204 @@
...
+        return ReachabilitySequenceValidator(
+            pose_sequence=[self.pose],
+            tip_link=self.tip_link,
+            robot=self.robot,
+            world=self.world,
+        ).__call__()
+
+
+@dataclass
+class ReachabilitySequenceValidator(PoseValidator):
```
AbdelrhmanBassiouny: AreReachableBy?

pycram/src/pycram/robot_plans/actions/core/container.py
```diff
@@ -83,12 +83,11 @@ def pre_condition(
 
         return and_(
             GripperIsFree(manipulator),
-            reachability_validator(
-                kwargs["object_designator"].global_pose,
-                manipulator.tool_frame,
-                context.robot.from_world(test_world),
-                test_world,
-                context.robot.full_body_controlled,
+            ReachabilityValidator(
```
AbdelrhmanBassiouny: See, it would sound nicer here to say IsReachableBy and would be consistent with GripperIseFree

pycram/src/pycram/utils.py
AbdelrhmanBassiouny: I found that making methods that have two different semantic jobs like checking if a collision happend and also return the collisions are not easy to know when to use and also will have a name that conflicts with one of the two jobs.

I think it would be nicer if you have function that gets the collisions, and a Predicate that checks if a collision happend and this check can use th function.

AbdelrhmanBassiouny (changes requested):
Well done 👍

pycram/src/pycram/utils.py
LucaKro: id be open to see what that looks like :eyes:

pycram/src/pycram/locations/factories.py
Tigul: I initiall tried that and it didn't make sense since the factories don't implement any functionality and just combine the different back-endds and validator

pycram/src/pycram/utils.py
Tigul: I deleted that and replaced it with the collision detector of semDT

LucaKro (approved):

sunava (approved):

AbdelrhmanBassiouny (approved):

PR #312, opened 2026-05-11
Tracy Canister insertion demo 

maltehue:
The cram code that was updated to get the tracy canister insertion demo to work. The demo execution script is not part of the cram repo.

this PR mainly includes 
- changes to the tracy setup
- port of the canister insertion goal
- ForceZMonitor Port

@MarvinDufner Please check if there are other debug changes that should be reverted.


giskardpy/src/giskardpy/motion_statechart/goals/tracebot.py
```diff
@@ -6,129 +6,253 @@
...
-        if self.up is None:
-            self.up = Vector3.Z()
-            self.up.reference_frame = self.root
-        self.root_V_up = context.world.transform(
-            target_frame=self.root, spatial_object=self.up
-        )
-
-        self.weight = DefaultWeights.WEIGHT_ABOVE_CA
+@dataclass(eq=False, repr=False)
+class _ReachTopTask(Task):
```
Tigul: Doc strings for the whole class

giskardpy/src/giskardpy/motion_statechart/goals/tracebot.py
```diff
@@ -6,129 +6,253 @@
...
+    root_link: KinematicStructureEntity = field(kw_only=True)
+    tip_link: KinematicStructureEntity = field(kw_only=True)
+    root_P_hole: Point3 = field(kw_only=True)
+    root_V_up: Vector3 = field(kw_only=True)
+    pre_grasp_height: float = field(kw_only=True)
```
Tigul: Also for the parameter

giskardpy/src/giskardpy/motion_statechart/goals/tracebot.py
```diff
@@ -6,129 +6,253 @@
...
+        return artifacts
 
-        # %% tilted orientation goal
-        tilt_error = root_V_cylinder_z.angle_between(root_V_up)
-        tilt_task = Task(name="Slightly Tilted")
-        self.add_task(tilt_task)
-        tilt_task.add_position_constraint(
+
+@dataclass(eq=False, repr=False)
+class _SlightlyTiltedTask(Task):
```
Tigul: DOC

giskardpy/src/giskardpy/motion_statechart/goals/tracebot.py
```diff
@@ -6,129 +6,253 @@
...
-        root_V_cylinder_z.vis_frame = self.tip
-        tilt_task.observation_expression = sm.abs(tilt_error - self.tilt) <= 0.01
+        artifacts.observation = sm.abs(tilt_error - self.tilt) <= 0.01
+        return artifacts
 
-        init_done = f"{reach_top} and {tilt_task}"
 
-        reach_top.end_condition = init_done
+@dataclass(eq=False, repr=False)
+class _StayOnLineTask(Task):
```
Tigul: DCO

giskardpy/src/giskardpy/motion_statechart/goals/tracebot.py
```diff
@@ -6,129 +6,253 @@
...
+            quadratic_weight=self.weight,
             name="pregrasp",
         )
-        stay_on_line.observation_expression = distance_to_line < 0.01
+        artifacts.observation = distance_to_line < 0.01
+        return artifacts
+
+
+@dataclass(eq=False, repr=False)
+class _InsertTask(Task):
```
Tigul: DoC

Tigul (changes requested):

giskardpy/src/giskardpy/motion_statechart/goals/tracebot.py
LucaKro: why private class?

giskardpy/src/giskardpy/motion_statechart/goals/tracebot.py
```diff
@@ -6,129 +6,253 @@
...
+        tip_P_cylinder_bottom = -Vector3.Z() * self.cylinder_height / 2
+        root_P_tip = root_T_tip.to_position() + root_T_tip @ tip_P_cylinder_bottom
+        root_P_top = self.root_P_hole + self.root_V_up * self.pre_grasp_height
         distance_to_top = root_P_tip.euclidean_distance(root_P_top)
 
-        distance_to_line, root_P_on_line = root_P_tip.distance_to_line_segment(
-            root_P_hole, root_P_top
-        )
-        distance_to_hole = (root_P_hole - root_P_tip).norm()
+        print(f"goal in ReachTop: {root_P_top.to_np()}")
```
LucaKro: remove prints

giskardpy/src/giskardpy/motion_statechart/goals/tracebot.py
```diff
@@ -6,129 +6,253 @@
...
+        root_V_cylinder_z = root_T_tip @ Vector3.Z()
+        tilt_error = root_V_cylinder_z.angle_between(self.root_V_up)
+
+        artifacts = NodeArtifacts()
+        artifacts.constraints.add_vector_goal_constraints(
             frame_V_current=root_V_cylinder_z,
-            frame_V_goal=root_V_up,
-            reference_velocity=0.1,
+            frame_V_goal=self.root_V_up,
+            reference_velocity=0.025, # 0.001 was to small, the real robot would not move
```
LucaKro: no split between angular velocity and linear velocity?

LucaKro (changes requested):

giskardpy/src/giskardpy/motion_statechart/goals/tracebot.py
MarvinDufner: thats a vector goal constraint which just aligns the tip with the goal vector, so only angular velocity  is used and not linear velocity, like in a rotation goal constraint.


the split you are thinking of is probably for poses (which consist of a rotation and point constraint) and doesnt apply for constraints

coraplex/src/coraplex/datastructures/dataclasses.py
```diff
@@ -72,6 +81,13 @@ def debug(self, value):
             raise ValueError("Debug mode requires a ROS node")
         logging.getLogger("coraplex").setLevel(logging.DEBUG if self.debug else logging.INFO)
 
+    @property
+    def giskard_wrapper(self):
```
MarvinDufner, line 145: It currently takes way too long in between each action because the giskard wrapper is reinitialized every time (which also fetches the world every time ...), so this is a cached instance to make things faster. But i dont know where that instance actually should go. I figured the context is the best place, does anyone have a better idea how to properly do this?

MarvinDufner: I could also just pass the world to the GiskardWrapper which does reduce the slowness but then the action client is still built every time which is just unnecessary imo

MarvinDufner: Also as far as I can see the action client in the giskard wrapper never gets destroyed, so I guess we currently just accumulate an action client for each action, no?

ichumuh: probably true @Tigul what do you think?

giskardpy/src/giskardpy/motion_statechart/goals/tracebot.py
```diff
@@ -6,129 +6,321 @@
...
+
+    cylinder_height: float = field(kw_only=True)
+    """Height of the body, used to locate its bottom."""
+
+    def build(self, context: MotionStatechartContext) -> NodeArtifacts:
+        root_T_tip = context.world.compose_forward_kinematics_expression(
+            self.root_link, self.tip_link
         )
-        distance_to_hole = (root_P_hole - root_P_tip).norm()
+        tip_P_cylinder_bottom = -Vector3.Z() * self.cylinder_height / 2
```
ichumuh: i don't like the hard coded -Z

ichumuh: also name should be `tip_V_cylinder_bottom` because its a vector

giskardpy/src/giskardpy/motion_statechart/goals/tracebot.py
```diff
@@ -6,129 +6,321 @@
...
+
+    tilt: float = field(kw_only=True)
+    """Target angle in rad between the body's axis and the up axis."""
+
+    def build(self, context: MotionStatechartContext) -> NodeArtifacts:
+        root_T_tip = context.world.compose_forward_kinematics_expression(
+            self.root_link, self.tip_link
         )
-        reach_top.observation_expression = distance_to_top < 0.01
+        root_V_cylinder_z = root_T_tip @ Vector3.Z()
```
ichumuh: again hard coded Z

ichumuh: ah there is even `root_V_up` as a parameter, use that grrr

giskardpy/src/giskardpy/motion_statechart/goals/tracebot.py
```diff
@@ -6,129 +6,321 @@
...
+    """Body that is controlled."""
 
-        # %% move down
-        stay_on_line = Task(name="Stay on Straight Line")
-        self.add_task(stay_on_line)
-        stay_on_line.add_point_goal_constraints(
+    root_P_hole: Point3 = field(kw_only=True)
+    """Position of the hole, expressed in the root frame."""
+
+    root_V_up: Vector3 = field(kw_only=True)
```
ichumuh: grrr

giskardpy/src/giskardpy/motion_statechart/goals/tracebot.py
```diff
@@ -6,129 +6,321 @@
...
+    """Distance above the hole along the up axis at which the line ends."""
+
+    cylinder_height: float = field(kw_only=True)
+    """Height of the body, used to locate its bottom."""
+
+    def build(self, context: MotionStatechartContext) -> NodeArtifacts:
+        root_T_tip = context.world.compose_forward_kinematics_expression(
+            self.root_link, self.tip_link
+        )
+        tip_P_cylinder_bottom = -Vector3.Z() * self.cylinder_height / 2
```
ichumuh: double grrr

giskardpy/src/giskardpy/motion_statechart/goals/tracebot.py
```diff
@@ -6,129 +6,321 @@
...
+    """Position of the hole, expressed in the root frame."""
+
+    cylinder_height: float = field(kw_only=True)
+    """Height of the body, used to locate its bottom."""
+
+    def build(self, context: MotionStatechartContext) -> NodeArtifacts:
+        root_T_tip = context.world.compose_forward_kinematics_expression(
+            self.root_link, self.tip_link
+        )
+        tip_P_cylinder_bottom = -Vector3.Z() * self.cylinder_height / 2
```
ichumuh: double grrr

giskardpy/src/giskardpy/motion_statechart/goals/tracebot.py
```diff
@@ -6,129 +6,321 @@
...
+    """Axis pointing out of the hole, expressed in the root frame."""
 
-        tilt_task.end_condition = bottom_reached
-        # %% tilt straight
-        # tilt_monitor.observation_expression = cas.less(tilt_error, 0.01)
+    def build(self, context: MotionStatechartContext) -> NodeArtifacts:
+        root_T_tip = context.world.compose_forward_kinematics_expression(
+            self.root_link, self.tip_link
+        )
+        root_V_cylinder_z = root_T_tip @ Vector3.Z()
```
ichumuh: grrr

giskardpy/src/giskardpy/motion_statechart/goals/tracebot.py
```diff
@@ -6,129 +6,321 @@
...
+    cylinder_name: Body = field(kw_only=True)
+    """Cylinder body to insert. Used as the tip link of the kinematic chain."""
+
+    hole_point: Point3 = field(kw_only=True)
+    """Position of the hole to insert the cylinder into."""
+
+    cylinder_height: Optional[float] = None
+    """Height of the cylinder. If None, it is taken from the cylinder's collision shape."""
+
+    up: Optional[Vector3] = None
```
ichumuh: inconsistent name

giskardpy/src/giskardpy/motion_statechart/goals/tracebot.py
```diff
@@ -6,129 +6,321 @@
...
+
+        if self.cylinder_height is None:
+            self.cylinder_height = self.cylinder_name.collision.shapes[0].height
+
+        if self.up is None:
+            up = Vector3.Z(root)
+        else:
+            up = self.up
+        root_P_hole = context.world.transform(self.hole_point, root)
+        root_V_up = context.world.transform(up, root)
```
ichumuh: you don't need to transform when self.up was none

giskardpy/src/giskardpy/motion_statechart/ros2_nodes/topic_monitor.py
```diff
@@ -59,6 +59,11 @@ class TopicSubscriberNode(TopicNode[MsgType]):
 
     def build(self, context: MotionStatechartContext) -> NodeArtifacts:
         node_artifacts = super().build(context)
+
+        existing = getattr(self, "_subscriber", None)
```
ichumuh: why getattr? and how can it exist if you clean up?

giskardpy/src/giskardpy/motion_statechart/ros2_nodes/topic_monitor.py
```diff
@@ -100,13 +112,25 @@ class TopicPublisherNode(TopicNode[MsgType]):
 
     def build(self, context: MotionStatechartContext) -> NodeArtifacts:
         node_artifacts = super().build(context)
+
+        existing = getattr(self, "_publisher", None)
```
ichumuh: same as above

semantic_digital_twin/src/semantic_digital_twin/robots/tracy.py
```diff
@@ -109,6 +109,14 @@ def setup_default_configuration_in_world_below_robot_root(
...
         )
 
+    def _setup_velocity_limits(self):
+        """
+        The UR10e default velocity limits are either 2.09 or 3.14 depending on the joint. Wrist joints are 3.14.
+        Clipping the limits below 3.14 causes weird movements in end effector orientation space,
+        as the wrist stops moving with the correct speed relative to the other arm joints.
+        """
+        vel_limits = defaultdict(lambda: 4.0)
+        self.tighten_dof_velocity_limits_of_1dof_connections(new_limits=vel_limits)
```
ichumuh: we now have the proportional scaling, maybe use that?

test/giskardpy_test/test_motion_statechart/test_insert_cylinder.py
```diff
@@ -0,0 +1,80 @@
...
+            world.add_connection(
+                connection_type.create_with_dofs(
+                    world=world, parent=parent, child=child, axis=axis
+                )
+            )
+        MinimalRobot.from_world(world)
+    return world
+
+
+def test_insert_cylinder():
```
ichumuh: a test where a robot is holding it would be nice

ichumuh (changes requested):

giskardpy/src/giskardpy/motion_statechart/ros2_nodes/topic_monitor.py
MarvinDufner: nodes get cleaned after end- or cancelmotion, but currently build gets called multiple times if they are part of a goal during msc construction (bcs you go through every node to build but if they are part of a goal they get built multiple times bcs children have to be built before the parents). so i dont want to create multiple subscriber in case build is called multiple times. removed getattr though :+1:

ichumuh: instead of fighting the symptoms, why not fix the cause? anyway it'll be fixed with this pr https://github.com/cram2/cognitive_robot_abstract_machine/pull/407

giskardpy/src/giskardpy/motion_statechart/goals/tracebot.py
```diff
@@ -11,81 +11,78 @@
...
 from semantic_digital_twin.world_description.world_entity import (
-    Body,
     KinematicStructureEntity,
 )
 
 
 @dataclass(eq=False, repr=False)
-class ReachTopTask(Task):
-    """Moves the body's bottom to the pre-grasp point above the hole."""
+class ReachPointTask(Task):
```
ichumuh: isn't that a cartesian position task?

giskardpy/src/giskardpy/motion_statechart/goals/tracebot.py
```diff
@@ -235,66 +196,65 @@ class InsertCylinder(Goal):
...
 
     def expand(self, context: MotionStatechartContext) -> None:
         root = context.world.root
 
-        if self.cylinder_height is None:
-            self.cylinder_height = self.cylinder_name.collision.shapes[0].height
-
-        if self.up is None:
-            up = Vector3.Z(root)
+        tip_V_axis = self.tip_V_axis if self.tip_V_axis is not None else Vector3.Z()
```
ichumuh: default factory gaming

test/giskardpy_test/test_motion_statechart/test_cartesian_tasks.py
```diff
@@ -299,15 +306,26 @@ def test_simple_cartesian_pose(self, cylinder_bot_world: World):
...
+        motion_statechart.add_node(EndMotion.when_true(goal))
 
-        kin_sim = Executor(MotionStatechartContext(world=cylinder_bot_world))
-        kin_sim.compile(motion_statechart=msc)
-        kin_sim.tick_until_end()
+        executor = Executor(MotionStatechartContext(world=cylinder_bot_world))
+        executor.compile(motion_statechart=motion_statechart)
+        executor.tick_until_end()
+
+        forward_kinematics = cylinder_bot_world.compute_forward_kinematics_np(
```
ichumuh: why are you touching this?

test/giskardpy_test/test_motion_statechart/test_cartesian_tasks.py
```diff
@@ -368,33 +384,44 @@ def test_cart_goal_1eef(self, pr2_world_state_reset: World):
...
             MotionStatechartContext(
                 world=pr2_world_state_reset,
             )
         )
-        kin_sim.compile(motion_statechart=msc)
-        kin_sim.tick_until_end()
+        executor.compile(motion_statechart=motion_statechart)
+        executor.tick_until_end()
 
+        forward_kinematics = executor.context.world.compute_forward_kinematics_np(
```
ichumuh: and this

ichumuh: why are you touching this file at all?

test/giskardpy_test/test_motion_statechart/test_insert_cylinder.py
```diff
@@ -1,80 +1,92 @@
...
+        pre_grasp_height=0.1,
     )
+    close_gripper = JointPositionList(
+        goal_state=robot.right_arm.end_effector.get_joint_state_by_type(
+            GripperState.CLOSE
+        )
+    )
+    msc.add_node(close_gripper)
     msc.add_node(goal)
+    goal.start_condition = close_gripper.observation_variable
```
ichumuh: use sequence instead of manually linking the nodes

test/giskardpy_test/test_motion_statechart/test_json_parsing.py
```diff
@@ -264,7 +264,9 @@ def test_cart_goal_simple(pr2_world_state_reset: World):
     kin_sim.tick_until_end()
 
     fk = pr2_world_state_reset.compute_forward_kinematics_np(root, tip)
-    assert np.allclose(fk, tip_goal, atol=cart_goal.threshold)
+    expected = tip_goal.to_np()
```
ichumuh: no need to convert to numpy

test/giskardpy_test/test_motion_statechart/test_motion_statechart.py
```diff
@@ -1692,3 +1794,34 @@ def test_long_pause(self):
...
+        msc.add_node(EndMotion.when_true(cart_goal))
+
+        kin_sim = Executor(MotionStatechartContext(world=pr2_world_state_reset))
+        kin_sim.compile(motion_statechart=msc)
+        kin_sim.tick_until_end()
+
+        fk = pr2_world_state_reset.compute_forward_kinematics_np(
+            pr2_world_state_reset.root, tip
+        )
+        expected = goal_pose.to_np()
```
ichumuh: no need to convert to numpy

test/giskardpy_test/test_motion_statechart/test_json_parsing.py
ichumuh: symbolic math objects are converted automatically when passed to numpy functions

ichumuh (changes requested):

test/giskardpy_test/test_motion_statechart/test_cartesian_tasks.py
ichumuh: ah you are touching because of the two thresholds. you still don't need to convert to np though

giskardpy/src/giskardpy/motion_statechart/goals/tracebot.py
MarvinDufner: oh true its kind of the same thing. But this one drives an offset instead of the tool frame itself. I use it to move the bottom of the cylinder instead of the cylinder center. 

I dont know if its replacable by CartesianPosition exactly because then I would need to offset the hole point by the tip_V_tool offset. but the direction would be difficult because the cylinder is inserted while tilted so the vector along which to offset is kinda difficult to determine.

MarvinDufner: i mean i could also just add some sort of tip_P_offset parameter to cartesianPosition to offset the goal and then thats the same thing i guess

ichumuh: hm true, i don't have a strong opinion on that right now. a cartesian position where the offset is 0 by default would keep the same behavior, without being too confusing i think. even if we don't merge it with cartesian position, i think a feature function thing would be a better fit. what does claude think?

MarvinDufner: claude thinks a feature function would be better bcs youll probably never need a goal relative to something other than the world root or a binding policy while driving the controlled point

ichumuh: Yes, and use auto mode

PR #322, opened 2026-05-14
Moved cram_ros2_packages to own repo

Tigul:
This PR moves the cram_ros2_packages to its own repo and installs it in the Docker for testsing


J-Schaefer (changes requested):
The PR itself is fine, there are two issues with the setup script for the ROS workspace. As I see it, it's made to setup the docker environment and is unsuitable for a setup on a real machine.
Right now the script wouldn't work when executed like this. Since it's made to run in a docker container it's missing `sudo` before `apt install`. But we also don't want to run the whole script with `sudo`. Better to add a dedicated script that fixes that.
Also the script is a bit inflexible and assumes the workspace doesn't exist yet. Meaning the script can only be used for the first setup; when run a second time it either doesn't do anything or throws an error when a directory, that it tries to create, already exists. An option would be to add checks and automatically pull newer versions of the repos if available. In contrast, the `apt install` command in the script would update the packages as well when run a second time and could install versions that are too new for the current versions of the local repos.

Tigul:
What exactly do you mean with a second script that fixes the sudo thing? 
The script is intended for a first-time setup so I think it is fine to assume that the workspace does not exist yet.

sunava (approved):

J-Schaefer (approved):
New changes solve my previous comments.

PR #323, opened 2026-05-14
[KRROOD] Roles Hacky Version

AbdelrhmanBassiouny:
This has no code generation, but in my opinion, this is a bit more hacky, as I currently modify the init.

The good thing is that it works, with no extra code.

So Now for IDE support you also inherit from the role taker class, the data is not not duplicated as I take the role_taker as a keyword only argument and that's where the data comes from.

I have tested it against the Rooms, and the GraspPose. 

```python
@dataclass(eq=False)
class Kitchen(Role[Room], Room): # The Role should also inherit from the role taker type (only for the IDE support).
    room: Room = field(kw_only=True) # the role taker should be a keyword-only argument
    utilities: list[str] = field(default_factory=list)

    @classmethod
    def role_taker_attribute(cls) -> Attribute[Room]:
        return variable_from(cls).room
        
room = Room(floor=ground_floor)
kitchen = Kitchen(room=room) # Or Kitchen.from_role_taker(room)
# Or you can create the room implicitly
kitchen = Kitchen(floor=ground_floor)
# then
room = kitchen.room

# Identity is shared
assert kitchen == room
assert hash(kitchen) == hash(room)

# Role registry: the room knows about the kitchen
assert Role.roles_for(room, Kitchen)[0] is kitchen

# Room attributes are accessible directly on the kitchen (via delegation)
kitchen.floor   # → room.floor

# Role-specific attributes live on the kitchen
kitchen.utilities
```

All seems good, however I fear it will make debugging harder. In addition, this cannot support normal delegation without being a Role unlike the previous PR where I made it possible to either choose just delegation or also semantic identity sharing. That's because this approach depends on you inheriting from the type you are delegating, so if you want to delegate torso attributes for a robot you have to assume that a robot is a role for a torso and inherit from torso which is wrong.


LucaKro:
@AbdelrhmanBassiouny at lunch @tomsch420 said this can be closed because there is already a better version, do you agree with that?

AbdelrhmanBassiouny:
> @AbdelrhmanBassiouny at lunch @tomsch420 said this can be closed because there is already a better version, do you agree with that?

Yeah

PR #328, opened 2026-05-24
Tighten velocity limits proportinal

maltehue:
Adds an alternative to tighten_dof_velocity_limits_of_1dof_connections that scales all joint velocity limits by a common factor instead of applying a flat cap. This preserves the relative magnitudes defined in the URDF, which is important for robots like Tracy where matching the ratio between joint velocities is necessary for smooth rotational and translational end-effector motion. Tracy's _setup_velocity_limits is updated to use the new method as the first use case.

### DIFF
--- a/semantic_digital_twin/src/semantic_digital_twin/orm/ormatic_interface.py
+++ b/semantic_digital_twin/src/semantic_digital_twin/orm/ormatic_interface.py
@@ -52,6 +52,7 @@ import semantic_digital_twin.robots.armar
 import semantic_digital_twin.robots.armar7
 import semantic_digital_twin.robots.boxy
 import semantic_digital_twin.robots.donbot
+import semantic_digital_twin.robots.garmi
 import semantic_digital_twin.robots.hsrb
 import semantic_digital_twin.robots.icub3
 import semantic_digital_twin.robots.justin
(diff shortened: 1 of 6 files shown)


semantic_digital_twin/src/semantic_digital_twin/reasoning/robot_predicates.py
```diff
@@ -108,9 +112,12 @@ def blocking(
     :param tip: The threshold between the end effector and the position.
     :return: A list of bodies the robot is in collision with when reaching for the specified object or None if the pose or object is not reachable.
     """
-    result = root._world.compute_inverse_kinematics(
-        root=root, tip=tip, target=pose, max_iterations=1000
-    )
+    try:
```
LucaKro: as discussed, please remove this again

LucaKro (changes requested):
weird test failures, we should have a look at it together when youve got time

semantic_digital_twin/src/semantic_digital_twin/orm/ormatic_interface.py
```diff
@@ -52,6 +52,7 @@
 import semantic_digital_twin.robots.armar7
 import semantic_digital_twin.robots.boxy
 import semantic_digital_twin.robots.donbot
+import semantic_digital_twin.robots.garmi
```
ichumuh: i don't think this pr should have garmi stuff in it, no?

ichumuh (changes requested):

semantic_digital_twin/src/semantic_digital_twin/orm/ormatic_interface.py
maltehue: i removed the changes  completely, but i think they are just autogenerated when running the tests

ichumuh: yeah maybe they were generated while you had a different branch checked out? otherwise i don't see why it would create garmi files.

ichumuh (approved):

LucaKro (approved):

tomsch420 (approved):

PR #330, opened 2026-05-26
Inference explanation

AbdelrhmanBassiouny:
### Summary

  This PR introduces **Inference Explanation** for EQL — a mechanism that automatically
  records *why* and *how* every instance produced by an `inference(...)` rule was created,
  and exposes that knowledge as first-class queryable objects.

  ---

  ### Motivation

  When EQL infers an instance, there was previously no easy way to answer:

  - Which conditions in the query were satisfied?
  - Which variable nodes (and their domain values) participated?
  - Which conditions relate specific typed variables?
  - Where in the user's code was the query written?

  This PR makes all of that introspectable, without any changes to the user-facing query
  authoring API.

  ---

  ### What Changed

  #### New: `InferenceExplanation` class (`explanation/explanation.py`)

  A `Symbol` subclass (first-class entity in the `SymbolGraph`) automatically attached to
  every inferred instance via `instance._inference_explanation_`. It stores:

  - `query_node` — the `InstantiatedVariable` that produced the instance
  - `stack` — the `CallStack` captured at query-definition time
  - `query_root` — the root of the query tree
  - `satisfied_condition_ids` — the `OrderedSet[UUID]` of conditions that evaluated to `True`
  - `operation_result` — the full `OperationResult`, carrying all variable bindings

  **Design rationale for `Symbol` inheritance:**

  1. **No global registry** — the explanation's lifecycle equals the instance's lifecycle
     (RAII). Clearing the `SymbolGraph` clears all explanations automatically.
  2. **`WeakKeyDictionary` anti-pattern avoided** — the previous design stored explanations
     in a `WeakKeyDictionary`; `explanation.instance` held a strong reference back to the key,
     preventing GC. The new design stores a `weakref` on `InferenceExplanation._instance_ref`,
     breaking the cycle.
  3. **Tell Don't Ask** — callers ask the object for its own metadata, not an external registry.

  ---

  #### New: Meta-query API on `InferenceExplanation`

  All methods return EQL `Entity` descriptors that can be composed and filtered further:

  | Method | Returns |
  |---|---|
  | `get_satisfied_condition_expressions_for_the_instance()` | All satisfied condition expressions (comparators + logical wrappers) |
  | `get_variable_nodes_of_given_type(T)` | Symbolic variable nodes whose `_type_` is a subclass of `T` |
  | `get_values_of_variable_nodes_of_given_type(T)` | Concrete bound values for those nodes |
  | `get_conditions_that_relate_the_variables_of_type(T)` | Conditions with ≥ 2 distinct `T`-typed descendant nodes |
  | `get_conditions_that_relate_variables_of_types(A, B)` | Conditions with ≥ 1 `A`-typed and ≥ 1 `B`-typed descendant (symmetric) |
  | `condition_graph()` | A `QueryGraph` with satisfaction overlaid for visualization |
  | `as_string()` | Human-readable explanation with source location, query text, and satisfied conditions |

  ---

  #### New: Observer-based evaluation pipeline (`evaluation.py`)

  An aspect-oriented mechanism for hooking into `_evaluate_` without polluting core expression
  classes. Uses a `ContextVar`-backed `EvaluationContext` with four notification hooks:

  | Hook | Purpose |
  |---|---|
  | `on_evaluate_enter` | Records which expression IDs have been entered |
  | `on_result_yielded` | Stamps `satisfied_condition_ids` and `evaluated_ids` onto each `OperationResult` |
  | `on_conclusions_processed` | Builds the final `satisfied_condition_ids` set from the truth map |
  | `on_evaluate_exit` | Available for custom observers |

  Three built-in observers:

  - **`EvaluationTracker`** — accumulates evaluated expression IDs
  - **`SatisfiedConditionTracker`** — propagates satisfied condition IDs through the result chain
  - **`InferenceRecorder`** — calls `register_inference` for every `InstantiatedVariable` that
    yielded a binding

  ---

  #### New: `CallStack` / `StackFrame` (`_stack.py`)

  `StackFrame` inherits from `Symbol`, making the captured call stack queryable via EQL.
  Supports filtering by package and extracting triggering classes/functions.

  ---

  #### New: `_monitoring.py`

  Isolated sub-module holding `MonitoredRegistry` and `@monitored` decorator. Placed in a
  separate file to break the circular import between `variable.py` and `explanation.py`.

  ---

  #### Changes to `Symbol` base class (`symbol_graph.py`)

  Added a single field:

  ```python
  _inference_explanation_: Optional[InferenceExplanation] = field(
      default=None, init=False, repr=False, compare=False
  )
```
  init=False ensures it is never a constructor parameter in any subclass. The import is
  guarded by TYPE_CHECKING to prevent a circular import between symbol_graph.py and
  explanation.py.

  ---
  
  #### Changes to OperationResult and evaluation core (base_expressions.py)

  - OperationResult now carries satisfied_condition_ids: Optional[OrderedSet[UUID]]
  - sources parameter in _evaluate_ unified to always be an OperationResult (not raw
  bindings dict)
  - Removed mutable _eval_parent_ state in favour of an explicit parent argument
  - Removed per-expression truth state; truth is now computed from OperationResult via
  compute_truth

  ---
  
  #### Bug fix: @lru_cache memory leak on _get_expression_by_id_

  @lru_cache on an instance method is a class-level cache whose keys contain self,
  keeping every looked-up expression strongly alive indefinitely. Because EQL Variable
  instances store their domain data in _re_enterable_domain_generator_.materialized_values,
  keeping a Variable alive kept the entire domain — and through back-references, the World
  object — alive.

  Fixed by replacing @lru_cache with a per-instance dict in self.__dict__:
```python
  def _get_expression_by_id_(self, id_):
      cache = self.__dict__.setdefault('_expression_id_cache_', {})
      ...
```
  When the expression is GC'd, __dict__ (and the embedded cache) is collected with it.

  ---
 
  ####  Visualization: satisfaction-overlaid QueryGraph

  condition_graph() returns a QueryGraph where:

  - Satisfied condition nodes are rendered at full opacity with their type-based colour
  - Unsatisfied or short-circuited nodes are faded

  OR-branch short-circuit is correctly reflected: when the left arm of or_() succeeds, the
  right arm appears faded in the graph.

  ---

  ####   Documentation

  - User guide — krrood/doc/eql/user/inference_explanation.md: end-to-end walkthrough
  with the Drawer domain, OR short-circuit example, and all meta-query methods
  - Developer guide — krrood/doc/eql/developer/inference_explanation.md: architecture
  overview (pipeline diagram), design rationale for Symbol inheritance, weakref
  back-reference, lru_cache pitfall, and observer hook table

  ---
 
  ####  Testing

  - test_meta_queries — covers all meta-query methods, including symmetry of
  get_conditions_that_relate_variables_of_types
  - test_inference_explanation in semantic annotation tests — cross-package integration
  - All existing EQL tests pass; the evaluation pipeline refactor is backwards-compatible


krrood/doc/eql/developer/inference_explanation.md
```diff
@@ -0,0 +1,150 @@
...
+- **`InferenceRecorder`** — on `on_result_yielded`, calls `register_inference` for every
+  `InstantiatedVariable` that produced a binding.
+
+## Why `InferenceExplanation` Inherits from `Symbol`
+
+This was a deliberate design choice driven by four principles:
+
+1. **No global registry** — the explanation is stored on the inferred instance
+   (`instance._inference_explanation_`). Its lifecycle is identical to the instance's lifecycle;
+   when the instance is GC'd, the explanation is too. This satisfies the *Object Ownership / RAII*
```
tomsch420: no abbrivations >:(

krrood/doc/eql/developer/inference_explanation.md
```diff
@@ -0,0 +1,150 @@
...
+    cache: dict = self.__dict__.setdefault('_expression_id_cache_', {})
+    if id_ not in cache:
+        try:
+            cache[id_] = next(...)
+        except StopIteration:
+            raise NoExpressionFoundForGivenID(self, id_)
+    return cache[id_]
+```
+
+When the expression object is GC'd, `__dict__` (and the embedded cache) is collected with it.
```
tomsch420: >:(

krrood/doc/eql/developer/inference_explanation.md
tomsch420, line 1: this document is very very technical, im unsure if this should be part of the user guide or just in the class docstrings

krrood/src/krrood/entity_query_language/explanation/explanation.py
```diff
@@ -0,0 +1,566 @@
...
+from functools import cached_property
+from types import ModuleType
+from typing import Any, Callable, List, Optional, Type
+from uuid import UUID
+
+from ordered_set import OrderedSet
+from typing_extensions import TYPE_CHECKING
+
+# Import monitoring infrastructure from the isolated sub-module that has no
+# EQL dependencies, breaking the variable.py ↔ explanation.py import cycle.
```
tomsch420: looks like a claude leftover

krrood/src/krrood/entity_query_language/explanation/explanation.py
```diff
@@ -0,0 +1,566 @@
...
+    )
+    from krrood.entity_query_language.core.variable import (
+        InstantiatedVariable,
+        Variable,
+    )
+    from krrood.entity_query_language.query.query import Entity, Query
+
+
+def _get_query_source(frame: StackFrame) -> Optional[str]:
+    """Return the full source statement at frame.lineno, or frame.code_snippet as fallback."""
```
tomsch420: docstring

krrood/src/krrood/entity_query_language/explanation/explanation.py
```diff
@@ -0,0 +1,566 @@
...
+class ConditionAndBindings:
+    """
+    Represents a condition and its associated bindings in the inference process.
+    """
+
+    condition: SymbolicExpression
+    """
+    The condition expression.
+    """
+    bindings: dict[UUID, Any]
```
tomsch420: you have a type for Bindings already

krrood/src/krrood/entity_query_language/explanation/explanation.py
```diff
@@ -0,0 +1,566 @@
...
+
+    Inherits from :class:`~krrood.symbol_graph.symbol_graph.Symbol` so that every
+    explanation is a first-class entity in the SymbolGraph and therefore queryable
+    via EQL like any other domain object.
+
+    Lifecycle is tied to the inferred instance: the instance stores a strong reference
+    to this object via its ``_inference_explanation_`` attribute (see
+    :func:`register_inference`), while this object stores only a *weak* reference back
+    to the instance.  This means the explanation is part of the same reference cluster
+    as the instance and is collected together with it — no global registry required.
```
tomsch420, line 254: grrrr hyphen

krrood/src/krrood/entity_query_language/explanation/explanation.py
```diff
@@ -0,0 +1,566 @@
...
+
+        :param focus_package: Optional package name; only applies when ``show_trace=True``.
+        :param show_trace: When ``True``, append the call stack recorded at query-definition time.
+        :return: A formatted string explaining the inference.
+        """
+        filtered_frames = self.stack.filter().frames
+
+        def _is_krrood_internal(f: StackFrame) -> bool:
+            if f.module_name and (f.module_name == "krrood" or f.module_name.startswith("krrood.")):
+                return True
```
tomsch420: hmmm is there no way to check this using os.path

krrood/src/krrood/entity_query_language/_monitoring.py
```diff
@@ -0,0 +1,67 @@
...
+            del cls._is_monitored_
+
+    @property
+    def monitored_classes(self) -> tuple[type, ...]:
+        """Return an immutable snapshot of all monitored classes."""
+        return tuple(self._monitored)
+
+
+# Singleton — import this wherever @monitored decoration is needed.
+monitored = MonitoredRegistry()
```
tomsch420, line 79: hmm why not just use Singleton as superclass for the registriy?

krrood/src/krrood/entity_query_language/_stack.py
```diff
@@ -0,0 +1,121 @@
...
+from dataclasses import dataclass
+from typing import Callable, List, Optional
+
+
+@dataclass
+class StackFrame:
+    """A single frame in a captured call stack."""
+
+    filename: str
+    lineno: int
```
tomsch420, line 24: doc and no abbrivations

tomsch420 (changes requested):

LucaKro:
> Fixed by replacing @lru_cache with a per-instance dict in self.dict:

this sounds like what @memoize is supposed to do no?

krrood/doc/eql/developer/inference_explanation.md
```diff
@@ -0,0 +1,150 @@
...
+A `@lru_cache` applied to an instance method is a **class-level** cache. Its keys contain
+`self`, so every expression object that has ever been looked up is kept strongly referenced by
+the cache — indefinitely, because the class itself is never collected.
+
+`SymbolicExpression._get_expression_by_id_` was originally decorated with `@lru_cache`. Because
+EQL `Variable` instances store their domain data (e.g. `world.bodies`) in
+`_re_enterable_domain_generator_.materialized_values`, keeping a `Variable` alive keeps the
+entire domain (and through the `world` back-reference on each entity, the `World` object) alive.
+
+The fix is to use a **per-instance dict** stored in `self.__dict__`:
```
LucaKro: memoize?

krrood/doc/eql/developer/inference_explanation.md
LucaKro: no abbreviation, just say garbage collected please

krrood/src/krrood/entity_query_language/_stack.py
```diff
@@ -0,0 +1,121 @@
...
+
+    @classmethod
+    def from_frame_info(cls, fi: inspect.FrameInfo) -> StackFrame:
+        """
+        Eagerly extract all data from a live ``FrameInfo`` and drop the frame reference.
+
+        Must be called while the frame is still on the call stack so that
+        ``f_locals`` is populated.
+        """
+        f = fi.frame
```
LucaKro: no abbreviations

krrood/src/krrood/entity_query_language/_stack.py
```diff
@@ -0,0 +1,121 @@
...
+            function_object=fn_obj,
+            module_name=module.__name__ if module else None,
+        )
+
+
+@dataclass
+class CallStack:
+    """An ordered sequence of :class:`StackFrame` objects, innermost frame first."""
+
+    frames: List[StackFrame]
```
LucaKro, line 91: docs

krrood/src/krrood/entity_query_language/evaluation.py
```diff
@@ -0,0 +1,270 @@
...
+    Arbitrary data storage for observers to share information across events during evaluation.
+     Observers should use well-known keys defined in EvaluationContextKey to avoid collisions.
+     This is the primary mechanism for observers to maintain state across the evaluation of an expression
+     and its sub-expressions without needing to modify the expression classes or the core evaluation logic.
+     For example, the EvaluationTracker observer uses the EVALUATED_IDS_KEY to track which expressions have been 
+     evaluated in the current context, and the SatisfiedConditionTracker uses the SATISFIED_IDS_KEY to track which 
+     condition expressions have been satisfied.
+    """
+
+    def on_evaluate_enter(self, *, expression, sources):
```
LucaKro: typing, same below

krrood/src/krrood/entity_query_language/evaluation.py
```diff
@@ -0,0 +1,270 @@
...
+    This tracking is the foundation for distinguishing evaluated-from-skipped logical operators (e.g.
+    short-circuited OR/AND branches) in inference explanations.
+    """
+
+    def on_evaluate_enter(self, expression, sources):
+        from krrood.entity_query_language.core.base_expressions import (
+            OperationResult,
+        )
+
+        ctx = get_evaluation_context()
```
LucaKro: no abbreviations

krrood/src/krrood/entity_query_language/evaluation.py
```diff
@@ -0,0 +1,270 @@
...
+        if not monitored.is_monitored(type(expression)):
+            return
+        if expression._id_ not in result.bindings:
+            return
+        # Only record for InstantiatedVariable subclasses whose _evaluate__
+        # delegates to _instantiate_using_child_vars_and_yield_results_ (i.e.
+        # those that actually create new instances).  Query and its subclasses
+        # (Entity, SetOf) override _evaluate__ and merely remap bindings
+        # without creating new inferred instances.
+        from krrood.entity_query_language.core.variable import (
```
LucaKro: why local imports?

krrood/src/krrood/entity_query_language/predicate.py
```diff
@@ -78,6 +81,31 @@ def __new__(cls, *args, **kwargs):
...
+        Construct a concrete predicate instance directly, bypassing the symbolic ``__new__`` check.
+
+        Normally, calling ``cls(**kwargs)`` when any kwarg is a :class:`Selectable` redirects
+        construction to an :class:`~krrood.entity_query_language.core.variable.InstantiatedVariable`
+        so the call can be represented as a symbolic expression in a query graph.  During evaluation,
+        however, the bound values themselves may be :class:`Selectable` objects (e.g. in a
+        meta-query that reasons about EQL nodes).  In that case the redirect is wrong — we want the
+        real predicate instance so its :meth:`__call__` can be evaluated immediately.
+
+        :meth:`_construct_normally_` is the escape hatch for that situation.  It calls
```
LucaKro, line 96: whats the point of this info? it doesnt give additional information compared to just reading the method. also this type of docstring is the one that will end up lying to you imo, as it describes details about the current implementation at the time of writing, but will probably not get updated sufficiently due to its complexity when the implementation changes

krrood/src/krrood/entity_query_language/query_graph.py
```diff
@@ -72,6 +113,51 @@ def __post_init__(self):
...
+        at least one unsatisfied condition node.  We compute this by BFS from the
+        root, refusing to traverse *through* unsatisfied condition nodes, then
+        marking every node the BFS did **not** reach as faded.
+
+        Exception: descendants of a *satisfied* QuantifiedConditional (Exists/ForAll)
+        are always reachable even though their internal condition IDs are not tracked
+        in the outer satisfied_condition_ids set.  The BFS carries a skip_gate flag
+        for exactly this case.
+        """
+        from krrood.entity_query_language.operators.logical_quantifiers import (
```
LucaKro, line 116: local imports everywhere

krrood/src/krrood/entity_query_language/utils.py
```diff
@@ -109,26 +109,27 @@ def make_set(value: Any) -> Set:
...
     passing the bindings from the previous evaluated generator to the next.
 
     :param expressions: The symbolic expressions to evaluate.
-    :param sources: The current bindings.
-    :param parent: The parent expression.
+    :param sources: The current OperationResult carrying bindings, or None.
     :return: An Iterable of Bindings for each combination of values.
     """
+    def _make_stage(inner_expression):
+        def stage(prev: Optional[OperationResult]):
```
LucaKro: what is meant by stage?

krrood/src/krrood/entity_query_language/explanation/explanation.py
```diff
@@ -0,0 +1,566 @@
...
+        tree = ast.parse(source)
+        candidates: list = list(tree.body)
+        for node in ast.walk(tree):
+            if isinstance(node, (ast.FunctionDef, ast.AsyncFunctionDef, ast.ClassDef)):
+                candidates.extend(node.body)
+        best = None
+        for node in candidates:
+            start = getattr(node, "lineno", None)
+            end = getattr(node, "end_lineno", None)
+            if start is not None and end is not None and start <= frame.lineno <= end:
```
LucaKro: less nesting by inverting condition and continuing

krrood/src/krrood/entity_query_language/explanation/explanation.py
```diff
@@ -0,0 +1,566 @@
...
+        Each ``QueryNode`` carries an ``is_satisfied`` flag grounded directly on
+        the satisfied condition IDs.  Unsatisfied condition subtrees are also
+        marked as *faded* for visualization purposes.
+
+        :return: A :class:`QueryGraph` instance, or None if no conditions exist
+            or no satisfaction data is available.
+        """
+        if self.query_root is None or not self.satisfied_condition_ids:
+            return None
+        from krrood.entity_query_language.query_graph import QueryGraph
```
LucaKro: local import necessary?

LucaKro (changes requested):

krrood/doc/eql/developer/inference_explanation.md
AbdelrhmanBassiouny: Well if you look carefully this is in the developers guide haha

AbdelrhmanBassiouny: Anyway, I also refactored it to be at a more architectural level.

krrood/src/krrood/entity_query_language/explanation/explanation.py
AbdelrhmanBassiouny: What's wrong with that?

krrood/src/krrood/entity_query_language/explanation/explanation.py
AbdelrhmanBassiouny: hmm good point lemme check

krrood/src/krrood/entity_query_language/_monitoring.py
AbdelrhmanBassiouny: yeah would work

krrood/doc/eql/developer/inference_explanation.md
AbdelrhmanBassiouny: Unfortunately, memoize interacts badly with eql in debug mode.

AbdelrhmanBassiouny: Also this was a stal doc, not it doesn't use __dict__ it uses an explicit field for storing the mapping between ids and variables.

krrood/src/krrood/entity_query_language/evaluation.py
AbdelrhmanBassiouny: I do not think those are needed, as they are mentioned already in the abstract class. I think this would increase maintenance effort when refactoring that, or even misleading types if forgotten.

krrood/src/krrood/entity_query_language/evaluation.py
AbdelrhmanBassiouny: Yeah, I think that is unfortunately unavoidable here, as the observers check on the types of the expressions, and the base expression imports the observer to actually invoke them on the different evaluation stages, but I will see what chat says about that, maybe there's a way I don't know of.

krrood/src/krrood/entity_query_language/predicate.py
AbdelrhmanBassiouny: I agree that the last part of it is not needed where it mentions where it is used, but I think the doc here is describing something very relevant to the Predicate class functionality, and why this method is introduced, so without this method `meta-queries` with predicates will not worlk correctly becuase not the data you are querying is itself variables so inside the EQL evaluation when I am trying to actually bind the predicate parameters and evaluate it this will cause again a symbolic construction instead of actual instantiation,  in that case it should actually be constructed normally and I know that inside of EQL evaluation and I needed this method because of that.

krrood/src/krrood/entity_query_language/evaluation.py
AbdelrhmanBassiouny: I was able to remove most of them, but one or two remain.

krrood/src/krrood/entity_query_language/evaluation.py
LucaKro: I personally still prefer the typing to be there, because when i look for the typing of a class i usually look at the implementation, and having to look up the abstract class is usually annoying, but i see your point

krrood/doc/eql/developer/inference_explanation.md
```diff
@@ -13,138 +13,265 @@ kernelspec:
...
-                                   EvaluationTracker
-                                   InferenceRecorder
-                                           │
-                                           ▼ on_result_yielded
-                                   register_inference(instance, variable_node, result)
-                                           │
-                                           ▼
-                              instance._inference_explanation_ = InferenceExplanation(...)
+┌─────────────────────────────────────────────────────────────────────┐
+│  LAYER 1: Call-stack capture (krrood/entity_query_language/_stack.py,  │
```
tomsch420: mermaid diagram time?

krrood/src/krrood/entity_query_language/_monitoring.py
```diff
@@ -59,9 +66,10 @@ def unregister(self, cls: Type) -> None:
...
-        """Return an immutable snapshot of all monitored classes."""
+        """
+        :return: An immutable snapshot of all currently monitored classes as a tuple.
+        """
         return tuple(self._monitored)
 
-
-# Singleton — import this wherever @monitored decoration is needed.
+# The decorator to use for classes to be monitored.
 monitored = MonitoredRegistry()
```
tomsch420, line 79: i dont think we need this then anymore

krrood/src/krrood/entity_query_language/enums.py
```diff
@@ -56,7 +56,7 @@ class DomainSource(Enum):
     """
 
 
-class EvaluationContextKey(StrEnum):
+class EvaluationContextKey(str, Enum):
```
tomsch420: StrEnum is correct here

krrood/src/krrood/entity_query_language/utils.py
```diff
@@ -185,7 +199,10 @@ def get_function_argument_names(function: Callable) -> List[str]:
 
 
 def merge_args_and_kwargs(
-    function_or_class: Union[Callable, Type], args, kwargs, ignore_first: bool = False
+    function_or_class: Union[Callable, Type],
```
tomsch420: just callable

tomsch420 (changes requested):

krrood/src/krrood/entity_query_language/_monitoring.py
AbdelrhmanBassiouny: yeah but I am used to decorators being small case words

tomsch420 (approved):

krrood/src/krrood/entity_query_language/_monitoring.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+@dataclass
+class MonitoredRegistry(metaclass=SingletonMeta):
+    """
+    Registry for monitoring EQL object creation stacks.
+    Acts as a class decorator and provides lookup methods.
+    """
+
+    _monitored: set[type] = field(default_factory=set)
```
LucaKro, line 28: docs

LucaKro (approved):
good job, please Just add docs for this one attribute

PR #333, opened 2026-05-28
subclass safe generic fix for multiple generics

LucaKro:
# Feature: Improved Generic Type Handling and Robust `SubClassSafeGeneric`

This PR refactors `AbstractSubClassSafeGeneric` and related utilities to properly handle complex generic inheritance patterns, multiple generic parameters, and transitive type variable resolution. It also cleans up duplicated utility functions and ensures dataclass field metadata is preserved during type updates.

Theoretical Motivation:
Generic programming in Python's typing system can become complex when dealing with deep inheritance trees and multiple generic parameters. `SubClassSafeGeneric` aims to automate the update of dataclass field types based on bound generic arguments, but required a more robust mechanism to track substitutions across multiple bases and resolve them transitively.

Practical Reasoning:
Robot models and entity queries often involve complex generic structures (e.g., `CombinedThreeGenericSubClassSafe`). The previous implementation failed to correctly map types in these scenarios, leading to incorrect type hints and runtime issues in the entity query language. This PR ensures that type safety is maintained even in complex hierarchies.

- **Robust Generic Substitutions**: Supports multiple generic bases and multiple type parameters per base.
- **Transitive Type Resolution**: Correctly resolves chains of `TypeVar` (e.g., `T` -> `U` -> `int`) to their most concrete form.
- **Dataclass Field Preservation**: Updates field types while preserving existing metadata like `default_factory` and `kw_only`.

---

### `krrood.patterns.subclass_safe_generic`

- Introduced `AbstractSubClassSafeGeneric` to encapsulate the substitution and field-update logic.
- Implemented `_get_generic_type_substitutions` with support for multiple `__orig_bases__`, mapping root `TypeVar`s to concrete arguments correctly.
- Added `_resolve_substitutions_transitively` to handle nested or chained generic parameters.
- Enhanced `_update_field_kwargs` to correctly handle inherited fields and ensure that both the field object and `__annotations__` are updated.

### `krrood.utils`

- Refactored `get_generic_type_param` into a more robust `get_generic_type_params`, which can now include or exclude root and specialized generic parameters.
- Centralized `is_hashable` and `ensure_hashable` from `entity_query_language.utils`.
- Improved `memoize` decorator to use a dedicated `TypeVar` (`F`) to avoid potential naming collisions with the commonly used `T`.

---

- **`krrood.class_diagrams.utils.get_generic_type_param`**: Removed in favor of the centralized version in `krrood.utils`.

---

- **Renamed `get_generic_type_param` to `get_generic_type_params`**: The utility now returns a list of parameters and has an updated signature. Internal calls have been updated, but external consumers will need to adjust.
- **Moved `ensure_hashable` and `is_hashable`**: These were moved from `krrood.entity_query_language.utils` to `krrood.utils`.

---

## Other Changes

- Added comprehensive new tests in `test_subclass_safe_generic.py` covering multi-base, deep inheritance, and transitive resolution scenarios.
- Cleaned up imports and improved code style consistency across modified modules.

---

### DIFF
--- a/krrood/src/krrood/patterns/subclass_safe_generic.py
+++ b/krrood/src/krrood/patterns/subclass_safe_generic.py
@@ -24,10 +24,10 @@ from krrood.class_diagrams.utils import (
     get_and_resolve_generic_type_hints_of_object_using_substitutions,
     resolve_type,
 )
-from krrood.entity_query_language.utils import ensure_hashable
 from krrood.utils import (
     get_generic_type_params,
     T,
+    ensure_hashable,
 )
 
 if TYPE_CHECKING:
@@ -143,7 +143,22 @@ class AbstractSubClassSafeGeneric(ABC):
         if "_subclass_safe_substitutions" in cls.__dict__:
             return cls._subclass_safe_substitutions
 
-        substitutions = cls.get_superclass_generic_type_substitution()
+        substitutions = {}
+        for base in getattr(cls, "__orig_bases__", []):
+            base_origin, resolved_types = cls._resolve_base_origin_and_arguments(base)
+            if base_origin is None:
+                continue
+
+            # Map the root TypeVars of the base to the concrete arguments provided here
+            root_parameters = base_origin.get_generic_types()
+            for old_type, new_type in zip(root_parameters, resolved_types):
+                if old_type is not new_type and new_type is not None:
+                    substitutions[ensure_hashable(old_type)] = new_type
+
+            # Recursively pull substitutions already defined by the parent
+            if base_origin is not cls:
+                substitutions.update(base_origin._get_generic_type_substitutions())
+
         if substitutions:
             substitutions = cls._resolve_substitutions_transitively(substitutions)
 
@@ -175,36 +190,6 @@ class AbstractSubClassSafeGeneric(ABC):
         return resolved_substitutions
 
     @classmethod
-    def get_superclass_generic_type_substitution(cls) -> Dict[Type, Type]:
-        """
-        Retrieve generic type substitutions from superclasses.
-
-        :return: A mapping from each superclass generic type to the resolved type in this class.
-        """
-        substitutions = {}
-        if not hasattr(cls, "__orig_bases__"):
-            return substitutions
-
-        for base in cls.__orig_bases__:
-            base_origin, resolved_types = cls._resolve_base_origin_and_arguments(base)
-            if base_origin is None:
-                continue
-
-            # Map the root TypeVars of the base to the concrete arguments provided here
-            root_parameters = base_origin.get_generic_types(True, False)
-            if not root_parameters:
-                root_parameters = base_origin.get_generic_types(False, True)
-
-            for old_type, new_type in zip(root_parameters, resolved_types):
-                if old_type is not new_type and new_type is not None:
-                    substitutions[ensure_hashable(old_type)] = new_type
-
-            # Recursively pull substitutions already defined by the parent
-            if base_origin is not cls:
-                substitutions.update(base_origin._get_generic_type_substitutions())
-        return substitutions
-
-    @classmethod
     def _resolve_base_origin_and_arguments(
         cls, base: Any
     ) -> Tuple[Optional[Type], Tuple[Type, ...]]:
@@ -227,6 +212,7 @@ class AbstractSubClassSafeGeneric(ABC):
         return None, ()
 
     @classmethod
+    @lru_cache
     def get_generic_types(
         cls,
         from_root_generic_base: bool = True,
--- a/krrood/src/krrood/patterns/subclass_safe_generic.py
+++ b/krrood/src/krrood/patterns/subclass_safe_generic.py
@@ -18,6 +18,7 @@ from typing_extensions import (
     List,
     get_origin,
     get_args,
+    assert_never,
 )
 
 from krrood.class_diagrams.utils import (
@@ -28,6 +29,7 @@ from krrood.utils import (
     get_generic_type_params,
     T,
     ensure_hashable,
+    memoize,
 )
 
 if TYPE_CHECKING:
@@ -148,14 +150,24 @@ class AbstractSubClassSafeGeneric(ABC):
                 continue
 
             # Map the root TypeVars of the base to the concrete arguments provided here
-            root_parameters = base_origin.get_generic_types()
-            for old_type, new_type in zip(root_parameters, resolved_types):
-                if old_type is not new_type and new_type is not None:
+            if resolved_types:
+                root_parameters = base_origin.get_generic_types(True, False)
+                if len(root_parameters) != len(resolved_types):
+                    assert_never(base)
+
+                for old_type, new_type in zip(root_parameters, resolved_types):
+                    if (
+                        not isinstance(old_type, TypeVar)
+                        or old_type is new_type
+                        or new_type is None
+                    ):
+                        continue
                     substitutions[ensure_hashable(old_type)] = new_type
 
             # Recursively pull substitutions already defined by the parent
-            if base_origin is not cls:
-                substitutions.update(base_origin._get_generic_type_substitutions())
+            if base_origin is cls:
+                continue
+            substitutions.update(base_origin._get_generic_type_substitutions())
 
         if substitutions:
             substitutions = cls._resolve_substitutions_transitively(substitutions)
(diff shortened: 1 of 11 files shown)


krrood/src/krrood/patterns/subclass_safe_generic.py
```diff
@@ -66,89 +75,170 @@ def _update_field_kwargs(
...
         else:
-            # If not, check if there's an existing field that needs to be updated
-            field_ = copy(next((f for f in fields(cls) if f.name == name), None))
-            if field_ is not None:
-                for key, value in kwargs.items():
-                    setattr(field_, key, value)
-                setattr(cls, field_.name, field_)
-            else:
-                setattr(cls, name, field(**kwargs))
+            # Create a new field if it doesn't exist
```
AbdelrhmanBassiouny: Check if this is actually needed

krrood/src/krrood/patterns/subclass_safe_generic.py
```diff
@@ -66,89 +75,170 @@ def _update_field_kwargs(
...
+        elif existing_field is not None:
+            resolved_type = existing_field.type
         else:
-            cls.__annotations__[name] = Any
+            resolved_type = Any
+        cls.__annotations__[name] = resolved_type
 
     @classmethod
-    def _get_generic_base(cls) -> Optional[Type]:
+    def _get_existing_field(cls, name: str) -> Optional[Field]:
```
AbdelrhmanBassiouny: This is generally useful to know the owner of the field.

krrood/src/krrood/patterns/subclass_safe_generic.py
```diff
@@ -66,89 +75,170 @@ def _update_field_kwargs(
...
+        if cls is AbstractSubClassSafeGeneric:
             return {}
-        generic_base = cls._get_generic_base()
-        if generic_base is None:
-            return {}
-        for base in cls.__bases__:
-            if not isinstance(base, type) or not issubclass(base, generic_base):
+
+        # Use a class-level cache to avoid redundant recursive calculations
+        if "_subclass_safe_substitutions" in cls.__dict__:
```
AbdelrhmanBassiouny, line 127: maybe that can be a classvar

krrood/src/krrood/patterns/subclass_safe_generic.py
```diff
@@ -66,89 +75,170 @@ def _update_field_kwargs(
...
+                    break
+                if resolution.resolved_type is current_type:
+                    break
+                current_type = resolution.resolved_type
+            resolved_substitutions[old_type] = current_type
+        return resolved_substitutions
+
+    @classmethod
+    def _resolve_base_origin_and_arguments(
+        cls, base: Any
```
AbdelrhmanBassiouny: Type instead of Any

krrood/src/krrood/patterns/subclass_safe_generic.py
```diff
@@ -66,89 +75,170 @@ def _update_field_kwargs(
...
+        """
+        :param include_root_generic_base: Whether to include type parameters the class gets from its own typing.Generic directly.
+        :param include_specialized_generic_base: Whether to include type parameters from superclasses that are generic, which are not typing.Generic.
         :return: The concrete generic type parameters bound for this class, in declaration order.
         """
-        generic_base = cls._get_generic_base()
-        if generic_base is None:
-            return None
-        return get_generic_type_param(cls, generic_base)
+        return get_generic_type_params(
```
AbdelrhmanBassiouny: maybe just use this function directly

krrood/src/krrood/utils.py
```diff
@@ -846,16 +887,16 @@ def _handle_import_from_node(
     return package_name
 
 
-T = TypeVar("T", bound=Callable[..., Any])
+F = TypeVar("F", bound=Callable[..., Any])
```
AbdelrhmanBassiouny: TCallable

AbdelrhmanBassiouny (changes requested):
Nice work! 💪

krrood/src/krrood/patterns/subclass_safe_generic.py
LucaKro: removed

krrood/src/krrood/patterns/subclass_safe_generic.py
LucaKro: tried that, but then subclasses inherit the cache causing tests to fail

AbdelrhmanBassiouny (approved):

krrood/src/krrood/utils.py
```diff
@@ -614,31 +614,89 @@ def run_subprocess_on_file(command: List[str]):
...
+    :return: True if the object is hashable, False otherwise.
+    """
+    try:
+        hash(obj)
+        return True
+    except TypeError:
+        return False
+
+
+def ensure_hashable(obj) -> Hashable:
```
Tigul, line 691: Isn't the return a little weird here, the id is usually an int. So the method either bounces the obj or returns and int

Tigul (approved):

krrood/src/krrood/utils.py
LucaKro: True, but this is nothing i added in this PR, its an existing functionality by @AbdelrhmanBassiouny maybe he has an opinion on that?

krrood/src/krrood/patterns/subclass_safe_generic.py
```diff
@@ -66,89 +75,144 @@ def _update_field_kwargs(
...
+            # Map the root TypeVars of the base to the concrete arguments provided here
+            if resolved_types:
+                root_parameters = get_generic_type_params(
+                    base_origin,
+                    AbstractSubClassSafeGeneric,
+                    include_root_generic_base=True,
+                    include_specialized_generic_base=False,
+                )
+                if len(root_parameters) != len(resolved_types):
+                    raise TypeError(
```
tomsch420: is this really a type error? type errors are usually expected str, got int.

krrood/src/krrood/patterns/subclass_safe_generic.py
```diff
@@ -66,89 +75,144 @@ def _update_field_kwargs(
...
+        :return: A new substitution map with fully resolved types.
         """
-        generic_base = cls._get_generic_base()
-        if generic_base is None:
-            return None
-        return get_generic_type_param(cls, generic_base)
+        resolved_substitutions = {}
+        for old_type, new_type in substitutions.items():
+            current_type = new_type
+            # Limit iterations to avoid infinite loops in case of circular references
```
tomsch420: why can this be circular? even if, this is a hacky way to solve circular stuff.  Use a visited list and a queue or something like that

krrood/src/krrood/patterns/subclass_safe_generic.py
```diff
@@ -173,7 +237,7 @@ def get_generic_type(cls) -> Optional[Type[T]]:
         """
         :return: The type of the role taker.
         """
-        generic_types = get_generic_type_param(cls, SubClassSafeGeneric)
-        if generic_types:
-            return generic_types[0]
+        generic_types = get_generic_type_params(cls, SubClassSafeGeneric)
+        for generic_type in generic_types:
```
tomsch420: i thin the old version of this was more readable with the explicit index return

krrood/src/krrood/utils.py
tomsch420, line 691: pretty sure the return value is int always

tomsch420 (changes requested):

krrood/src/krrood/utils.py
LucaKro: no, it just returns the int of the memory id, or the object itself. but also i just moved the method

krrood/src/krrood/patterns/subclass_safe_generic.py
LucaKro: oh yea fair i missed this tbh

krrood/src/krrood/patterns/subclass_safe_generic.py
LucaKro: added dedicated error

krrood/src/krrood/patterns/subclass_safe_generic.py
LucaKro: ill have a look at it again

krrood/src/krrood/utils.py
tomsch420: memory id is int. you can check with e. g. type(id("a"))

krrood/src/krrood/patterns/subclass_safe_generic.py
tomsch420: chat does this very often, but i have never encountered a case where this "fix" was the best thing to do :D

krrood/src/krrood/utils.py
LucaKro: yes, and int is hashable. you can check with issubclass(int, Hashable).

ensure hashable doesnt always return the int, if "is_hashable" is true, then it will just return the object. so returning just int here doesnt make any sense.

sadly in python, not all hashable objects are actually covered by issubclass(clazz, Hashable), so we cannot just check that. the only (as far as i know) reliable way is to call __hash__ on the object, and see if it throws a typeerror (which is what happens in is_hashable). 

So technically the typing "Hashable" is indeed wrong, but it accurately conveys to the typechecker and anyone reading the method that whatever comes back from ensure_hashable, i will be able to call hash on it.

What else do you think we should type it as? maybe @AbdelrhmanBassiouny also has input on this, as he implemented the method, while i just moved it

tomsch420: oopsi poopsi. my mistake

tomsch420 (approved):

PR #337, opened 2026-06-01
Fix generate_orm script in pycram and include script in conftest

J-Schaefer:
The script used the current workdir before and relied on that path. It can be executed from everywhere now. Also included the script in conftest.py so the ormatic_interface get rebuild before testing.
This change add a new dependency to generate_orm.py but pathlib is necessary for other modules in the project anyway.


J-Schaefer:
One small issue remains: the import `from pycram.orm.ormatic_interface import Base` fails if the ormatic_interface.py is actually empty or doesn't exist. Tried to put the include in the method that needs it (`pycram_testing_session`) but that also fails. How can I fix that?

tomsch420:
> One small issue remains: the import `from pycram.orm.ormatic_interface import Base` fails if the ormatic_interface.py is actually empty or doesn't exist. Tried to put the include in the method that needs it (`pycram_testing_session`) but that also fails. How can I fix that?

can you point to the location where it happens?

test/coraplex_test/conftest.py
```diff
@@ -16,6 +18,16 @@
 from semantic_digital_twin.robots.pr2 import PR2
 
 
+def pytest_configure(config):
```
AbdelrhmanBassiouny, line 22: I think @Tigul was against doing this for some reason.

J-Schaefer: Isn't this implementation exactly how it's done in test/semantic_digital_twin_test/conftest.py?
Or do you mean doing this at all for pycram? I asked @Tigul and he mentioned if I want to add this I can go ahead. I'll check with @Tigul again.

J-Schaefer:
> > One small issue remains: the import `from pycram.orm.ormatic_interface import Base` fails if the ormatic_interface.py is actually empty or doesn't exist. Tried to put the include in the method that needs it (`pycram_testing_session`) but that also fails. How can I fix that?
> 
> can you point to the location where it happens?

What do you mean "where it happens"?

When I run the tests with `pytest test/pycram_test` and the file `pycram/src/pycram/orm/ormatic_interface.py` is empty, I get this error

```bash
$ pytest test/pycram_test
`polytope` failed to import `cvxopt.glpk`.
will use `scipy.optimize.linprog`
ImportError while loading conftest '/home/user/libs/cognitive_robot_abstract_machine/test/pycram_test/conftest.py'.
test/pycram_test/conftest.py:13: in <module>
    from pycram.orm.ormatic_interface import Base
E   ImportError: cannot import name 'Base' from 'pycram.orm.ormatic_interface' (/home/user/libs/cognitive_robot_abstract_machine/pycram/src/pycram/orm/ormatic_interface.py)
```

When the file doesn't exist at all

```bash
$ pytest test/pycram_test
`polytope` failed to import `cvxopt.glpk`.
will use `scipy.optimize.linprog`
ImportError while loading conftest '/home/user/libs/cognitive_robot_abstract_machine/test/pycram_test/conftest.py'.
test/pycram_test/conftest.py:13: in <module>
    from pycram.orm.ormatic_interface import Base
E   ModuleNotFoundError: No module named 'pycram.orm.ormatic_interface'

```

Which both make sense, but I would like to cover this case as well.

test/coraplex_test/conftest.py
Tigul: I'm fine with that. Makes the tests more reliable and it is done in the CI anyways

Tigul (approved):

J-Schaefer:
Fixed merge conflicts and fixed import in pycram_test/conftest.py. Please rereview @Tigul @AbdelrhmanBassiouny @tomsch420 
Tests are failing on main as well. Not because of changes in this branch.

LucaKro:
needs merge with main

LucaKro (approved):

PR #341, opened 2026-06-04
Integrating rerun visualizer with semdt

AbdElRahmanFarhan:
# Integrate Rerun visualizer with Semantic Digital Twin

## Summary

Adds a [Rerun](https://rerun.io/) visualization adapter to the Semantic Digital Twin, providing a ROS-free alternative to the RViz. It is also Web-based so it can be easily integrated with other applications.

## What's included

**Rerun adapter** (`adapters/rerun/`)
- `rerun_logger.py` — stateless logging helpers:
  - `log_model` logs each body's static visual geometry (falling back to collision shapes) as `rerun.Mesh3D` in the body's local link frame, so only transforms need re-logging on a state change.
  - `log_state` logs the current forward-kinematics transform of every body; `static=True` overwrites in place for a constant-memory, current-state-only view.
  - `log_world` logs a full snapshot (geometry + state).
  - `shape_to_link_frame_mesh` bakes the shape origin into the vertices and normalizes Collada `TextureVisuals` to per-vertex colors.
- `rerun_visualizer.py` — `RerunVisualizer`, a callback-driven live view mirroring the ROS visualization adapter: a model-change callback re-logs static geometry and an owned state-change callback re-logs transforms as the world changes. Configurable output via `RerunSink` (`SPAWN`, `CONNECT`, `SAVE`, `NONE`), optional scrubbable state history, and viewer memory limits.

**Mesh color serialization** (`world_description/geometry.py`)
- `Mesh.to_json` bakes materials/textures down to per-vertex colors (`to_color()`) and writes `vertex_colors` into the mesh dict, so a mesh's color survives serialization across the ROS world synchronizer. `_from_json` restores `vertex_colors`, and exports via PLY when colors are present so they survive the temporary export/reload.
- The `Mesh.mesh` property now applies the shape's `self.color` only when it was explicitly set, preserving a mesh's own materials / per-vertex colors (e.g. from a `.dae` or from deserialization) by default. This removed the need for a separate mesh resolver on the Rerun side.

**URDF xacro mappings** (`adapters/urdf.py`)
- `URDFParser.from_xacro` accepts a `mappings` dict, forwarded to `xacro.process_file` (needed to select e.g. `ur_type` / `name` when loading the UR description).

**Example scripts** (`scripts/`)
- `rerun_visualizer_demo.py` — loads a UR5 from xacro, spawns a Rerun viewer, and animates the joints (standalone, no ROS).
- `rerun_distributed_publisher.py` / `rerun_distributed_viewer.py` — distributed setup where the publisher owns the authoritative world and serves it over ROS (`FetchWorldServer` + state/model synchronizers), and the viewer mirrors it and renders in Rerun.

**Packaging** (`pyproject.toml`)
- New optional `rerun` extra (`pip install semantic_digital_twin[rerun]`) pulling in `rerun-sdk`.

## Tests

`test/rerun_test/test_rerun_adapter.py` covers colored-geometry conversion, origin baking, the log functions running end to end, mesh color surviving round-trip serialization, and visualizer callback registration/teardown. All 5 pass.


semantic_digital_twin/scripts/rerun_distributed_publisher.py
```diff
@@ -0,0 +1,95 @@
...
+    """Parse the UR robot xacro into a world.
+
+    :return: The populated world.
+    """
+    return URDFParser.from_xacro(
+        UR_XACRO, mappings={"ur_type": UR_TYPE, "name": UR_TYPE}
+    ).parse()
+
+
+def movable_connections(world: World) -> List[Connection]:
```
LucaKro: this should be a property "active_connections" in world, where you filter for ActiveConnection instead of raw_dof

semantic_digital_twin/scripts/rerun_distributed_publisher.py
```diff
@@ -0,0 +1,95 @@
...
+
+UR_XACRO = os.path.join(
+    get_package_share_directory("ur_description"),
+    "urdf",
+    "ur.urdf.xacro",
+)
+"""Path to the UR description xacro (requires the ``ur_description`` package)."""
+
+
+def load_example_world() -> World:
```
LucaKro: I dont like those kind of oneliners

semantic_digital_twin/scripts/rerun_distributed_publisher.py
```diff
@@ -0,0 +1,95 @@
...
+
+    :param world: The world to inspect.
+    :return: The list of connections with a settable position.
+    """
+    return [
+        connection for connection in world.connections if hasattr(connection, "raw_dof")
+    ]
+
+
+def wiggle(world: World, duration_seconds: float = 10.0) -> None:
```
LucaKro: this is just to basically show position updates of active connections are working? i think it should go in semdt utils, with a better name than "wiggle"

semantic_digital_twin/scripts/rerun_distributed_publisher.py
```diff
@@ -0,0 +1,95 @@
...
+    start = time.time()
+    while time.time() - start < duration_seconds:
+        phase = time.time() - start
+        for index, connection in enumerate(connections):
+            connection.position = 0.6 * math.sin(phase + index)
+        time.sleep(0.02)
+
+
+def main() -> None:
+    """Own the authoritative world and serve it over ROS until interrupted."""
```
LucaKro: this script doesnt do anything special imo, I dont see why i needs to be its own script, instead of just a demo in experiments. This is not tied in any way to rerun, one could use this for other things, and one could also use other demos to test rerun no?

semantic_digital_twin/scripts/rerun_distributed_viewer.py
```diff
@@ -0,0 +1,45 @@
...
+
+from semantic_digital_twin.adapters.rerun import RerunSink, RerunVisualizer
+from semantic_digital_twin.adapters.ros.world_fetcher import fetch_world_from_service
+from semantic_digital_twin.adapters.ros.world_synchronizer import (
+    ModelSynchronizer,
+    StateSynchronizer,
+)
+
+
+def main() -> None:
```
LucaKro: I also think this fits into experiments better

semantic_digital_twin/scripts/rerun_distributed_viewer.py
```diff
@@ -0,0 +1,45 @@
...
+    ModelSynchronizer(_world=world, node=node)
+    RerunVisualizer(
+        _world=world,
+        application_id="semdt_rerun_viewer",
+        sink=RerunSink.SPAWN,
+        state_history=False,
+        memory_limit="1GB",
+    )
+
+    print("Viewer running; updates follow the publisher. Ctrl-C to stop.")
```
LucaKro: use a logger

semantic_digital_twin/scripts/rerun_visualizer_demo.py
```diff
@@ -0,0 +1,74 @@
...
+    """Parse the UR robot xacro into a world.
+
+    :return: The populated world.
+    """
+    return URDFParser.from_xacro(
+        UR_XACRO, mappings={"ur_type": UR_TYPE, "name": UR_TYPE}
+    ).parse()
+
+
+def movable_connections(world: World) -> List[Connection]:
```
LucaKro: this, and the stuff above is duplicated code

semantic_digital_twin/scripts/rerun_visualizer_demo.py
```diff
@@ -0,0 +1,74 @@
...
+        phase = time.time() - start
+        for index, connection in enumerate(connections):
+            connection.position = 0.6 * math.sin(phase + index)
+        time.sleep(0.02)
+
+
+def main() -> None:
+    """Load the world, spawn a Rerun viewer, and animate the joints."""
+    world = load_example_world()
+    RerunVisualizer(
```
LucaKro: whats the difference between this and the other rerun viewer script?

semantic_digital_twin/src/semantic_digital_twin/adapters/rerun/__init__.py
LucaKro, line 1: Do we really need this? im a fan of just keeping __init__.py files clean and empty

semantic_digital_twin/src/semantic_digital_twin/adapters/rerun/rerun_logger.py
```diff
@@ -0,0 +1,126 @@
...
+DEFAULT_ROOT_ENTITY_PATH = "world"
+"""Default Rerun entity path under which the kinematic tree is logged."""
+
+
+def _entity_path(root_entity_path: str, body: Body) -> str:
+    """Return the Rerun entity path for a body."""
+    return f"{root_entity_path}/{body.name.name}"
+
+
+def _visual_shapes(body: Body) -> List[Shape]:
```
LucaKro: I dont like functions that are just wrappers for a single line of code. i dont see what benefit this gives you. 
If you were to keep this, tho i advice against it, this would be a property for Body, not a function here

semantic_digital_twin/src/semantic_digital_twin/adapters/rerun/rerun_logger.py
```diff
@@ -0,0 +1,126 @@
...
+    """Return a body's visual shapes, falling back to collision shapes."""
+    return body.visual.shapes if body.visual.shapes else body.collision.shapes
+
+
+def shape_to_link_frame_mesh(shape: Shape) -> trimesh.Trimesh:
```
LucaKro: why do we need this? the geometry should already be in the origin frame.
And again, if this is not supported right now, then this should be added to the shape class, not be a function here

semantic_digital_twin/src/semantic_digital_twin/adapters/rerun/rerun_logger.py
```diff
@@ -0,0 +1,126 @@
...
+
+import rerun as rr
+import trimesh
+from typing_extensions import List, Optional
+
+from semantic_digital_twin.world import World
+from semantic_digital_twin.world_description.geometry import Shape
+from semantic_digital_twin.world_description.world_entity import Body
+
+DEFAULT_ROOT_ENTITY_PATH = "world"
```
LucaKro: is this arbitrarily chosen or is this predefined by rerun?

semantic_digital_twin/src/semantic_digital_twin/adapters/rerun/rerun_visualizer.py
```diff
@@ -0,0 +1,118 @@
...
+    log_model,
+    log_state,
+)
+from semantic_digital_twin.callbacks.callback import (
+    ModelChangeCallback,
+    StateChangeCallback,
+)
+
+
+class RerunSink(Enum):
```
LucaKro: Im not sure if i like Sink here, maybe just RerunMode?

semantic_digital_twin/src/semantic_digital_twin/adapters/rerun/rerun_visualizer.py
```diff
@@ -0,0 +1,118 @@
...
+    CONNECT = "connect"
+    """Stream to an already-running viewer over gRPC (uses ``sink_target`` URL)."""
+    SAVE = "save"
+    """Write an ``.rrd`` recording file (uses ``sink_target`` path)."""
+    NONE = "none"
+    """Do not configure a sink; the caller manages the recording's output."""
+
+
+@dataclass(eq=False)
+class _RerunStateCallback(StateChangeCallback):
```
LucaKro: why a private class?

semantic_digital_twin/src/semantic_digital_twin/adapters/urdf.py
```diff
@@ -169,11 +169,16 @@ def from_file(
         return urdf_parser
 
     @classmethod
-    def from_xacro(cls, xacro_path: str, prefix: Optional[str] = None) -> URDFParser:
+    def from_xacro(
+        cls,
+        xacro_path: str,
+        prefix: Optional[str] = None,
+        mappings: Optional[Dict[str, str]] = None,
```
LucaKro, line 159: what is this used for? docs please

semantic_digital_twin/src/semantic_digital_twin/adapters/rerun/rerun_visualizer.py
LucaKro, line 1: Im not convinced this needs its own package, i think you can probably just merge the logger and visualizer module and have it in a "rerun.py" module. Also i think the stuff that is currently in the logger.py file (minus the stuff i said to move/remove) should be put into a class, sth like RerunLogger or sth, that way you also avoid the DEFAULT_ROOT_ENTITY_PATH global variable

semantic_digital_twin/src/semantic_digital_twin/world_description/geometry.py
```diff
@@ -401,31 +401,42 @@ def local_frame_bounding_box(self) -> BoundingBox:
     def to_json(self) -> Dict[str, Any]:
         # Serialize the raw (unscaled, unprocessed) mesh geometry and the scale separately
         base_mesh = trimesh.load_mesh(self.filename, process=False)
+        # Bake materials/textures down to per-vertex colors so the mesh's color
```
LucaKro, line 404: if this was a problem before, could you please also add a test that tests this behaviour, that fails in the old version and passes in the new version?

semantic_digital_twin/src/semantic_digital_twin/world_description/geometry.py
```diff
@@ -401,31 +401,42 @@ def local_frame_bounding_box(self) -> BoundingBox:
...
         )
         origin = from_json(data["origin"], **kwargs)
         scale = from_json(data["scale"], **kwargs)
         file_type = data["file_type"]
-        color = from_json(data["color"], **kwargs)
-        instance = cls.from_trimesh(
+        # Per-vertex colors only survive the temporary export/reload in a
+        # color-preserving format, so prefer PLY when colors are present.
+        if vertex_colors is not None:
+            file_type = "ply"
```
LucaKro: I remember somewhere in the back of my head that changing filetype like this was dangerous, but maybe its fine i guess if the tests pass.
Do the tests cover this case where the file type is changed?

semantic_digital_twin/src/semantic_digital_twin/world_description/geometry.py
```diff
@@ -470,8 +481,13 @@ def mesh(self) -> trimesh.Trimesh:
         """
         mesh = trimesh.load_mesh(self.filename)
         mesh.apply_scale(self.scale.to_np())
-        if not isinstance(mesh.visual, TextureVisuals):
-            mesh.visual.vertex_colors = trimesh.visual.color.to_rgba(self.color.to_rgba())
+        # Apply the shape's color only when it was explicitly set, so a mesh's own
+        # materials or per-vertex colors (e.g. from a .dae or from serialization)
+        # are preserved by default. dye_shapes still works as it sets a color.
```
LucaKro, line 487: dye_shapes?

semantic_digital_twin/test/rerun_test/test_rerun_adapter.py
```diff
@@ -0,0 +1,127 @@
...
+    HomogeneousTransformationMatrix,
+)
+from semantic_digital_twin.world import World
+from semantic_digital_twin.world_description.connections import FixedConnection
+from semantic_digital_twin.world_description.geometry import Box, Color, Mesh, Scale
+from semantic_digital_twin.world_description.shape_collection import ShapeCollection
+from semantic_digital_twin.world_description.world_entity import Body
+
+
+def _make_box_world() -> World:
```
LucaKro: please check if there are existing fixtures you can use instead, and if not make this a new fixture in the conftest

LucaKro (changes requested):
cool stuff, can you maybe provide a video, so that we can see this in action?

semantic_digital_twin/scripts/rerun_distributed_publisher.py
AbdelrhmanBassiouny: I actually like it, but I would change the name to load_world_with_ur5() or something similar.

@LucaKro, why don't you like this? This way, I do not need to know about all the already passed in arguments; I just want to load a world with a UR5, and this function makes it easy and simple.

semantic_digital_twin/scripts/rerun_distributed_publisher.py
AbdelrhmanBassiouny: Yeah I also agree with Luca here.

semantic_digital_twin/src/semantic_digital_twin/adapters/rerun/__init__.py
AbdelrhmanBassiouny: yeah I hate that llms do that all the time haha

semantic_digital_twin/src/semantic_digital_twin/adapters/rerun/rerun_logger.py
AbdelrhmanBassiouny, line 1: This file can be made into dataclasses instead of standalone functions, as these functions share most of the parameters already and are all targeted toward rerun logging, so make a `ReRunLogger` or something similar.

semantic_digital_twin/src/semantic_digital_twin/adapters/rerun/rerun_visualizer.py
AbdelrhmanBassiouny: haha same as my comment above

semantic_digital_twin/src/semantic_digital_twin/adapters/rerun/rerun_visualizer.py
AbdelrhmanBassiouny: or ReRunRecordingOutput

AbdelrhmanBassiouny (changes requested):

semantic_digital_twin/scripts/rerun_distributed_publisher.py
LucaKro: 1. it is only used in one instance
2. if everyone starts to define their own "example worlds" everywhere, then we will have duplicate code en masse, and everyone will do it slightly different.

I am not against this in general, but it needs to be done in a way to fits into the overall system.

For example in my robot PR, each robot knows about their ros path, to abstract exactly these kinds of parameters, because i agree, the user doesnt need to know those. 
But also if we have sth like this, then it should also be used in the fixtures, because otherwise we have methods creating "default worlds" in fixtures, and duplicated somewhere else

semantic_digital_twin/src/semantic_digital_twin/adapters/rerun/__init__.py
LucaKro: yup

semantic_digital_twin/scripts/rerun_visualizer_demo.py
AbdElRahmanFarhan: This runs without ROS, the other two scripts: rerun_distributed_publisher, and rerun_distributed_viewer are doing the same thing but viewing is decoupled. 

I actually was not sure if I should even commit these 3 scripts. I only used them to visually see the output from rerun. Also I am not sure if scripts folder is the right one for these 3 scripts. 

If you think we should remove them then I can do so before doing any changes

@AbdelrhmanBassiouny 
@LucaKro

LucaKro: i think they could also be removed, one useful thing they did is show how one uses it, but the testcase should fulfill the same purpose

semantic_digital_twin/src/semantic_digital_twin/adapters/rerun.py
```diff
@@ -0,0 +1,200 @@
...
+    Separate from the state-change callback so geometry is re-sent only on model
+    changes, not on every state update.
+    """
+
+    recording: rr.RecordingStream = field(kw_only=True)
+    """The recording stream geometry is logged to."""
+    root_entity_path: str = field(default="world", kw_only=True)
+    """Entity path under which the kinematic tree is logged."""
+
+    def _notify(self, **kwargs) -> None:
```
LucaKro: this method does not exist anymore, it should be "on_model_change" now

semantic_digital_twin/src/semantic_digital_twin/adapters/rerun.py
```diff
@@ -0,0 +1,200 @@
...
+            if self.target is None:
+                raise ValueError("RerunMode.SAVE requires a target file path.")
+            self.recording.save(self.target)
+        self.model_cb = RerunModelCallback(
+            _world=self._world,
+            recording=self.recording,
+            root_entity_path=self.root_entity_path,
+        )
+        self.model_cb.notify()
+        self._notify()
```
LucaKro: "on_state_change"

LucaKro (changes requested):
Nice!

semantic_digital_twin/src/semantic_digital_twin/adapters/rerun.py
AbdElRahmanFarhan: Yeah I think this was introduced in world synchronizer PR merged two days ago I think.

semantic_digital_twin/src/semantic_digital_twin/adapters/urdf.py
```diff
@@ -151,11 +151,28 @@ def from_file(
...
+        The xacro file is resolved and processed into a URDF string, applying the
+        given substitution arguments, before constructing the parser.
+
+        :param xacro_path: The path to the xacro file to expand.
+        :param prefix: The prefix for every name used in this world.
+        :param mappings: The xacro substitution arguments to apply during expansion
+            (the ``arg`` values, e.g. ``{"ur_type": "ur5"}``).
+        :return: A parser for the world described by the expanded xacro file.
+        """
         from xacro import process_file
```
AbdelrhmanBassiouny: No need for the inline import here, just put it in the beginning of the file.

semantic_digital_twin/src/semantic_digital_twin/world_description/geometry.py
```diff
@@ -401,31 +401,42 @@ def local_frame_bounding_box(self) -> BoundingBox:
     def to_json(self) -> Dict[str, Any]:
         # Serialize the raw (unscaled, unprocessed) mesh geometry and the scale separately
         base_mesh = trimesh.load_mesh(self.filename, process=False)
+        # Bake materials/textures down to per-vertex colors so the mesh's color
+        # survives serialization (e.g. across the ROS world synchronizer).
+        if hasattr(base_mesh.visual, "to_color"):
```
AbdelrhmanBassiouny: Can this field really be missing?

semantic_digital_twin/src/semantic_digital_twin/world_description/geometry.py
```diff
@@ -401,31 +401,42 @@ def local_frame_bounding_box(self) -> BoundingBox:
...
+            base_mesh.visual = base_mesh.visual.to_color()
+        mesh_dict = base_mesh.to_dict()
+        vertex_colors = getattr(base_mesh.visual, "vertex_colors", None)
```
AbdelrhmanBassiouny: Same here can vertex_colors be missing? Or it just can be None? If it is always there but sometimes None, then there's no need for getattr and the string attribute name. 

This is a known issue with llms that they are over defensive but overusing getattr and hasattr when we are sure the attribue exists always.

LucaKro (approved):

semantic_digital_twin/src/semantic_digital_twin/adapters/rerun.py
```diff
@@ -0,0 +1,200 @@
...
+
+    recording: rr.RecordingStream = field(init=False)
+    """The Rerun recording stream all data is logged to."""
+    model_cb: RerunModelCallback = field(init=False)
+    """The owned callback that logs and re-logs geometry on model changes."""
+
+    def __post_init__(self) -> None:
+        super().__post_init__()
+        self.recording = rr.RecordingStream(self.application_id)
+        if self.mode is RerunMode.SPAWN:
```
Tigul: This could be a match-case

Tigul (changes requested):

semantic_digital_twin/src/semantic_digital_twin/world_description/geometry.py
AbdElRahmanFarhan: When trimesh loads an untextured mesh (e.g. STL, or a PLY without texture), base_mesh.visual is a ColorVisuals instance, which has no to_color() method — that method only exists on TextureVisuals (UV + material, e.g. OBJ+MTL, GLTF/GLB). The hasattr guard handles exactly that case.

semantic_digital_twin/src/semantic_digital_twin/world_description/geometry.py
AbdElRahmanFarhan: Actually, it can be missing if the visual is TextureVisuals.

AbdElRahmanFarhan: It is better to handle this using isinstance I changed it

AbdelrhmanBassiouny (approved):

semantic_digital_twin/src/semantic_digital_twin/adapters/rerun.py
```diff
@@ -0,0 +1,203 @@
...
+from typing_extensions import Optional
+
+from semantic_digital_twin.callbacks.callback import (
+    ModelChangeCallback,
+    StateChangeCallback,
+)
+from semantic_digital_twin.world import World
+
+
+class RerunMode(Enum):
```
tomsch420: StrEnum

semantic_digital_twin/src/semantic_digital_twin/adapters/rerun.py
```diff
@@ -0,0 +1,203 @@
+# Rerun visualization adapter. Requires the rerun-sdk dependency; does not require ROS.
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from enum import Enum
+
+import rerun as rr
```
tomsch420: no abbrivations

tomsch420 (changes requested):
Use docformatter

tomsch420 (approved):

Tigul (approved):

semantic_digital_twin/src/semantic_digital_twin/adapters/rerun.py
```diff
@@ -0,0 +1,257 @@
...
+        parented under, so the two compose. With ``static=True`` the
+        transforms overwrite in place (no timeline history), giving a
+        constant-memory current-state view.
+
+        :param world: The world whose state is logged.
+        :param static: Whether to log without timeline history
+            (overwrite in place).
+        """
+        for body in world.bodies:
+            world_transform_body = world.compute_forward_kinematics_np(world.root, body)
```
ichumuh: there are faster ways of doing this. you can compute all fks in a batch to safe time, see https://github.com/cram2/cognitive_robot_abstract_machine/blob/6fcb273f1b5c609872fb99160bcc1850841defa8/semantic_digital_twin/src/semantic_digital_twin/collision_checking/collision_detector.py#L128

ichumuh (changes requested):

ichumuh (approved):

PR #344, opened 2026-06-05
Fixed flaky giskard test

MideDay:
Fixed flaky Giskard test, by waiting for the state to not change after executing for 0.1s
moved version updates for world to after callback notify


LucaKro:
hmm a few tests are failing, the only reason i can see for that in your PR is moving the versions. I think the problem is that the version affect the hash, so some of the method hashes may not be invalidated even though they need to be.

I think for now revert the version changes, and just keep the better assertion in the tests

AbdelrhmanBassiouny (approved):

tomsch420 (approved):

LucaKro (approved):

PR #345, opened 2026-06-08
Feature: WorldSynchronizer (Unified Model & State Synchronization)

LucaKro:
This PR merges the previously separate `StateSynchronizer` and `ModelSynchronizer` into a single `WorldSynchronizer` class that publishes both model and state updates on one shared ROS topic.

**Theoretical Motivation:** When model and state changes travel on separate ROS topics, DDS provides no inter-topic ordering guarantee. A state update referencing newly added DOF UUIDs can arrive before the model update that introduced them, causing a `KeyError` at runtime. Sending both message types on the same topic exploits DDS FIFO delivery semantics: a model update published before a state update is always received first, eliminating the race entirely.

**Practical Reasoning:** The old two-class design required users to manage two synchronizer instances, two topics, and coordinate their pausing/resuming manually. The unified `WorldSynchronizer` is a single object with explicit `synchronize_model` / `synchronize_state` flags that can also act as a receive-only participant on either axis.

---

- `StateSynchronizer` and `ModelSynchronizer` removed; replaced by `WorldSynchronizer` on a single `/semantic_digital_twin/world_sync` topic, eliminating the cross-topic ordering race.
- New `WorldUpdate` message type wraps an optional `ModificationBlock` and an optional `WorldStateUpdate` in one envelope, keeping the two change types in strict publication order.

---

### `WorldSynchronizer` (`adapters/ros/world_synchronizer.py`)

- Consolidates `StateSynchronizer` and `ModelSynchronizer` into one class. The intermediate `SynchronizerOnCallback` base class is removed entirely.
- Registers with both `model_change_callbacks` and `state_change_callbacks` and implements `on_model_change` / `on_state_change` separately, keeping the outgoing publication logic isolated.
- `synchronize_model: bool = True` and `synchronize_state: bool = True` fields allow a node to act as a receive-only participant for either axis without spawning separate objects.
- `_apply_state` now guards against unknown DOF identifiers and raises `StateUpdateContainsUnknownDegreesOfFreedomError` instead of silently producing a `KeyError` or writing garbage into the state vector.
- `apply_missed_messages` raises `ApplyMissedMessagesWhileWorldIsBeingModifiedError` if called inside an active `modify_world` context, preventing conflicting `publish_changes` policies from deadlocking or producing inconsistent state.
- `stop()` / `close()` now cleanly deregister from both callback lists and destroy all ROS publishers/subscribers (including the previously leaked `acknowledge_publisher`).
- `acknowledge_message` guards against `None` publisher so `Synchronizer` subclasses that do not create an acknowledge channel do not crash.

### `WorldUpdate` message (`adapters/ros/messages.py`)

- New `WorldUpdate` dataclass wraps an optional `ModificationBlock` and an optional `WorldStateUpdate`. Either field may be `None` when only one type of change is being communicated.
- Serialisation round-trip test added (`test_world_update_serialization_round_trip`).

### Callback infrastructure (`callbacks/callback.py`)

- `Callback.stop()` is no longer abstract; it now serves as a cooperative super-chain endpoint so subclasses can call `super().stop()` safely.
- The `notify` / `_notify` indirection removed from `Callback`; `StateChangeCallback` and `ModelChangeCallback` instead expose `notify_state_change` / `notify_model_change` which check `_is_paused` before delegating to the new abstract hooks `on_state_change` / `on_model_change`.
- Both `StateChangeCallback.stop()` and `ModelChangeCallback.stop()` now call `super().stop()` to correctly chain through the MRO.

### Tests (`test/semantic_digital_twin_test/test_ros/test_world_synchronizer.py`)

- Comprehensive new test suite for `WorldSynchronizer` covering: basic state and model synchronisation, cross-topic ordering guarantee (`test_world_synchronizer_ordering_no_key_error_after_model_change`), missed-message ordering (`test_world_synchronizer_missed_messages_applied_in_order`), `synchronize_model=False` / `synchronize_state=False` flags (both outgoing suppression and incoming reception), lifecycle idempotency (`test_stop_is_idempotent`), callback deregistration, error paths (`test_apply_state_with_unknown_identifier_raises`, `test_apply_missed_messages_inside_modify_world_raises`), and deadlock-prevention cases.

---

- **`StateSynchronizer`**: Removed. Replace with `WorldSynchronizer(synchronize_model=False)` for state-only participation, or a full `WorldSynchronizer` for both axes.
- **`ModelSynchronizer`**: Removed. Replace with `WorldSynchronizer(synchronize_state=False)` for model-only participation, or a full `WorldSynchronizer` for both axes.
- **`SynchronizerOnCallback`**: Removed intermediate base class; its responsibilities are absorbed into `WorldSynchronizer` and the refactored `Callback` hierarchy.

---

- **`StateSynchronizer` and `ModelSynchronizer` removed**: Any code constructing these classes must migrate to `WorldSynchronizer`. The `/semantic_digital_twin/world_state` and `/semantic_digital_twin/world_model` topics no longer exist; the unified topic is `/semantic_digital_twin/world_sync`. All participating nodes must be updated together.
- **`Callback._notify` / `Callback.notify` removed**: Subclasses that overrode `_notify` must instead implement `on_state_change` (for `StateChangeCallback`) or `on_model_change` (for `ModelChangeCallback`).
- **`Callback.stop()` no longer abstract**: Subclasses that previously relied on the abstract constraint to enforce implementation must add their own `stop()` bodies; failing to call `super().stop()` will silently skip the cooperative chain.

---

## Other Changes

- Minor: `acknowledge_message` parameter renamed from `msg` to `message` for consistency with the rest of the `WorldSynchronizer` API.
- Minor: Unused `Callback` import removed from `world_synchronizer.py`.

---


tomsch420 (approved):
hopefully peace from now on

semantic_digital_twin/src/semantic_digital_twin/adapters/ros/world_synchronizer.py
```diff
@@ -527,3 +353,204 @@ def _replace_world(self, new_world: World):
...
+            return
+
+        state_message = WorldStateUpdate(
+            meta_data=self.meta_data,
+            ids=list(changes.keys()),
+            states=list(changes.values()),
+        )
+        update = WorldUpdate(meta_data=self.meta_data, state_update=state_message)
+        self.publish(update, synchronous=self.synchronous)
+        self.update_previous_world_state()
```
ichumuh, line 442: so model and state updates are still separate, they just go over the same topic?

semantic_digital_twin/src/semantic_digital_twin/collision_checking/collision_detector.py
```diff
@@ -193,8 +193,8 @@ def __post_init__(self):
         self.world_state_updater = CollisionDetectorStateUpdater(
             collision_detector=self
         )
-        self.world_model_updater.notify()
-        self.world_state_updater.notify()
+        self.world_model_updater.on_model_change()
+        self.world_state_updater.on_state_change()
```
ichumuh, line 197: why are there still two class fields?

ichumuh (changes requested):
you likely have to change more in giskard.
there is ProcessWorldUpdate and we hacked the state update into PublishWorldState.
I also think we should test this pr on a real robot, maybe the stretch demo, before merging

semantic_digital_twin/src/semantic_digital_twin/collision_checking/collision_detector.py
LucaKro: this is collision detector stuff, model changes and state changes still exist as separate concepts.

semantic_digital_twin/src/semantic_digital_twin/adapters/ros/world_synchronizer.py
LucaKro: yes, but that guarantees that we always have the correct order of messages, so any state update will always only come after the corresponding model update, which was the goal of this PR

LucaKro:
> you likely have to change more in giskard. there is ProcessWorldUpdate and we hacked the state update into PublishWorldState. I also think we should test this pr on a real robot, maybe the stretch demo, before merging

@ichumuh yea true, we can try that. The giskard stuff im fixing rn there were tests failing. also how would you do the state update in PublishWorldState in a none hacky way? i swapped it out with my stuff, and thus far it still seems to work (i just replaced `state_synchronizer.world_callback` with `world_synchronizer.on_state_change`)

giskardpy/src/giskardpy/tree/behaviors/publish_joint_states.py
```diff
@@ -11,9 +11,11 @@ class PublishWorldState(GiskardBehavior):
...
     def update(self):
-        sync = GiskardBlackboard().giskard.state_synchronizer
+        sync = GiskardBlackboard().giskard.world_synchronizer
         for msg in sync.missed_messages:
             sync.apply_message(msg)
         sync.missed_messages.clear()
-        GiskardBlackboard().giskard.state_synchronizer.world_callback()
+        GiskardBlackboard().giskard.world_synchronizer.on_state_change(
+            publish_changes=True
+        )
```
ichumuh, line 16: we probably don't need this anymore because the world_updater handles it now

ichumuh (changes requested):

ichumuh (approved):

PR #356, opened 2026-06-10
Fixed loader tests for parallel

Tigul:
This should fix the last failing CI

### DIFF
--- a/semantic_digital_twin/src/semantic_digital_twin/adapters/sage_10k_dataset/loader.py
+++ b/semantic_digital_twin/src/semantic_digital_twin/adapters/sage_10k_dataset/loader.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import json
 import logging
 import os
+import random
 import shutil
 from dataclasses import dataclass, field
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


semantic_digital_twin/src/semantic_digital_twin/adapters/sage_10k_dataset/loader.py
```diff
@@ -65,7 +66,11 @@ def _download_scene_if_not_exists(self, scene_url: str) -> Path:
         # unzip the scene
         extraction_directory.mkdir(parents=True, exist_ok=True)
         with ZipFile(zipped_scene, "r") as zip_ref:
-            zip_ref.extractall(extraction_directory)
+            try:
+                zip_ref.extractall(extraction_directory)
+            except FileExistsError:
+                extraction_directory = self.directory / zipped_scene.stem / str(random.randint(0, 100000))
```
AbdelrhmanBassiouny: Why random here?

ichumuh (approved):

sunava (approved):

semantic_digital_twin/src/semantic_digital_twin/adapters/sage_10k_dataset/loader.py
tomsch420: yeah, why not os.getpid()?

Tigul: The problem is that two test try to extract the same scene at the same time and the one that is slower can't unpack it because the file is already extracted. So I just add a random number there to create a unique path

AbdelrhmanBassiouny: Can't you use a process ID from pytest or something?

tomsch420 (changes requested):

semantic_digital_twin/src/semantic_digital_twin/adapters/sage_10k_dataset/loader.py
Tigul: Found a better fix

tomsch420 (approved):

AbdelrhmanBassiouny (approved):

PR #361, opened 2026-06-11
[LMMTM] Fix stiffness and damping for mujoco>=3.7.0

HoangGiang93:
I fix the stiffness and damping for mujoco>=3.7.0 because of the breaking API changes (https://mujoco.readthedocs.io/en/stable/changelog.html#id4)


LucaKro:
hmm i dont like that :smiling_face_with_tear: we cannot make special cases everywhere to support multiple versions. choose one (preferably the newest one) and support that one.

semantic_digital_twin/src/semantic_digital_twin/adapters/multi_sim.py
```diff
@@ -922,7 +922,7 @@ class MujocoTendon(SimulatorAdditionalProperty):
     Inertia associated with changes in tendon length.
     """
 
-    damping: float = 0.0
+    damping: Union[float, List[float]] = 0.0
```
tomsch420: These type annotations are illegal. Why not always list float?

tomsch420 (changes requested):

semantic_digital_twin/src/semantic_digital_twin/adapters/multi_sim.py
HoangGiang93: ok, then I will make a list of floats, either with one element for the old versions or three elements with the new versions

tomsch420: good idea

LucaKro (changes requested):
Please stick to only supporting one API version, not multiple

semantic_digital_twin/src/semantic_digital_twin/adapters/multi_sim.py
tomsch420: but also remove support for old mujoco

HoangGiang93: the old one actually runs most stable now, so for the sake of compatibility I will support as much as I can.

HoangGiang93: I updated it, please check again.

HoangGiang93:
> Please stick to only supporting one API version, not multiple

it's hard to see whether there are still other errors because the version 3.5.0 is the most stable version that I'm using until now. MuJoCo will keep updating a lot so for the sake of compatibility and stability I will keep this fix for now. At the end it doesn't produce too many swapping codes.

tomsch420 (approved):

LucaKro (approved):

PR #362, opened 2026-06-11
Improving Exceptions

LucaKro:
Improved Exceptions in CRAM, improved some error messages in semdt and added automatic suggestion of similar body names on lookup


ichumuh (approved):

AbdelrhmanBassiouny (changes requested):
Maybe add a suggest_correction abstract method to separate error reporting from fix suggestion.

.github/workflows/build_doc_reusable.yml
```diff
@@ -25,7 +25,7 @@ jobs:
   build:
     runs-on: ubuntu-latest
     container:
-      image: ${{ inputs.use_container && 'coraplex/semantic_world:jazzy' || '' }}
+      image: ${{ inputs.use_container && 'pycram/semantic_world:jazzy' || '' }}
```
AbdelrhmanBassiouny: is this intended?

semantic_digital_twin/src/semantic_digital_twin/adapters/multi_sim.py
```diff
@@ -142,6 +141,32 @@ class MultiSimError(DataclassException):
     """Base class for all MultiSim-related exceptions."""
 
 
+@dataclass
```
AbdelrhmanBassiouny: Move these to a separate exceptions.py instead of them being here.

AbdelrhmanBassiouny (changes requested):
Noice

ichumuh (approved):

.github/workflows/build_doc_reusable.yml
ichumuh: i think so, otherwise the CI wouldn't run

sunava: gotta use this till we have it on dockerhub as dicussed in the zulip

semantic_digital_twin/src/semantic_digital_twin/adapters/multi_sim.py
sunava: +1

sunava (changes requested):

sunava (approved):

AbdelrhmanBassiouny (approved):

PR #370, opened 2026-06-17
[Totally no sync issues] Fix deadlocks and races in WorldSynchronizer                                                                                                                              

LucaKro:
## Summary
The world synchronizer could deadlock or stall when two synchronized worlds published concurrently, and had several thread-safety gaps around publishing, message application, and callback handling. This PR fixes the real cross-process issues and adds regression tests for each.

## Changes

**Publishing no longer blocks under the world lock (deadlock fix)**
Publishing used to happen inside `modify_world().__exit__()` while `_world_lock` was held. In synchronous mode this blocked the receiving executor, which needs the same lock to apply/ack the inbound message causing a cross-process stall. Publications are now deferred to the world model manager's `pending_publications` and flushed *after* the lock is released. Standalone state changes (not inside a modification) publish immediately as before.

**`publish()` is concurrency-safe**
Concurrent synchronous publishes shared ack-tracking state (`_current_publication_event_id`, etc.), so one could clobber another's event id and stall it until timeout. The synchronous path is now serialized with a `_publish_lock`, with the event id set under the lock. The async path stays lock-free.

**Inbound deserialization runs under the world lock**
`subscription_callback` built the entity tracker and ran `from_json` (reading world structure) before locking, racing with concurrent modifications. It now holds `_world_lock` across deserialization and apply.

**Atomic message application**
`apply_message` applied the model block and state update under separate lock acquisitions; it now applies both under a single (reentrant) hold so a combined update is never observed half-applied. `apply_missed_messages` similarly holds the lock across the whole batch.

**Callback iteration over a copy**
Model/state change notifications iterated the live callback lists, so a callback removing itself/another (or a concurrent `stop()`/`close()`) could skip a callback. Notifications now iterate over a copy.

## Tests
Added targeted regression tests for each fix above (deadlock sentinels, lock-held probes, and instrumented interleavings). Two scenarios that are only problematic when two worlds/synchronizers share a single process, `_snapshot_subscribers` undercounting an in-process peer, and the ABBA nested-`modify_world` deadlock (this is the one @tomsch420 tagged me on) are `@pytest.mark.skip` or `@pytest.mark.xfail`ed with reasons explaining they cannot occur in a real per-process deployment.

## Notes
- Touches `world.py` (`modify_world` exit ordering) and `world_state.py` (callback iteration); full synchronizer module passes (91 passed, 2 skipped) and core world/callbacks tests pass.

### DIFF
--- a/semantic_digital_twin/src/semantic_digital_twin/adapters/ros/world_synchronizer.py
+++ b/semantic_digital_twin/src/semantic_digital_twin/adapters/ros/world_synchronizer.py
@@ -170,24 +170,24 @@ class Synchronizer(WorldEntityWithClassBasedID):
             process_id=os.getpid(),
         )
 
-    def subscription_callback(self, msg: std_msgs.msg.String):
+    def subscription_callback(self, message: std_msgs.msg.String):
         """
         Wrap the origin subscription callback by self-skipping and disabling the next world callback.
         Holds the world lock while deserializing to ensure no changes happen while building the tracker and
         running from_json.
 
-        :param msg: The incoming ROS string message containing a serialized synchronization message.
+        :param message: The incoming ROS string message containing a serialized synchronization message.
         """
         with self._world._world_lock:
             tracker = WorldEntityWithIDKwargsTracker.from_world(self._world)
-            deserialized_msg = from_json(
-                json.loads(msg.data), **tracker.create_kwargs()
+            deserialized_message = from_json(
+                json.loads(message.data), **tracker.create_kwargs()
             )
 
-            if deserialized_msg.meta_data == self.meta_data:
+            if deserialized_message.meta_data == self.meta_data:
                 return
 
-            self._subscription_callback(deserialized_msg)
+            self._subscription_callback(deserialized_message)
 
     def acknowledge_message(self, message: message_type):
         if self.acknowledge_publisher is None:
@@ -330,6 +330,8 @@ class ModelReloadSynchronizer(Synchronizer):
         Save the current world model to the database and publish the primary key to the ROS topic such that other
         processes can subscribe to the model changes and update their worlds.
         """
+        from semantic_digital_twin.orm.ormatic_interface import WorldMappingDAO  # type: ignore
+
         dao = to_dao(self._world)
         self.session.add(dao)
         self.session.commit()
--- a/semantic_digital_twin/src/semantic_digital_twin/adapters/ros/world_synchronizer.py
+++ b/semantic_digital_twin/src/semantic_digital_twin/adapters/ros/world_synchronizer.py
@@ -1,7 +1,7 @@
 import json
 import os
 import threading
-from abc import ABC, abstractmethod
+from abc import abstractmethod
 from dataclasses import dataclass, field
 from functools import cached_property
 from typing import ClassVar, Optional, Set, Type, List, Dict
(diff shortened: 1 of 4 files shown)


semantic_digital_twin/src/semantic_digital_twin/world_description/world_state.py
```diff
@@ -123,7 +123,7 @@ def _notify_state_change(self, **kwargs) -> None:
         the state version.
         """
         self.version += 1
-        for callback in self.state_change_callbacks:
+        for callback in list(self.state_change_callbacks):
```
tomsch420, line 126: Why?

semantic_digital_twin/src/semantic_digital_twin/adapters/ros/world_synchronizer.py
```diff
@@ -322,8 +330,6 @@ def publish_reload_model(self):
         Save the current world model to the database and publish the primary key to the ROS topic such that other
         processes can subscribe to the model changes and update their worlds.
         """
-        from semantic_digital_twin.orm.ormatic_interface import WorldMappingDAO
```
tomsch420, line 325: Doesn't this lead to dao_not_found errors?

tomsch420 (changes requested):
So synchronizer fixed Fr fr RR?

semantic_digital_twin/src/semantic_digital_twin/world_description/world_state.py
LucaKro: copying the list so that, if a callback is removed while they are looping, no callback is skipped

semantic_digital_twin/src/semantic_digital_twin/adapters/ros/world_synchronizer.py
LucaKro: so far it doesnt, unless its not tested

semantic_digital_twin/src/semantic_digital_twin/adapters/ros/world_synchronizer.py
```diff
@@ -166,17 +173,21 @@ def meta_data(self) -> MetaData:
...
-        msg = from_json(json.loads(msg.data), **tracker.create_kwargs())
+        with self._world._world_lock:
+            tracker = WorldEntityWithIDKwargsTracker.from_world(self._world)
+            deserialized_msg = from_json(
+                json.loads(msg.data), **tracker.create_kwargs()
+            )
 
-        if msg.meta_data == self.meta_data:
-            return
+            if deserialized_msg.meta_data == self.meta_data:
```
ichumuh: Mildly Suspicious Goose?

Tigul (approved):

ichumuh (changes requested):

semantic_digital_twin/src/semantic_digital_twin/adapters/ros/world_synchronizer.py
tomsch420: test it grrrrrr

LucaKro: readded it as discussed

ichumuh (approved):

tomsch420 (approved):

PR #325, opened 2026-05-21
Stretch Demo

J-Schaefer:
This PR adds the new Stretch demo that we implemented for a recent event. Included are also many smaller fixes in other packages.

PR #373, opened 2026-06-17
Fixes from stretch demo

Tigul:
Destilled fixes for CRAM that were made during the making of the stretch demo, taken from #325


tomsch420 (approved):

LucaKro (changes requested):
So i just got 1 third of my weekly tokens reset just before they will fully in 7 hours, so I thought might as well try out the new agent skill. and i must say it worked really well. I let it review this PR and it found a bunch of issue, dead imports, forgotten typing, docstrings, missing tests etc, but also actual bugs. Quote:
`StretchMoveReal._motion_chart` is a plain method, not a `@property`, violating the abstract property contract in `BaseMotion`; every real-execution navigation call silently returns a bound-method object instead of the task. Add `@property` before `def _motion_chart`.

which I confirmed to be correct.

> Overall this review with sonnet took ~25% of my 6h session limit, and roughly 2% of the weekly limit. the review took about 20 minutes, and when it prompted me i told it to use multiple agents for the review.
At the time of writing this, I have not had it fix the issues yet.

this is a summary of the review:

# Review Summary — fixes_from_stretch_demo (2026-06-18)

Branch reviewed against `cram2/main` at `90677a7a`. 

---

## coraplex

The most critical bug is in `stretch_motion_mapping.py`: `StretchMoveReal._motion_chart` is a plain method where an abstract `@property` is expected, so every real navigation call silently returns a bound-method object instead of executing the motion task. Two test decorators use `@pytest.mark.skipIf` (capital I), which pytest does not recognise, causing those tests to run and fail with `NameError` rather than being skipped. A third test imports `MoveMotion` and `NavigateActionServerTask` from behind a guarded star-import but has no skip guard of its own, so it fails when the import is unavailable. In the multi-robot location designator test, `context.debug_mode = True` sets a non-existent attribute — the correct name is `context.debug` — meaning the debug code-path is never actually exercised.

On the optional side: the `StretchMoveReal` docstring still refers to HSRB; `NavigateActionServerTask` is imported at module level but only appears in a commented-out block and its return-type annotation is wrong; several files are missing `from __future__ import annotations` and use `typing` instead of `typing_extensions`; there are a couple of docstring typos (`:praam context:`, `:parm test_world:`); a `try/except TimeoutError` block violates the project's no-try-except rule; `correct_arm` can remain `None` after a loop and then be passed as an invalid argument; and `StretchMoveReal` is unnecessarily nested inside `StretchMoveSim`.

---

## giskardpy

`configs.py` has a duplicated `from dataclasses import dataclass, field` import at the top of the file. `stretch_standalone.py` imports `WorldWithDiffDriveRobot` which is no longer used after a refactor. In `ros_tasks.py`, `NavigateActionServerTask.build` calls `self._action_client.wait_for_server()` a second time after the parent `super().build(context)` already does it; the class also re-declares the `action_topic` field it already inherits from `ActionServerTask`, causing surprising dataclass behaviour. More seriously, class-level references to `NavigateToPose.Goal`, `.Result`, and `.Feedback` are evaluated at import time, but `NavigateToPose` may be `None` when `nav2_msgs` is absent, which would raise `AttributeError` at import. When a navigation goal is rejected, `goal_response_callback` does a bare `return`, leaving `_result` as `None` and causing `on_tick` to stay `UNKNOWN` indefinitely — failure must be signalled explicitly. A stray `1` character at the end of the `NavigateActionServerTask` docstring should be removed. In `cartesian_goals.py` the Python-2 `from __future__ import division` shim must be replaced with `from __future__ import annotations`.

Optional items: the `threshold` and `weight` fields in `cartesian_goals.py` should use `field(default=..., kw_only=True)` to match sibling fields; inline index comments (`# 0`, `# 1`, …) in `configs.py` restate list positions without semantic value; no tests cover `DifferentialDriveBaseGoal.threshold`, the two-stage nav action result handling, or the corrected joint list.

---

## semantic_digital_twin

The most severe issue is in `world_modification.py`: `update_world_references_in_updated_kwargs` references `self.updated_kwargs`, which does not exist — the correct field is `updated_kwargs_json_list` — causing a runtime `AttributeError` on every call. The same method also lacks a return-type annotation and a docstring. The file has a duplicated `TYPE_CHECKING` guard block with a redundant `World` import, and uses `Dict`, `Any`, `Self`, `Optional` from `typing` instead of `typing_extensions`. `WorldEntityWithIDNotFoundError` is imported but never used and should be removed. None of the three changed files include tests for the new `Stretch._setup_velocity_limits`, `World.is_entity_in_world_by_name`, or `AttributeUpdateModification.update_world_references_in_updated_kwargs`.

Optional items: `is_entity_in_world_by_name` builds a full list for a membership test where a short-circuit `any(...)` would suffice; it is not decorated with `@memoize` unlike comparable lookup helpers; the name is ambiguous given that the world also holds annotations and connections; bare float literals in `stretch.py` (`0.01`, `0.4`, `0.5`, `0.1`) have no explanation; and `stretch.py` imports `Union` and `Self` from `typing` instead of `typing_extensions`.

LucaKro:
Okay i let it fix the issues and made a PR to jonas branch.

It took 40 minutes to fix the PR, so about 1h overall for the whole thing. It used 50% of my 6h capacity, and 5% of my weekly limit with sonnet. but i must say, it found a lot of legitimate things, even if some of them were minor, like missing imports etc

LucaKro (approved):

giskardpy/src/giskardpy/motion_statechart/ros2_nodes/ros_tasks.py
```diff
@@ -99,13 +99,11 @@ def on_start(self, context: MotionStatechartContext):
         future.add_done_callback(self.result_callback)
 
     def result_callback(self, future):
-        self._result = future.result().result
-        logger.info(
-            f"Action server {self.action_topic} returned result: {self._result}"
-        )
+        self._result = future.result()
+        logger.info(f"Action server {self.action_topic} done.")
```
ichumuh, line 120: this is wrong, see below

giskardpy/src/giskardpy/motion_statechart/ros2_nodes/ros_tasks.py
```diff
@@ -178,11 +171,45 @@ def build(self, context: MotionStatechartContext) -> NodeArtifacts:
             position_error < 0.01, sm.abs(rotation_error) < 0.01
         )
 
-        logger.info(f"Waiting for action server {self.action_topic}")
-        self._action_client.wait_for_server()
-
         return artifacts
 
+    def on_start(self, context: MotionStatechartContext):
```
ichumuh: `on_start`, `goal_response_callback` and `result_callback` must be moved to the parent class, the navigation specific stuff needs to be extracted beforehand of course

ichumuh (changes requested):

giskardpy/src/giskardpy/motion_statechart/ros2_nodes/ros_tasks.py
ichumuh: you are writing the wrong type into self._result. doesn't match the super implementation

giskardpy/src/giskardpy/tree/behaviors/joint_group_vel_controller_publisher.py
```diff
@@ -34,8 +38,19 @@ def __init__(self, cmd_topic: str, connections: List[ActiveConnection1DOF]):
     @record_time
     def update(self):
         msg = Float64MultiArray()
+        minimum_valid_velocity = 0.03
```
ichumuh: this should be a parameter, not something that's applied to all robots

ichumuh (changes requested):

giskardpy/src/giskardpy/motion_statechart/ros2_nodes/ros_tasks.py
Tigul: Should it only be result_response?

giskardpy/src/giskardpy/motion_statechart/ros2_nodes/ros_tasks.py
ichumuh: this method should still be moved to the parent class

giskardpy/src/giskardpy/motion_statechart/ros2_nodes/ros_tasks.py
```diff
@@ -178,16 +188,30 @@ def build(self, context: MotionStatechartContext) -> NodeArtifacts:
...
+        """
+        future = self._action_client.send_goal_async(self._msg)
+        future.add_done_callback(self.goal_response_callback)
+
+    def result_callback(self, future):
+        """
+        Stores the navigation result returned by the action server.
+        """
+        result_response = future.result()
+        self._result = result_response
```
ichumuh: delete this method, should be same as parent class now. unless you really want the print, then call super

ichumuh (changes requested):

giskardpy/src/giskardpy/motion_statechart/ros2_nodes/ros_tasks.py
Tigul: I would like to keep the log

ichumuh (approved):

PR #392, opened 2026-06-25
Move semantic-digital-twin and coraplex publisher to correct path and fix unify version numbers

J-Schaefer:
Workflow files have to be in the .github/workflows dir, not in a subdir.


Tigul:
They both trigger on push, please change that such that they only trigger when a new version tag is  pused

J-Schaefer:
> They both trigger on push, please change that such that they only trigger when a new version tag is pused

@Tigul No problem, should it trigger on all release tags or contain some keyword or something?

Tigul:
I think all release tags are fine

tomsch420:
dont use the tags, do it exactly as krrood, pm and re. This is easier to use and less easy to f up. Just ask claude to do it in your workflows like in the other workflows

tomsch420 (approved):

J-Schaefer:
Checks failing right now with a parsing error of the pyproject.toml files
```bash
error: Failed to parse: `physics_simulators/pyproject.toml`
  Caused by: TOML parse error at line 32, column 11
   |
32 | license = "MIT"
   |           ^^^^^
invalid type: string "MIT", expected a sequence
Error: Process completed with exit code 2.
```

This this syntax should by correct according to documentation. Could some tool be outdated? Will check tomorrow.

cognitive_robot_abstract_machine/__init__.py
```diff
@@ -0,0 +1,4 @@
+try:
+    from ._version import __version__
+except ImportError:
```
AbdelrhmanBassiouny: Why handle the exception, shouldn't the version file be always importable?

semantic_digital_twin/src/semantic_digital_twin/__init__.py
```diff
@@ -1,11 +1,20 @@
...
 
+try:
+    from ._version import __version__
+except ImportError:
+    __version__ = "0.0.0"
+
 logger = logging.getLogger("semantic_digital_twin")
 logger.setLevel(logging.INFO)
 
-from krrood.symbol_graph.symbol_graph import SymbolGraph
```
AbdelrhmanBassiouny, line 9: Did you encounter issues because of this?

VERSION
```diff
@@ -0,0 +1 @@
+26.07.0
```
AbdelrhmanBassiouny, line 1: So we set this one ourselves right?

AbdelrhmanBassiouny (changes requested):

cognitive_robot_abstract_machine/__init__.py
J-Schaefer: Not necessarily. The main file we set the version number in is `VERSION` in the root dir. The _version.py are generated using the script `scripts/sync_version.py`. This is because the CI in individual packages is isolated and cannot access the VERSION file in a parent directory.

semantic_digital_twin/src/semantic_digital_twin/__init__.py
J-Schaefer: I think I had an import error but this doesn't really make sense to keep. Will delete

VERSION
J-Schaefer: Yes correct. Only here. Other files (_version.py in the packages) are generated.

cognitive_robot_abstract_machine/__init__.py
AbdelrhmanBassiouny: But once we commit these files, they stay permanently here, so there's no fear of them not being there anymore, right?

J-Schaefer: Removed this.

AbdelrhmanBassiouny (approved):

LucaKro (approved):

ichumuh (approved):

sunava (approved):

PR #396, opened 2026-06-26
feat: add prototype confidence-aware OOD evaluation engine

ganeshyadavyatham:
Adds a confidence-aware evaluation engine intended for KRROOD's rule evaluation. The rule engine currently always returns a confident answer, even for objects it has never seen. This change checks whether an object is familiar (by scoring it against a probability model learned from example objects) and raises a warning if it isn't — for example, a 50 kg cup or an object with a missing material tag gets flagged, while normal objects pass.
first prototype:- The engine runs standalone for now; it is not yet wired into KRROOD's EQL. It uses synthetic data and probabilistic-model 6.3.0 (the newer version needs pyjpt, which doesn't build on aarch64). Next steps: wiring it into EQL evaluation, using real stored data, and replacing the GMM with JPT.


experiments/src/experiments/confidence_aware_eql/engine/__init__.py
```diff
@@ -0,0 +1,28 @@
...
+data-driven threshold, and flags anomalous / incomplete objects with traceable
+warnings. The circuit supports log_likelihood / marginal / conditional (the
+same kind of object a JPT produces), learned without the `jpt` package.
+
+To add a DOMAIN: drop a file in domains/  (no edits here).
+To add TESTS:    drop a file in tests/    (no edits here).
+Pick a domain at runtime:  python run.py
+"""
+
+from .domain import Domain, Feature
```
AbdelrhmanBassiouny: Do not put any code in the __init__.py

experiments/src/experiments/confidence_aware_eql/engine/circuit_model.py
```diff
@@ -0,0 +1,97 @@
...
+from sklearn.mixture import GaussianMixture
+
+from random_events.variable import Continuous
+from probabilistic_model.distributions.gaussian import GaussianDistribution
+from probabilistic_model.probabilistic_circuit.rx.probabilistic_circuit import (
+    ProbabilisticCircuit, SumUnit, ProductUnit, leaf,
+)
+
+
+class CircuitModel:
```
AbdelrhmanBassiouny: use dataclass, and add docs to the class, and all the methods and fields.

experiments/src/experiments/confidence_aware_eql/engine/domain.py
```diff
@@ -0,0 +1,77 @@
...
+SymbolicDistribution leaf per categorical without changing the public API.
+"""
+
+from dataclasses import dataclass
+from typing_extensions import List, Dict, Optional, Tuple
+import numpy as np
+
+
+@dataclass
+class Feature:
```
AbdelrhmanBassiouny: Add docs for all classes and their fields in this file.

experiments/src/experiments/confidence_aware_eql/engine/evaluator.py
AbdelrhmanBassiouny, line 1: docs

experiments/src/experiments/confidence_aware_eql/engine/pipeline.py
AbdelrhmanBassiouny, line 1: docs are missing parameters documentation, and do not use appreviated names, always full names for variables, classes, methods, and parameters.

experiments/src/experiments/confidence_aware_eql/engine/pipeline.py
```diff
@@ -0,0 +1,58 @@
...
+
+def build_evaluator(domain: Domain, spec: dict, *, percentile: float = 1.0,
+                    n_per_class: int = 80, seed: int = 0, n_components="auto"):
+    """Run steps 2–4 for a domain and return (evaluator, model, strategy, data)."""
+    data = generate_dataset(domain, spec, n_per_class=n_per_class, seed=seed)
+    model = CircuitModel.fit(data, domain.names, n_components=n_components, seed=seed)
+    strategy = PercentileThreshold(percentile=percentile)
+    strategy.fit(model.log_likelihood(data))
+    evaluator = ConfidenceAwareEvaluator(domain, model, strategy.threshold)
+    return evaluator, model, strategy, data
```
AbdelrhmanBassiouny: Do not return a big tuple like that; make a dataclass that wraps these as fields, and return just an instance of that class. Then add the return type hint to be that new class.

experiments/src/experiments/confidence_aware_eql/engine/pipeline.py
```diff
@@ -0,0 +1,58 @@
...
+    corrupt = _first_continuous(domain)
+    familiar = sample_objects(domain, spec, n=n, seed=seed)
+    rng = np.random.default_rng(seed + 100)
+    anomalies = []
+    for o in sample_objects(domain, spec, n=n, seed=seed + 1):
+        bad = dict(o)
+        bad[corrupt] = float(rng.uniform(*anomaly_range))
+        anomalies.append(bad)
+
+    fp = sum(0 if evaluator.is_familiar(o) else 1 for o in familiar) / len(familiar)
```
AbdelrhmanBassiouny: no abbreviations, and make a dataclass that encompasses the outputs

experiments/src/experiments/confidence_aware_eql/domains/__init__.py
tomsch420, line 1: do not put anything in __init__ files

experiments/src/experiments/confidence_aware_eql/domains/_template.py
tomsch420, line 1: remove this kind of files.  They look like test cases

experiments/src/experiments/confidence_aware_eql/domains/kitchen.py
```diff
@@ -0,0 +1,36 @@
+"""Kitchen domain."""
+
+from experiments.confidence_aware_eql.engine import Domain, Feature
+
+DOMAIN = Domain("kitchen", [
+    Feature("weight", "continuous"),
```
tomsch420: also definitions like this should be done using classes, not the custom approach

experiments/src/experiments/confidence_aware_eql/engine/__init__.py
```diff
@@ -0,0 +1,28 @@
...
+from .circuit_model import CircuitModel
+from .threshold import PercentileThreshold, StdDevThreshold
+from .warning import UnfamiliarSampleWarning
+from .evaluator import ConfidenceAwareEvaluator
+from .datasets import generate_dataset, sample_objects
+from .pipeline import build_evaluator, evaluate_detection
+
+__all__ = [
```
tomsch420: never use __all__. Also empty __init__ files. Also no relative imports

experiments/src/experiments/confidence_aware_eql/engine/circuit_model.py
tomsch420: use dataclasses. Dont write custom __init__ constructors. If this is suppossed to be a GMM, why not create a real GMM somewhere? Also where doc?

experiments/src/experiments/confidence_aware_eql/engine/datasets.py
```diff
@@ -0,0 +1,62 @@
...
+into a matrix in domain order and pass it to CircuitModel.fit.)
+"""
+
+from typing_extensions import Dict
+import numpy as np
+
+from .domain import Domain
+
+
+def generate_dataset(domain: Domain, spec: Dict[str, Dict],
```
tomsch420: these look like testcase fixtures

experiments/src/experiments/confidence_aware_eql/engine/domain.py
tomsch420: looks duplicated with what is in the RSPN classes

experiments/src/experiments/confidence_aware_eql/engine/evaluator.py
```diff
@@ -0,0 +1,63 @@
...
+            )
+            self.warnings.append(w)
+            return log_p, w
+
+        return log_p, None
+
+    def is_familiar(self, obj: Dict) -> bool:
+        log_p, w = self.check(obj, node_name="_probe_")
+                                          
+        if self.warnings and self.warnings[-1].node_name == "_probe_":
```
tomsch420: thats weird

experiments/src/experiments/confidence_aware_eql/engine/evaluator.py
```diff
@@ -0,0 +1,63 @@
...
+from typing_extensions import Dict, List, Optional, Tuple
+import numpy as np
+
+from .domain import Domain
+from .circuit_model import CircuitModel
+from .warning import UnfamiliarSampleWarning
+
+
+@dataclass
+class ConfidenceAwareEvaluator:
```
tomsch420, line 26: hmmm i think this needs to be integrated in the backend of EQL directly. @AbdelrhmanBassiouny why do you think?

experiments/src/experiments/confidence_aware_eql/engine/pipeline.py
```diff
@@ -0,0 +1,58 @@
...
+    """Run steps 2–4 for a domain and return (evaluator, model, strategy, data)."""
+    data = generate_dataset(domain, spec, n_per_class=n_per_class, seed=seed)
+    model = CircuitModel.fit(data, domain.names, n_components=n_components, seed=seed)
+    strategy = PercentileThreshold(percentile=percentile)
+    strategy.fit(model.log_likelihood(data))
+    evaluator = ConfidenceAwareEvaluator(domain, model, strategy.threshold)
+    return evaluator, model, strategy, data
+
+
+def _first_continuous(domain: Domain) -> str:
```
tomsch420: also weird that this file exists

experiments/src/experiments/confidence_aware_eql/engine/threshold.py
tomsch420, line 1: i like this idea. Perhaps we can estimate such calibrations from the training data directly

experiments/src/experiments/confidence_aware_eql/engine/threshold.py
```diff
@@ -0,0 +1,50 @@
...
+        self.threshold = float(np.percentile(lls, self.percentile))
+        return self.threshold
+
+    def describe(self) -> str:
+        return (f"PercentileThreshold(p={self.percentile}) "
+                f"-> {self.threshold:.2f} (~{self.percentile:.1f}% expected false positives)")
+
+
+@dataclass
+class StdDevThreshold:
```
tomsch420: no abbrivations

experiments/src/experiments/confidence_aware_eql/engine/warning.py
```diff
@@ -0,0 +1,14 @@
+"""The traceable out-of-distribution warning."""
+
+from dataclasses import dataclass
+from typing_extensions import Optional
+
+
+@dataclass
+class UnfamiliarSampleWarning:
+    node_name: str
+    log_p: Optional[float]
```
tomsch420: no abbrivations. Also i think this should be some form of exception

experiments/src/experiments/confidence_aware_eql/README.md
tomsch420, line 1: dont write readmes. Use the sphinx documentation engine to create documentation

experiments/src/experiments/confidence_aware_eql/registry.py
```diff
@@ -0,0 +1,17 @@
+import importlib
+import pkgutil
+
+import domains
+
+REQUIRED_ATTRS = ("DOMAIN", "SPEC")
+
+
```
tomsch420: whats the purpose of this?

experiments/src/experiments/confidence_aware_eql/tests/_assertions.py
tomsch420, line 1: also looks like its not needed if you properly use pytest

experiments/src/experiments/confidence_aware_eql/tests/registry.py
```diff
@@ -0,0 +1,26 @@
...
+start with "_" (e.g. _template.py) are skipped. This is what lets you add a
+domain by simply creating a file.
+"""
+
+import importlib
+import pkgutil
+
+import domains
+
+REQUIRED_ATTRS = ("DOMAIN", "SPEC")
```
tomsch420: use the general tests structure. Just tell your claude to look at how the project is structured and how the other code looks

tomsch420 (changes requested):
I think the direction is good. There are many things you need to clean up such that we can properly use that. Keep the effort up and we'll get there soon!

experiments/src/experiments/confidence_aware_eql/engine/threshold.py
```diff
@@ -0,0 +1,50 @@
...
+    A Gaussian-tail style cutoff; flags objects k standard deviations less
+    likely than the average familiar object.
+"""
+
+from dataclasses import dataclass
+import numpy as np
+
+
+@dataclass
+class PercentileThreshold:
```
AbdelrhmanBassiouny: docs

experiments/src/experiments/confidence_aware_eql/engine/threshold.py
AbdelrhmanBassiouny: docs

experiments/src/experiments/confidence_aware_eql/engine/warning.py
```diff
@@ -0,0 +1,14 @@
+"""The traceable out-of-distribution warning."""
+
+from dataclasses import dataclass
+from typing_extensions import Optional
+
+
+@dataclass
+class UnfamiliarSampleWarning:
```
AbdelrhmanBassiouny: docs

experiments/src/experiments/confidence_aware_eql/eql_integration.py
```diff
@@ -0,0 +1,77 @@
...
+                                                                                 
+                                                                               
+ObjectExtractor = Callable[[Any, Any], Optional[Dict]]
+
+
+@dataclass
+class ConfidenceObserver(EvaluationObserver):
+    """Runs an OOD check on each evaluated node and records warnings."""
+
+    evaluator: ConfidenceAwareEvaluator
```
AbdelrhmanBassiouny, line 29: docs

experiments/src/experiments/confidence_aware_eql/kitchen_eql_demo.py
```diff
@@ -0,0 +1,117 @@
...
+        with refinement(obj.weight > 2.0, obj.material == "glass"):
+            Add(grasp, inference(GraspChoice)(obj=obj, style="two_handed"))
+
+                                                                    
+    evaluator, model, strategy, data = build_evaluator(kitchen.DOMAIN, kitchen.SPEC, seed=0)
+
+                                                             
+    results, warnings = run_with_confidence(query, evaluator, object_extractor)
+
+    print("Grasp decisions:")
```
AbdelrhmanBassiouny: can you make tests for this demo that assert the expected behaviour? this would also help us understant how this is usefult and how it behaves with eql

AbdelrhmanBassiouny (changes requested):
Looks nice, maybe move it outside experimets and put in krrood, and put the tests in krrood tests, but first answer the current review feedback, and the move can be maybe another PR.

experiments/src/experiments/confidence_aware_eql/domains/__init__.py
ganeshyadavyatham: Done, domains/__init__.py is empty now.

experiments/src/experiments/confidence_aware_eql/domains/_template.py
ganeshyadavyatham: Deleted. With domains defined as classes there's nothing to template.

experiments/src/experiments/confidence_aware_eql/domains/kitchen.py
ganeshyadavyatham: Reworked. KitchenObject is a Symbol dataclass whose fields are the features, following the Body/Handle pattern in dataset/semantic_world_like_classes.py. Material is an enum. The DOMAIN = Domain(...) construction and the SPEC dictionary are gone — FeatureSchema.from_dataclass() derives the feature schema from the class fields.

experiments/src/experiments/confidence_aware_eql/engine/__init__.py
ganeshyadavyatham: All three done: __all__ removed everywhere, every __init__.py is empty, and all imports are absolute.

experiments/src/experiments/confidence_aware_eql/engine/circuit_model.py
ganeshyadavyatham: All four addressed. GaussianMixtureCircuit is a dataclass holding an actual sklearn.mixture.GaussianMixture as a field; fit_gaussian_mixture_circuit fits it and the circuit is compiled from that mixture. There is no custom __init__ — only __post_init__ to compile the circuit and compute the column permutation. Class, method and field docstrings added. test_circuit_likelihood_matches_gaussian_mixture asserts the compiled circuit reproduces the mixture's likelihood.

experiments/src/experiments/confidence_aware_eql/engine/warning.py
ganeshyadavyatham: Both done. The class moved to exceptions.py and now subclasses Warning, with fields node_name, reason, log_likelihood (the abbreviation log_p is gone). I also added UnknownFeatureValueError for unrecognised categorical values. warning.py is deleted.

experiments/src/experiments/confidence_aware_eql/registry.py
ganeshyadavyatham: It discovered domain modules by scanning the folder for DOMAIN/SPEC attributes. That indirection isn't needed, so the file is deleted — a domain is now just a class you import.

experiments/src/experiments/confidence_aware_eql/tests/_assertions.py
ganeshyadavyatham: deleted

experiments/src/experiments/confidence_aware_eql/tests/registry.py
ganeshyadavyatham: Done — I went through AGENTS.md and the existing packages. Tests are now plain pytest modules under tests/ (test_engine.py, test_kitchen.py, test_eql_integration.py) with fixtures, no helper modules and no custom runner. tests/registry.py is deleted.

experiments/src/experiments/confidence_aware_eql/engine/__init__.py
ganeshyadavyatham: Done, it's empty.

experiments/src/experiments/confidence_aware_eql/engine/circuit_model.py
ganeshyadavyatham: Done. It's a dataclass now, with docstrings on the class, every method, and every field.

experiments/src/experiments/confidence_aware_eql/engine/domain.py
ganeshyadavyatham: This file is replaced by engine/schema.py and engine/feature.py, both fully documented at class, method and field level. Feature now carries a FeatureKind enum instead of a string.

experiments/src/experiments/confidence_aware_eql/engine/evaluator.py
ganeshyadavyatham: Added. ConfidenceAwareEvaluator and FamiliarityResult are documented, including parameter documentation on check.

experiments/src/experiments/confidence_aware_eql/engine/pipeline.py
ganeshyadavyatham: Both fixed. Parameters are documented in RST, and names are spelled out — n_components → number_of_components, std → standard_deviation, reg_covar → covariance_regularisation, and the tp/fp returns are gone.

experiments/src/experiments/confidence_aware_eql/kitchen_eql_demo.py
ganeshyadavyatham: Done — the demo script is removed and replaced by tests/test_eql_integration.py, which builds a real rule (an(entity(o).where(o.weight > 2.0))), registers the observer on the evaluation context, and asserts: the deterministic solutions are unchanged by observing; familiar objects raise no warning; a 50 kg glass cup is flagged with a log-likelihood below the threshold; and every warning names the node that rejected it. 13 tests pass.

experiments/src/experiments/confidence_aware_eql/eql_integration.py
ganeshyadavyatham: Documented. ConfidenceAwareEvaluationObserver now has a class docstring explaining how it hooks into evaluation, plus method and field docs.

experiments/src/experiments/confidence_aware_eql/engine/warning.py
ganeshyadavyatham: Documented; the class moved to exceptions.py and now subclasses Warning.

experiments/src/experiments/confidence_aware_eql/engine/threshold.py
ganeshyadavyatham: Documented and renamed to StandardDeviationThreshold.

experiments/src/experiments/confidence_aware_eql/engine/threshold.py
ganeshyadavyatham: Documented, along with a FamiliarityThreshold base class defining the interface.

experiments/src/experiments/confidence_aware_eql/engine/pipeline.py
ganeshyadavyatham: Done. build_evaluator and its four-tuple return are gone. ConfidenceModelBuilder.build() now returns a single ConfidenceAwareEvaluator with an explicit return type hint, and that evaluator carries the schema, the compiled circuit and the fitted threshold as fields — so there is no tuple to unpack anywhere.

experiments/src/experiments/confidence_aware_eql/engine/pipeline.py
ganeshyadavyatham: Both done. evaluate_detection and its (tp, fp) tuple are removed; the detection and false-positive behaviour is asserted directly in the tests instead. Where a call does produce a compound result, it is a dataclass: ConfidenceAwareEvaluator.check() returns a FamiliarityResult with log_likelihood, warning and an is_familiar property, and FeatureSchema.encode() returns an EncodedInstance with row, missing_features and is_complete. Abbreviated names (tp, fp, rng, bad) are gone.

experiments/src/experiments/confidence_aware_eql/engine/threshold.py
ganeshyadavyatham: Renamed to StandardDeviationThreshold, and its parameter is now number_of_standard_deviations instead of k. Abbreviated names are gone across the package — log_p → log_likelihood, n_components → number_of_components, std → standard_deviation, reg_covar → covariance_regularisation, and the tp/fp returns are removed entirely.

experiments/src/experiments/confidence_aware_eql/engine/evaluator.py
ganeshyadavyatham: it called check() with a fake node name and then popped the warning it had just appended. It's gone. check() is now side-effect free and returns a FamiliarityResult dataclass with log_likelihood, warning and an is_familiar property, so there's no hidden state and nothing to undo.

experiments/src/experiments/confidence_aware_eql/engine/threshold.py
ganeshyadavyatham: Thanks — yes,  PercentileThreshold.fit() looks at the log-likelihoods of the training instances and sets the cutoff at the first percentile, so the threshold comes from the data instead of a fixed number I picked. StandardDeviationThreshold (renamed from StdDevThreshold) does the same using the mean minus k standard deviations. Both now use a common FamiliarityThreshold base class, so another way of calibrating can be added easily. If you would prefer a method with a guarantee, such as a conformal bound on the training scores, I am happy to add it.

experiments/src/experiments/confidence_aware_eql/engine/pipeline.py
ganeshyadavyatham: pipeline.py now contains only the single ConfidenceModelBuilder dataclass (derives the schema, samples training instances, fits the model and threshold, returns one evaluator). The loose helpers you flagged are gone.

experiments/src/experiments/confidence_aware_eql/engine/datasets.py
ganeshyadavyatham: You're right — generate_dataset and sample_objects were fixture-like helpers sitting in the library. datasets.py is deleted. Training instances now come from TrainingSampler, which samples around explicit InstancePrototype objects that the caller supplies, and the test data is defined in the tests themselves as pytest fixtures.

experiments/src/experiments/confidence_aware_eql/engine/domain.py
ganeshyadavyatham: Agreed, that was a duplicate abstraction. Feature and FeatureKind are deleted along with domain.py. FeatureSchema now holds random_events.variable.Continuous variables directly — the same variable objects the circuit is built over — instead of wrapping them in a parallel class. Enumeration fields are encoded by the numeric value of their member so the feature vector stays continuous for the Gaussian mixture; if you'd prefer Symbolic variables with a proper set domain, that's a natural next step but would mean swapping the mixture for a model with symbolic leaves.

experiments/src/experiments/confidence_aware_eql/README.md
ganeshyadavyatham: Removed the README. Documentation is now in RST docstrings on every class, method and field, so Sphinx builds it from the source. I did not add a doc/ folder, since only krrood/doc is wired into the docs build at the moment — when this moves into krrood I will add a page there.

ganeshyadavyatham (commented):
Thank you both for the detailed review. I have gone through AGENTS.md and the existing packages and reworked the code to match.
What changed:

All __init__.py files are empty, no __all__, and all imports are absolute.
Everything is a dataclass with no custom __init__, and all classes, methods and fields have docstrings.
The circuit now holds a real sklearn GaussianMixture and compiles it into the sum-product circuit.
UnfamiliarSampleWarning is now an exception, and Material is an enum instead of a string.
Full names everywhere, no abbreviations, and no tuple returns — compound results are dataclasses.
Feature is gone; the schema now uses random_events Continuous variables directly, so there is no duplicate abstraction.
Removed README.md, run.py, registry.py, datasets.py, _template.py, _assertions.py and the demo script. Domains are now Symbol dataclasses whose fields are the features.
Added test_eql_integration.py, which runs the observer during real EQL evaluation and checks that the deterministic result does not change, that normal objects pass, and that a 50 kg glass cup is flagged.

All 13 tests pass.
Three things I want your help to decide, instead of guessing:

Whether pipeline.py should exist at all, or be folded into a classmethod on the evaluator.
Whether the check should go into the EQL backend instead of using an observer.
The move into krrood, which I understand is a separate PR.

Two small notes: KitchenObject does not use from __future__ import annotations, because the class diagram builder then fails to resolve the Material enum (department_and_employee.py also leaves it out). For the same reason, the EQL imports in the integration test are inside the function. Please tell me if there is a better way.

experiments/src/experiments/confidence_aware_eql/engine/training.py
```diff
@@ -0,0 +1,74 @@
...
+@dataclass
+class InstancePrototype:
+    """A representative instance together with the spread of its continuous features.
+
+    A prototype describes one cluster of familiar instances, such as "a cup".
+    Sampling draws continuous features from a normal distribution around the
+    prototype's values and keeps categorical features fixed.
+    """
+
+    instance: object
```
AbdelrhmanBassiouny: use Any instead of object as type hint

experiments/src/experiments/confidence_aware_eql/tests/test_eql_integration.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+def evaluate_heavy_object_rule(world, observer):
+    """Evaluate the rule "an object heavier than two kilograms" over a world.
+
+    The query-language factories are imported inside this function because
+    importing them before the first :class:`Symbol` instance is created prevents
+    the class-diagram builder from resolving the symbol hierarchy.
+    """
+    from krrood.entity_query_language.evaluation_context import (
```
AbdelrhmanBassiouny: move imports to top of module

experiments/src/experiments/confidence_aware_eql/tests/test_eql_integration.py
```diff
@@ -0,0 +1,88 @@
...
+        warning.log_likelihood < evaluator.threshold.value
+        for warning in observer.warnings
+    )
+
+
+def test_warning_names_the_node_that_flagged_the_instance(evaluator):
+    """Every warning records the node that rejected the instance."""
+    observer = ConfidenceAwareEvaluationObserver(evaluator, KitchenObject)
+    evaluate_heavy_object_rule([IMPOSSIBLE_CUP], observer)
+    assert all(warning.node_name for warning in observer.warnings)
```
AbdelrhmanBassiouny: Also assert the expected value for node_name, not just its existence.

experiments/src/experiments/confidence_aware_eql/tests/test_kitchen.py
```diff
@@ -1,25 +1,55 @@
...
+
+
+def test_missing_material_is_flagged():
+    """An object without a material is reported as incomplete."""
+    evaluator = build_kitchen_evaluator()
+    result = evaluator.check(
+        KitchenObject(0.25, 0.10, None), node_name="is_glass"
+    )
+    assert not result.is_familiar
+    assert "material" in result.warning.reason
```
AbdelrhmanBassiouny: Can the reason also state why? Like because it is not given/None?

experiments/src/experiments/confidence_aware_eql/eql_integration.py
```diff
@@ -1,77 +1,70 @@
...
+    warnings: List[UnfamiliarSampleWarning] = field(default_factory=list)
+    """Warnings collected during the evaluation, in the order they were raised."""
 
+    _checked_nodes: set = field(default_factory=set, repr=False)
+    """Instance and node pairs already checked, preventing duplicate warnings."""
 
-def run_with_confidence(query, evaluator: ConfidenceAwareEvaluator,
-                        object_extractor: ObjectExtractor):
-    """Evaluate an EQL query while checking confidence at each node.
+    def on_evaluate_enter(self, expression: Any, sources: Optional[Any] = None) -> None:
```
AbdelrhmanBassiouny: I think the source type hint should be OperationResult

experiments/src/experiments/confidence_aware_eql/eql_integration.py
```diff
@@ -1,77 +1,70 @@
...
+        identity = (id(instance), node_name)
+        if identity in self._checked_nodes:
+            return
+        self._checked_nodes.add(identity)
+
+        result = self.evaluator.check(instance, node_name=node_name)
+        if result.warning is not None:
+            self.warnings.append(result.warning)
+
+    def _bound_instance(self, sources: Optional[Any]) -> Optional[Any]:
```
AbdelrhmanBassiouny: same type hint

experiments/src/experiments/confidence_aware_eql/eql_integration.py
```diff
@@ -1,77 +1,70 @@
...
+        """Return the instance bound in the operation result, if there is one."""
+        if sources is None:
+            return None
+        bindings = getattr(sources, "bindings", None)
```
AbdelrhmanBassiouny: I think "bindings" always exists just access it directly

experiments/src/experiments/confidence_aware_eql/eql_integration.py
```diff
@@ -1,77 +1,70 @@
...
+            return None
+        for value in bindings.values():
+            if isinstance(value, self.instance_class):
+                return value
+        return None
+
+    @staticmethod
+    def _node_name(expression: Any) -> str:
+        """Return a readable name for the rule-tree node being evaluated."""
+        name = getattr(expression, "name", None)
```
AbdelrhmanBassiouny: I do not think expressions have a 'name' attribute . isn't it `_name_`?

experiments/src/experiments/confidence_aware_eql/eql_integration.py
```diff
@@ -1,77 +1,70 @@
...
+        bindings = getattr(sources, "bindings", None)
+        if not bindings:
+            return None
+        for value in bindings.values():
+            if isinstance(value, self.instance_class):
+                return value
+        return None
+
+    @staticmethod
+    def _node_name(expression: Any) -> str:
```
AbdelrhmanBassiouny: Instead of Any use SymbolicExpression

experiments/src/experiments/confidence_aware_eql/exceptions.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def __post_init__(self) -> None:
+        super().__init__(str(self))
+
+    def __str__(self) -> str:
+        return f"UnfamiliarSampleWarning at '{self.node_name}': {self.reason}"
+
+
+@dataclass
+class UnknownFeatureValueError(Exception):
```
AbdelrhmanBassiouny: inherit from DataclassException instead of Exception.

AbdelrhmanBassiouny (changes requested):

experiments/src/experiments/confidence_aware_eql/engine/circuit_model.py
```diff
@@ -0,0 +1,150 @@
...
+
+    .. note::
+        The circuit orders its variables alphabetically, so queries permute
+        incoming rows from domain order into circuit order before evaluation.
+    """
+
+    gaussian_mixture: GaussianMixture
+    """The fitted mixture that defines the component means, variances, and weights."""
+
+    feature_names: List[str]
```
tomsch420: i think these should be random events Variables. Gaussian mixtures can kind of model discrete variables.

experiments/src/experiments/confidence_aware_eql/engine/circuit_model.py
```diff
@@ -0,0 +1,150 @@
...
+    ProbabilisticCircuit,
+    SumUnit,
+    ProductUnit,
+    leaf,
+)
+
+
+@dataclass
+class GaussianMixtureCircuit:
+    """A tractable probabilistic circuit compiled from a Gaussian mixture.
```
tomsch420: you completely ignore correlations here from the covariance matrix of a gaussian. At least document that

experiments/src/experiments/confidence_aware_eql/engine/circuit_model.py
```diff
@@ -0,0 +1,150 @@
...
+    gaussian_mixture = GaussianMixture(
+        n_components=number_of_components,
+        covariance_type="diag",
+        random_state=random_seed,
+        reg_covar=covariance_regularisation,
+    ).fit(data)
+    return GaussianMixtureCircuit(gaussian_mixture, list(feature_names))
+
+
+def _select_component_count(
```
tomsch420, line 173: rst docstring

experiments/src/experiments/confidence_aware_eql/engine/evaluator.py
```diff
@@ -0,0 +1,71 @@
...
+    The evaluator encodes an instance through its schema, computes the
+    log-likelihood with the compiled circuit, and returns a
+    :class:`FamiliarityResult` carrying a warning when the instance is incomplete
+    or falls below the fitted threshold.
+    """
+
+    schema: FeatureSchema
+    """The feature schema used to encode instances."""
+
+    circuit: GaussianMixtureCircuit
```
tomsch420: this should only depend on the probabilistic model inteface, not on gaussian mixture circuit

experiments/src/experiments/confidence_aware_eql/engine/evaluator.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def check(self, instance: object, node_name: str) -> FamiliarityResult:
+        """Check one instance and report whether it is familiar.
+
+        :param instance: The instance whose features are read by the schema.
+        :param node_name: Name of the rule node performing the check, recorded on
+            any warning for traceability.
+        """
+        encoded = self.schema.encode(instance)
+        if not encoded.is_complete:
```
tomsch420: if its incomplete, just marginalize the missing information

experiments/src/experiments/confidence_aware_eql/engine/schema.py
```diff
@@ -0,0 +1,88 @@
...
+    """Names of features that were absent on the instance."""
+
+    @property
+    def is_complete(self) -> bool:
+        """Whether every feature was present and encodable."""
+        return not self.missing_features
+
+
+@dataclass
+class FeatureSchema:
```
tomsch420, line 12: i am pretty sure that something like this exists in the RSPN package. Please check that package completely.

experiments/src/experiments/confidence_aware_eql/engine/training.py
```diff
@@ -0,0 +1,74 @@
...
+    """The representative instance whose feature values anchor the cluster."""
+
+    continuous_spread: Dict[str, float]
```
tomsch420: there is a type, AnnotatedVariable that you can use instead of this

experiments/src/experiments/confidence_aware_eql/engine/training.py
```diff
@@ -0,0 +1,74 @@
...
+    """Standard deviation to sample around each continuous feature."""
+
+
+@dataclass
+class TrainingSampler:
+    """Generates familiar training instances from a set of prototypes."""
```
tomsch420: why not just generate these using EQL and the existing generative fully factorized backend?

experiments/src/experiments/confidence_aware_eql/tests/test_eql_integration.py
```diff
@@ -0,0 +1,88 @@
...
+NORMAL_POT = KitchenObject(3.00, 0.30, Material.METAL)
+IMPOSSIBLE_CUP = KitchenObject(50.0, 0.10, Material.GLASS)
+
+
+@pytest.fixture
+def evaluator():
+    """An evaluator trained on the familiar kitchen object prototypes."""
+    prototypes = [
+        InstancePrototype(
+            KitchenObject(0.25, 0.10, Material.CERAMIC), {"weight": 0.05, "size": 0.02}
```
tomsch420: can be done with plain eql + generative backend

experiments/src/experiments/confidence_aware_eql/exceptions.py
```diff
@@ -0,0 +1,51 @@
...
+    .. note::
+        This subclasses :class:`Warning` so it can be issued through
+        :func:`warnings.warn` or collected and raised, while still being a plain
+        dataclass with documented fields.
+    """
+
+    node_name: str
+    """Name of the rule-tree node that evaluated the instance."""
+
+    reason: str
```
tomsch420: you provide the same reason always. Do not make it a variable then

tomsch420 (changes requested):
run the docformatter script

experiments/src/experiments/confidence_aware_eql/eql_integration.py
```diff
@@ -0,0 +1,70 @@
...
+    instance_class: Type
+    """The class of the instances that should be checked."""
+
+    warnings: List[UnfamiliarSampleWarning] = field(default_factory=list)
+    """Warnings collected during the evaluation, in the order they were raised."""
+
+    _checked_nodes: set = field(default_factory=set, repr=False)
+    """Instance and node pairs already checked, preventing duplicate warnings."""
+
+    def on_evaluate_enter(self, expression: Any, sources: Optional[Any] = None) -> None:
```
AbdelrhmanBassiouny: Instead of subscribing to evaluate_enter, subscribe to (implement) `on_conclusions_processed` beacuse you model check should run for rule conclusions, not for the evaluation of every expression and for every query. What we want is that your model can raise a custom exception that we can intercept and handle when a rule fires for a case that is unfamiliar.

experiments/src/experiments/confidence_aware_eql/domains/kitchen.py
davidprueser, line 1: If you need example classes, use existing classes from the semantic_digital_twin package, see https://github.com/cram2/cognitive_robot_abstract_machine/blob/main/semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/semantic_annotations.py

experiments/src/experiments/confidence_aware_eql/tests/__init__.py
davidprueser, line 1: Put everything related to testing into the test folder at root level

experiments/src/experiments/confidence_aware_eql/engine/schema.py
davidprueser, line 12: I see what you are trying to do here. What you want is actually already implemented. Take a look at https://github.com/cram2/cognitive_robot_abstract_machine/blob/main/krrood/src/krrood/parametrization/feature_extraction/feature_extractor.py, there is a class FeatureExtractor that creates a list of symbolic features based on a list of instances. You can then create a dataframe on that, preprocess it and infer random_events variables by putting the resulting dataframe into this function https://github.com/cram2/cognitive_robot_abstract_machine/blob/main/probabilistic_model/src/probabilistic_model/learning/jpt/variables.py#L59

experiments/src/experiments/confidence_aware_eql/engine/training.py
ganeshyadavyatham: Done. I changed object to Any here and in all other places in the package.

experiments/src/experiments/confidence_aware_eql/eql_integration.py
ganeshyadavyatham: Done. The parameter is now expression: SymbolicExpression, imported from krrood.entity_query_language.core.base_expressions.

experiments/src/experiments/confidence_aware_eql/eql_integration.py
ganeshyadavyatham: Done. It is now sources: Optional[OperationResult].

experiments/src/experiments/confidence_aware_eql/tests/test_eql_integration.py
ganeshyadavyatham: Done. The imports are at the top now.

experiments/src/experiments/confidence_aware_eql/tests/test_eql_integration.py
ganeshyadavyatham: Done. The test now checks the actual name:
 assert KitchenObject.__name__ in {w.node_name for w in observer.warnings}

experiments/src/experiments/confidence_aware_eql/tests/test_kitchen.py
ganeshyadavyatham: Done. The reason now says: "the features [material] were not given (their value is None), so the instance cannot be scored". The test checks that the reason contains both material and not given, and that log_likelihood is None.

experiments/src/experiments/confidence_aware_eql/eql_integration.py
ganeshyadavyatham: Done. _bound_instance also takes Optional[OperationResult] now.

experiments/src/experiments/confidence_aware_eql/eql_integration.py
ganeshyadavyatham: yes,OperationResult.bindings is a required field, so I removed the getattr and now read sources.bindings directly.

experiments/src/experiments/confidence_aware_eql/eql_integration.py
ganeshyadavyatham: yes,it is _name_. I changed it. The node names are also better now: the warning says KitchenObject instead of the old KitchenObject.name.

experiments/src/experiments/confidence_aware_eql/exceptions.py
ganeshyadavyatham: Done. UnknownFeatureValueError now inherits from DataclassException and implements error_message() and suggest_correction(). I kept UnfamiliarSampleWarning as a Warning, because it is not an error -- the evaluation still finishes and returns a result, it only reports doubt. Please tell me if  it needs to be a DataclassException too.

experiments/src/experiments/confidence_aware_eql/engine/circuit_model.py
ganeshyadavyatham: I added a warning in the class docstring saying that the mixture is fitted with a diagonal covariance, that only the diagonal is used, and that inside one component the variables are treated as independent. It also says that dependencies are only captured through the mixture weights, and that a full covariance would need leaves over several variables or a model like a joint probability tree.

experiments/src/experiments/confidence_aware_eql/engine/circuit_model.py
ganeshyadavyatham: Done. It has an RST docstring now with :param: for every parameter and :return:. All other functions in the package were checked too.

experiments/src/experiments/confidence_aware_eql/engine/evaluator.py
ganeshyadavyatham: Done. The field is now model: ProbabilisticModel. The evaluator only calls marginal and log_likelihood, so any tractable model can be used. The Gaussian mixture is only used when the model is built, not when instances are checked.

experiments/src/experiments/confidence_aware_eql/engine/evaluator.py
ganeshyadavyatham: Now the missing variables are marginalised out and the instance is scored with the features it does have. There is a new test: an object with no material is still scored and comes out familiar, and if the weight is also impossible it is flagged and the warning says which features were marginalised.

experiments/src/experiments/confidence_aware_eql/engine/schema.py
ganeshyadavyatham: I checked the package. The closest one is krrood.parametrization.feature_extraction.FeatureExtractor, which is used by RelationalProbabilisticCircuit. It does the same thing (find variables, turn an instance into a row), but from_instances needs DataAccessObject instances, so the domain class must be mapped by ORMatic. My KitchenObject is a plain Symbol dataclass, so I cannot use it directly yet. Should I make the domain class a DAO so FeatureExtractor can be used.

experiments/src/experiments/confidence_aware_eql/engine/training.py
ganeshyadavyatham: Done. InstancePrototype is deleted. The moments are now given as AnnotatedVariable from probabilistic_model.learning.jpt.variables, using its mean and standard_deviation fields.

experiments/src/experiments/confidence_aware_eql/exceptions.py
```diff
@@ -0,0 +1,77 @@
+from __future__ import annotations
+
+from dataclasses import dataclass
+
+from krrood.exceptions import DataclassException
+from typing_extensions import Any
+
+
+@dataclass
+class UnfamiliarSampleWarning(Warning):
```
davidprueser: We dont do warnings this way. Either raise a dedicated exception if you want to prohibit certain behavior or create a warning using a logger, but no dedicated warning classes that get initialized.

davidprueser (changes requested):
Good work. Lets work on integrating existing functionality. Have a look at the feature_extraction and rspn tests to get a better understanding of how we currently extract features and learn probabilistic models :)

tomsch420:
resolve resolved conversations and reply to them @ganeshyadavyatham

experiments/src/experiments/confidence_aware_eql/engine/schema.py
tomsch420: yes, also the author of RSPNs told you in the next review mr. claude

experiments/src/experiments/confidence_aware_eql/engine/evaluator.py
```diff
@@ -0,0 +1,107 @@
...
+@dataclass
+class FamiliarityResult:
+    """
+    The outcome of checking one instance against the learned distribution.
+    """
+
+    log_likelihood: float
+    """Log-likelihood of the instance under the model."""
+
+    warning: Optional[UnfamiliarSampleWarning]
```
LucaKro: if its optional, give it a default None

LucaKro (changes requested):

experiments/src/experiments/confidence_aware_eql/engine/circuit_model.py
ganeshyadavyatham: feature_names: List[str] is now variables: List[Variable] (from random_events), so the circuit stores the actual variable objects instead of just their names.
For the second point,   treating the categorical feature as continuous is incorrect. Renumbering the enum changes the log-likelihood.
 now I have  switched to the framework's intended flow (FeatureExtractor → dataframe → infer_variables_from_dataframe), which correctly infers it as a Symbolic variable. The  thing which is left is training a JPT with symbolic leaves, which needs pyjpt. It does not build on my aarch64 machine, so I will get access to an x86 environment for that.

experiments/src/experiments/confidence_aware_eql/engine/evaluator.py
ganeshyadavyatham: done, added = None as the default

experiments/src/experiments/confidence_aware_eql/exceptions.py
ganeshyadavyatham: Replaced the Warning subclass with UnfamiliarSampleException

experiments/src/experiments/confidence_aware_eql/eql_integration.py
ganeshyadavyatham: done.the observer subscribes to on_conclusions_processed instead of on_evaluate_enter, so the check runs only when a rule's conditions hold and its conclusions fire, not on every expression. When the fired rule's instance is unfamiliar it raises UnfamiliarSampleException, which the caller intercepts to handle the out-of-distribution case instead of applying the deterministic conclusion.

experiments/src/experiments/confidence_aware_eql/engine/training.py
ganeshyadavyatham: I changed this . Now the training instances are generated through EQL with underspecified(KitchenObject)(weight=..., size=..., material=...) and the ProbabilisticBackend with a fully factorized model. The moments (mean and variance) come from the cluster's AnnotatedVariables.

experiments/src/experiments/confidence_aware_eql/tests/test_eql_integration.py
ganeshyadavyatham: Done. This is now built with plain EQL and the generative backend — underspecified() builds the match statement and ProbabilisticBackend generates the instances. The old prototype list was removed.

experiments/src/experiments/confidence_aware_eql/exceptions.py
ganeshyadavyatham: Removed. There is no reason field now. The exception builds its message in error_message() from the node name, log-likelihood and threshold, so there is no repeated string stored as a variable

experiments/src/experiments/confidence_aware_eql/engine/schema.py
ganeshyadavyatham: I did this. I mapped KitchenObject as a DAO. I generated a small self-contained ORMatic interface for just the domain package (rather than regenerating the full shared experiments interface), so the mapping stays local to this feature. to_dao(KitchenObject(...)) returns a KitchenObjectDAO.
The full pipeline runs: to_dao → FeatureExtractor.from_instances → create_dataframe → infer_variables_from_dataframe, and material comes out as a Symbolic variable (weight and size as Continuous).

experiments/src/experiments/confidence_aware_eql/tests/__init__.py
ganeshyadavyatham: Done. Moved the tests to test/experiments_test/confidence_aware_eql/

experiments/src/experiments/confidence_aware_eql/engine/schema.py
ganeshyadavyatham: I'm stuck on  scoring the fitted JPT. After pc = tree.fit(df), the training data itself scores -inf for every row — both with pc.log_likelihood(df[order].to_numpy()) and with pc.log_likelihood(tree.preprocess_data(df)). 
```
import numpy as np, pandas as pd, random
from enum import Enum
from probabilistic_model.learning.jpt.jpt import JointProbabilityTree
from probabilistic_model.learning.jpt.variables import infer_variables_from_dataframe
class Material(Enum):
    CERAMIC="ceramic"; GLASS="glass"; METAL="metal"
rng=np.random.default_rng(0); random.seed(0)
weight=np.concatenate([rng.normal(0.25,0.05,80),rng.normal(2.5,0.3,80),rng.normal(3.0,0.4,80)])
size=np.concatenate([rng.normal(0.10,0.02,80),rng.normal(0.25,0.03,80),rng.normal(0.30,0.03,80)])
mats=[Material.CERAMIC]*80+[Material.GLASS]*80+[Material.METAL]*80
df=pd.DataFrame({"weight":weight,"size":size,"material":mats})
tree=JointProbabilityTree(annotated_variables=infer_variables_from_dataframe(df),min_samples_per_leaf=10)
pc=tree.fit(df)
order=[v.name for v in pc.variables]
print("raw:", int(np.isneginf(pc.log_likelihood(df[order].to_numpy())).sum()), "of 240")
print("preprocessed:", int(np.isneginf(pc.log_likelihood(tree.preprocess_data(df))).sum()), "of 240")
```

Output: both print 240 of 240. What's the correct way to score a dataframe against a fitted JPT with a symbolic variable? For now I kept the Gaussian mixture

experiments/src/experiments/confidence_aware_eql/domains/kitchen.py
ganeshyadavyatham: I looked into this properly, the physical features do exist on the real classes (a Cup's mass is at root.inertial.mass, dimensions from the collision geometry), so FeatureExtractor walking the DAO graph would extract them, and the class (Cup/Pot/Bowl) is the natural categorical feature. I'm going to work on switching from the synthetic KitchenObject to the real SDT classes. Two things I need to solve first: constructing them needs a World, and to_dao needs geometry_msgs in my environment. I'll follow up here once I have the pipeline running on the real classes.

tomsch420: so did you do it? if so, remove this file

experiments/src/experiments/confidence_aware_eql/engine/circuit_model.py
tomsch420, line 1: this file still doesnt really make sense to me. Its neither a proper learning method implementation nor a full probabilistic model implementation.

experiments/src/experiments/confidence_aware_eql/engine/evaluator.py
```diff
@@ -0,0 +1,86 @@
...
+
+    log_likelihood: float
+    """Log-likelihood of the instance under the model."""
+
+    is_familiar: bool
+    """Whether the instance was accepted as familiar."""
+
+
+@dataclass
+class ConfidenceAwareEvaluator:
```
tomsch420, line 26: as bass told you, make the correct subclass here, than you dont need all of this

experiments/src/experiments/confidence_aware_eql/engine/schema.py
```diff
@@ -0,0 +1,100 @@
...
+from dataclasses import dataclass, fields
+from enum import Enum
+
+import numpy as np
+from random_events.variable import Continuous, Variable
+from typing_extensions import Any, List, Type
+
+
+@dataclass
+class FeatureSchema:
```
tomsch420, line 12: remove this class finally

experiments/src/experiments/confidence_aware_eql/engine/training.py
```diff
@@ -0,0 +1,124 @@
...
+from probabilistic_model.learning.jpt.variables import AnnotatedVariable
+from probabilistic_model.probabilistic_circuit.rx.helper import fully_factorized
+from probabilistic_model.probabilistic_model import ProbabilisticModel
+from typing_extensions import Any, Dict, List
+
+from experiments.confidence_aware_eql.engine.schema import FeatureSchema
+
+
+@dataclass
+class MomentRegistry(ModelRegistry):
```
tomsch420, line 18: this class name also makes no sense. A moment is not a model, but this builds a model

experiments/src/experiments/confidence_aware_eql/engine/training.py
```diff
@@ -0,0 +1,124 @@
...
+
+        :param variable: A variable of the match statement, whose name may carry a class
+            prefix such as ``KitchenObject.weight``.
+        :return: The bare feature name, for example ``weight``.
+        """
+        return variable.name.split(".")[-1]
+
+
+@dataclass
+class FamiliarCluster:
```
tomsch420, line 67: also architecturally not sound

tomsch420 (changes requested):
this is not working for me this way @ganeshyadavyatham . It looks like you answer everything with claude and you are not learning anything really here. I think this PR is going the wrong way

ganeshyadavyatham:
> this is not working for me this way @ganeshyadavyatham . It looks like you answer everything with claude and you are not learning anything really here. I think this PR is going the wrong way

yes sir, I am sorry , I used Claude for coding purpose and fetch information and also  to correct my statements while replying to comments . 
earlier it all went messy because I was having some compatibility issues due to my aarch64 machine.  now I am using GitHub code space . so currently I have created a new branch locally and started working step by step again. so now I am using real sdt data samples (the raw tree object -> data access object using ormatic interface.py which helps them register in respective object's dao) . once we  get the dao we extract the features based on the reference code shared by David and from there we create a dataframe and I have skipped preprocessing and went with inferring variables from data frame directly because I am using string format for object class not enum and preprocessing handles only enum/bool columns .I am not sure if this is fine or not but for me it worked without doing preprocessing. with this we end up with a feature vector and we fit it to jpt . it scores the loglikelihood of all objects . after training I calibrate single threshold from training scores and test a new object if it is a  familiar object. I have done till here .

can you please give me an appointment to discuss this in person and gain more clarity about few choices regarding designing.

tomsch420:
> > this is not working for me this way @ganeshyadavyatham . It looks like you answer everything with claude and you are not learning anything really here. I think this PR is going the wrong way
> 
> yes sir, I am sorry , I used Claude for coding purpose and fetch information and also to correct my statements while replying to comments . earlier it all went messy because I was having some compatibility issues due to my aarch64 machine. now I am using GitHub code space . so currently I have created a new branch locally and started working step by step again. so now I am using real sdt data samples (the raw tree object -> data access object using ormatic interface.py which helps them register in respective object's dao) . once we get the dao we extract the features based on the reference code shared by David and from there we create a dataframe and I have skipped preprocessing and went with inferring variables from data frame directly because I am using string format for object class not enum and preprocessing handles only enum/bool columns .I am not sure if this is fine or not but for me it worked without doing preprocessing. with this we end up with a feature vector and we fit it to jpt . it scores the loglikelihood of all objects . after training I calibrate single threshold from training scores and test a new object if it is a familiar object. I have done till here .
> 
> can you please give me an appointment to discuss this in person and gain more clarity about few choices regarding designing.

yes, we can talk after the exam today

PR #402, opened 2026-07-01
Semdt specifications rewrite

LucaKro:
# Feature: World, Entity and Annotation Specifications

This PR introduces a *specification* API layer for the semantic digital twin: world-independent, reusable
descriptions of bodies, regions, connections, semantic annotations, robots and whole worlds, which are
turned into real world entities by an explicit materialization step.

Until now, building a world meant mutating one. Every annotation type carried a
`create_with_new_body_in_world` / `create_with_new_region_in_world` factory that took a world, a pose, a
scale and a flat bag of connection parameters (`active_axis`, `connection_limits`, `connection_multiplier`,
`connection_offset`, `**kwargs`), created a body, guessed a connection type from a `_parent_connection_type`
classproperty, and wired everything up in one call.

The specification layer separates *what an entity is* from *where it ends up*. A specification can be
built long before any world exists. Materialization is a separate verb — `spawn` for things that create an
entity plus the connection carrying it, `connect` for a connection joining two entities that already exist,
`to_domain_object` for a world-independent object.

Related documentation: the new user-guide page
[Building Worlds with Specifications](https://lucakro.github.io/cognitive_robot_abstract_machine/semantic_digital_twin/examples/building_worlds_with_specifications.html),
plus new self-assessment exercises and quizzes for part-whole relationships, annotation factories, and
specifications.

- New `semantic_digital_twin.api` module holding the whole specification API: entity, connection,
  annotation, robot and world specifications.
- Connection parameters are typed per connection family (`PrismaticConnectionSpecification(axis=...)`,
  `FixedConnectionSpecification()`, …) instead of one flat bag of optional keyword arguments.
- Annotation types expose three deliberate hooks: `get_specification`, `get_default_root_specification`, and
  `parent_connection_specification` — with every geometry parameter living on exactly one of them.
- Nested annotation parts are part of the specification (`part_specifications=...`), spawned and mounted
  automatically, and validated against the annotation's part-whole fields before any world is touched.
- `WorldSpecification` describes an environment plus its robots and starting objects, and can be applied
  repeatedly to yield independent worlds; `RobotSpecification` merges a robot as
  `root -> odom -> drive -> robot` with the drive type derived from the robot's own mobile base.
- The `IsPartWholeRelationship` metadata now says whether mounting a part cuts its volume out of the whole,
  so an aperture carves the wall it opens but not the door that fills it.
- Test-suite tooling: ORM regeneration is guarded against concurrent `pytest-xdist` workers, and
  `scripts/regenerate_all_orm.py` is runnable from any working directory and now covers `segmind`.

---

### `semantic_digital_twin.api` — the specification surface

The new module holds one small hierarchy:

- `NamedSpecification` — carries a name and normalizes it into a `PrefixedName` at materialization time. It
  deliberately declares *no* materialization contract, so entity specs and connection specs can derive
  incompatible verbs from it without one masquerading as the other.
- `SpawnSpecification[TWorldEntity]` — anything that materializes an entity *together with* the connection
  attaching it to its parent, via `spawn(world, name=..., parent=..., parent_T_self=...)`.
- `ConnectionSpecification[TConnection]` — a connection joins two pre-existing entities, so it is not a
  `SpawnSpecification`; it is materialized with `connect(world, child, parent=..., …)`. The connection type
  is read from the bound generic parameter and every public field is forwarded to that type's
  `create_with_dofs`, so a new connection family is one small subclass. Concrete families:
  `FixedConnectionSpecification`, `Connection6DoFSpecification`, and the single-DoF
  `PrismaticConnectionSpecification` / `RevoluteConnectionSpecification` (axis mandatory, plus multiplier,
  offset and DoF limits).
- `KinematicStructureEntitySpecification[TKinematicStructureEntity]` — prototype shapes, child
  specifications, a default `parent_T_self`, and the connection attaching it to its parent. Materialization
  copies the shapes and the pose, so a specification never becomes bound to an entity or a world and can be
  spawned any number of times. `BodySpecification` and `RegionSpecification` bind the concrete entity type;
  convenience constructors (`box`, `sphere`, `cylinder`, `mesh`, `from_event`, point-cloud hulls) cover the
  common geometry cases, and `scale` measures a specification before anything is spawned.
- `SemanticAnnotationWithRootSpecification[TSemanticAnnotation]` — the annotation type, its root entity
  specification, inert `annotation_kwargs`, and `part_bindings`. The annotation's parent connection lives on
  `root_specification.connection_specification` and nowhere else, because the root is what attaches to the
  parent; leaving it unset falls back to the annotation type's own `parent_connection_specification()`.

Two design points worth calling out, both of which were tried the other way first:

- **Parts are a list of bindings, not a mapping.** `PartSpecificationBinding(field_name=..., specifications=[...])`
  exists because a `dict` field is dropped on the way to the database, which would have made nesting
  non-persistable. The ergonomic mapping form still exists at the call site — `get_specification` accepts
  `part_specifications={"handles": [...]}` and converts it.
- **No validation of the supplied parent connection.** An earlier iteration made it mandatory at
  construction, a later one restricted it to `MechanicalJoint`s; both were reverted. A wrong connection type
  is the caller's business for now, and `spawn` simply falls back to the annotation type's default when none
  is stated.

### Semantic annotation API

`_create_with_connection_in_world` and the `_parent_connection_type` classproperty are gone. In their place,
each annotation type owns three methods:

- `get_specification(name, root_specification, *, parent_connection_specification, annotation_kwargs, part_specifications)`
  — the only annotation-specification constructor. `root_specification` is required; there is deliberately no
  implicit default-geometry fallback here.
- `get_default_root_specification(name=None, scale=None, connection_specification=None, **per_type_geometry)`
  — abstract on `HasRootKinematicStructureEntity`, implemented once per root kind (`HasRootBody` returns a
  `BodySpecification`, `HasRootRegion` a `RegionSpecification`) and overridden per type to add its own
  geometry parameters (a `Handle`'s `thickness`, a case's `wall_thickness`, a `Door`'s `color`). Every
  geometry parameter lives here and nowhere else, so a type's geometry is described in exactly one place.
- `parent_connection_specification()` — replaces the connection-type classproperty and returns a full
  specification. `Hinge` returns a revolute one, `Slider` a prismatic one, both taking `axis`, `multiplier`,
  `offset` and `dof_limits`; every override must stay callable with no arguments, since that is how the
  default is resolved at spawn time.

The encapsulated `Slider.parent_connection_specification(axis=...)` form was kept over the shorter
`PrismaticConnectionSpecification(axis=...)` at call sites so that a future change of a type's default
connection family stays a one-line change per class.

`create_with_new_body_in_world` / `create_with_new_region_in_world` survive as thin wrappers over
`get_specification(...).spawn(...)`, but with a much smaller signature: `name` is now a plain `str`, the four
loose connection parameters collapse into one optional `parent_connection_specification`, and `**kwargs` is
gone. Deleting these wrappers entirely is a follow-up: they have ~281 call sites across tests, notebooks and
downstream packages, and doing it here is out of scope for this PR in my opinion.

### Part-whole relationships

- `IsPartWholeRelationship` moved out of `mixins.py` into its own `semantic_annotations/part_whole.py`, so
  the annotation mixins that declare part-whole fields and the specification API that fills them can both
  depend on the vocabulary without depending on each other.
- The metadata gained `removes_part_geometry_from_whole`: mounting a part into that field subtracts the
  part's volume from the whole's collision and visual geometry. This is a property of the *relation*, not of
  the part — the same `EntryWay` cuts the wall it is an aperture of, but not the door whose passage it marks
  — which is why `_mount_strategy` now receives the relationship alongside the whole.
- Field discovery moved to `WrappedClass.fields_with_metadata(IsPartWholeRelationship)` (see below),
  replacing a module-level `lru_cache`d helper function.
- `SemanticAnnotationWithRootSpecification` validates bindings in `__post_init__`, before any world mutation:
  unknown field names raise `UnknownPartWholeRelationshipField`, several parts on a singular field raise
  `PartWholeCardinalityError`, and part-whole fields smuggled through `annotation_kwargs` raise
  `PartWholeFieldInAnnotationKwargs`.

### Robots

- `MobileBase` became `MobileBase[TGenericDrive]`: each concrete base binds its drive connection as the
  generic parameter (`PR2MobileBase(MobileBase[OmniDrive], …)`,
  `TiagoMobileBase(MobileBase[DifferentialDrive], …)`), and `get_drive_connection_type()` resolves it. The
  drive type was previously something every caller and test fixture had to know and pass in by hand.
- `RobotSpecification` describes which robot, where its `odom` sits in the world, and where the robot starts
  within `odom`. `spawn` parses the robot from its own description and merges it as
  `world.root -> odom -> drive -> robot`, marking an active drive as hardware-controlled and applying the
  localization and start poses afterwards. Each robot gets its own `odom` body, prefixed with its id, so
  several robots in one world stay distinguishable.
- `AbstractRobotPart` now rejects both `create_with_new_body_in_world` and `get_default_root_specification`
  with `UselessConceptError`: robot-part bodies come out of the parsed URDF and cannot be spawned from a
  scale.

### World, geometry and connection plumbing

- `World.create_with_root_body(root_body_name="map")` replaces the recurring three-line "make a world, make a
  root body, add it" incantation, taking an unprefixed string.
- `World.remove_branch_from_world(root)` atomically removes a subtree: the root, its descendants, every
  connection inside the branch, and the branch's connection to its parent.
- `Shape.copy_for_world(world)` / `ShapeCollection.copy_for_world(world)` became
  `copy_without_reference_frame()`. The old name suggested a world-bound copy while the implementation only
  ever stripped the frame — and specifications need exactly the frame-free copy.
- `ShapeCollection` now raises `MismatchingWorld` when asked to transform a shape across worlds, instead of
  silently attempting the transform; the cross-frame-within-one-world case still warns and transforms.
- `Shape.as_shape_collection()` and `Shape.recenter_origin()` support building geometry for specifications.
- `Body.from_shape_collection` accepts an optional separate `visuals_shape_collection`; collision shapes are
  still used for both when it is omitted.
- The soft connections `PiecewiseConstantCurvatureConnection` and `CosseratRodConnection` now declare their
  `create_with_dofs` parameters explicitly (keyword-only DoF ids and segment length) rather than accepting
  `*args, **kwargs` — required for the specification layer to forward parameters, and it makes the shared-DoF
  contract of these connections visible in the signature.
- The module-level `IDGenerator` instances in `world.py`, `geometry.py` and `world_entity.py` are gone;
  default body names derive from the body's own `id`. `IDGenerator` and the unused `copy_lru_cache` were
  removed from `utils.py`.
- `JointState` JSON now references its connections by child entity id rather than by connection name, because
  two instances of the same robot description name their joints identically.

### `krrood`

- `WrappedClass.fields_with_metadata(metadata_type)` returns the fields carrying a given `FieldMetadata`
  subtype, in declaration order, and `FieldMetadata.of_wrapped_field(wrapped_field)` is its per-field
  counterpart. Both replace ad-hoc per-call-site filtering.
- `WrappedField.is_optional` now recognises the PEP 604 spelling `X | None` (`types.UnionType`) in addition
  to `Optional[X]`. Without this, PEP 604 optional fields were silently left unmapped by the ORM.
- `classproperty` is now generic and fully typed, so `classproperty`-backed attributes keep their type.
- The unresolvable-`TYPE_CHECKING`-import warning dropped to `debug`: it fires routinely during ORM
  generation and drowned real output.

### Exceptions

New, all with actionable `suggest_correction` text: `MissingConnectionParentError`,
`MissingConnectionAxisError`, `PartWholeCardinalityError`, `PartWholeFieldInAnnotationKwargs`.
`UnknownPartWholeRelationshipField` now carries the annotation *type* rather than an instance, since
specification-time validation happens before any instance exists.

### Documentation and self-assessment

- New user-guide page `examples/building_worlds_with_specifications.md` walking from a single body
  specification through nesting, geometry constructors, regions, connections, annotation specifications,
  to-many parts, geometry-cutting parts, parts that carry the whole, and finally full world specifications
  with robots.
- New exercises: `8_semantic_annotation_part_whole`, `9_semantic_annotation_factories`,
  `10_building_worlds_with_specifications`; new quizzes for part-whole relationships and specifications; both
  wired into `_toc.yml`, `user_guide.md` and `quizzes_intro.md`.
- Existing examples and exercises were updated to the new API, and docstrings across the touched modules were
  reformatted with `scripts/format_docstrings.py`.

### Test tooling

- `scripts/regenerate_all_orm.py` was rewritten around an `OrmPackage` dataclass: it resolves every path
  relative to the script, so it runs from any working directory, it covers `segmind` in addition to
  `semantic_digital_twin` / `coraplex` / `experiments`, and it clears *all* interfaces before regenerating
  any of them, so a stale interface cannot be imported mid-generation.
- The `pytest_configure` ORM-generation hooks in the `coraplex` and `segmind` test packages now return early
  on `pytest-xdist` workers: every worker runs the hook, and concurrent writers truncate the interface file
  while another process is formatting it. Both are covered by tests asserting the interface file is left
  untouched on a worker.
- `world_with_urdf_factory` no longer takes a drive connection type; it asks the robot annotation.
- New `test_specifications.py` (~1600 lines) covers spawning, connection parameters, shape constructors,
  robot and world specifications, default-geometry parity with the old factories, nested parts, cardinality
  and kwargs validation, non-shared factory defaults, zero-argument connection resolution, and round-tripping
  specifications through persistence.

---

- **`HasRootKinematicStructureEntity._parent_connection_type`**: removed. Use
  `parent_connection_specification()`, which returns a full connection specification instead of a bare type.
- **`HasRootKinematicStructureEntity._create_with_connection_in_world`**: removed. The equivalent path is
  `get_specification(...).spawn(world, ...)`.
- **`Shape.copy_for_world` / `ShapeCollection.copy_for_world`**: removed in favour of
  `copy_without_reference_frame()`.
- **`semantic_digital_twin.utils.IDGenerator` and `copy_lru_cache`**: removed; both were unused after entity
  names started deriving from entity ids.
- **`create_with_new_body_in_world` / `create_with_new_region_in_world`**: not removed here, but they are now
  thin wrappers over the specification API and are intended to be deleted in a follow-up PR.

---

- **Annotation factory signatures**: `create_with_new_body_in_world` / `create_with_new_region_in_world` now
  take `name: str` (not `PrefixedName`) and a single optional `parent_connection_specification`. The
  parameters `connection_limits`, `active_axis`, `connection_multiplier`, `connection_offset` and `**kwargs`
  are gone. *Mitigation*: pass the plain name, and replace connection parameters with e.g.
  `parent_connection_specification=Hinge.parent_connection_specification(axis=Vector3.Z())`.
- **`_parent_connection_type` overrides no longer take effect.** Types that customized their parent
  connection must override `parent_connection_specification()` instead, and the override must remain callable
  with no arguments.
- **`_mount_strategy(self, whole)` is now `_mount_strategy(self, whole, relationship)`.** Overrides must
  accept the relationship metadata.
- **`IsPartWholeRelationship` moved** from `semantic_annotations.mixins` to `semantic_annotations.part_whole`.
  *Mitigation*: update the import.
- **`copy_for_world` callers** must switch to `copy_without_reference_frame()` (no `world` argument).
- **`ShapeCollection` raises `MismatchingWorld`** when a shape's reference frame belongs to a different world
  than the collection's, where it previously attempted a transform.
- **Soft connection `create_with_dofs`** (`PiecewiseConstantCurvatureConnection`, `CosseratRodConnection`) now
  requires explicit keyword arguments for the DoF ids and segment length; positional/`**kwargs` calls break.
- **`JointState` JSON format changed** from a `"connections"` list of names to a `"child_ids"` list of entity
  ids. Previously serialized joint states will not load.
- **`UnknownPartWholeRelationshipField.annotation`** is now a `Type[HasRootBody]` rather than an instance.
- **`sv_ratio_tol` renamed to `singular_value_ratio_tolerance`** in `Mesh.from_3d_points` and
  `KinematicStructureEntity.from_3d_points`.
- **`MobileBase` is now generic**: concrete mobile bases must bind their drive type, e.g.
  `class MyMobileBase(MobileBase[OmniDrive])`, or `get_drive_connection_type()` cannot be resolved.
- **`world_with_urdf_factory` (test fixture) lost its `drive_connection_type` parameter**; it is derived from
  the robot annotation.

---

## Other Changes

- `TrashCan` is now a `HasCaseAsRootBody` (with a `hole_direction`) rather than a plain `HasRootBody`, so it
  gains the hollow-case geometry and supporting-surface behaviour of the other case-shaped furniture.
- `EntryWay` gained a default root specification and a `default_scale`.
- The ProcTHOR and SAGE-10k adapters now carry plain `str` names internally and wrap them into `PrefixedName`
  only where a domain object requires one, matching the new factory signatures.
- `PrefixedName` fields are now documented, and it no longer imports the JSON serializer it did not use.
- The generated `ormatic_interface.py` files for `semantic_digital_twin`, `coraplex`, `segmind` and
  `experiments` were regenerated to reflect the new specification classes.

---


.github/workflows/build_all_docs.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - semdt_specifications_rewrite
```
ichumuh: don't forget to delete this

semantic_digital_twin/examples/building_worlds_with_specifications.md
```diff
@@ -0,0 +1,494 @@
...
+Every specification carries a `name` and knows how to turn itself into a domain object. There
+are two distinct contracts, and which one a specification offers tells you what kind of thing it
+describes:
+
+- **`spawn(world, ...)`** — used by specifications that describe an *entity together with the
+  connection that attaches it to a parent*. Spawning materializes the entity, attaches it, and
+  recursively materializes its children, all inside one modification block.
+- **`connect(world, parent=, child=)`** — used by [connection
+  specifications](connecting-existing-entities), which join two *already existing* entities. A
+  connection is not an entity, so it is never spawned.
```
ichumuh: but connection is a world entity?

semantic_digital_twin/examples/building_worlds_with_specifications.md
```diff
@@ -0,0 +1,494 @@
...
+
+There is also `to_domain_object(...)`, which materializes an entity in isolation (without
```
ichumuh: add to list

semantic_digital_twin/examples/building_worlds_with_specifications.md
```diff
@@ -0,0 +1,494 @@
...
+attaching it to any world) — useful when you want a free-standing `Body` to pass somewhere else.
+
+Throughout this guide we need a world with a root to materialize into.
+`World.create_with_root_body()` gives us exactly that: a fresh world whose single root body is
+named `map`. The cells below verify their own results with `assert`, so running this guide as a
```
ichumuh: "the cells below..." sentence sounds like a claude thought

semantic_digital_twin/src/semantic_digital_twin/api/specifications.py
```diff
@@ -0,0 +1,1067 @@
...
+                child=child,
+                name=self._resolved_name(name),
+                parent_T_connection_expression=parent_T_connection,
+                **self._create_with_dofs_kwargs(),
+            )
+            world.add_connection(connection)
+        return connection
+
+
+@dataclass
```
AbdelrhmanBassiouny: why do you have to make these empty classes?

semantic_digital_twin/src/semantic_digital_twin/predetermined_maps/kitchen_environment.py
```diff
@@ -324,7 +324,7 @@ def _build_environment_furniture(self, world: World):
 
             trash_can = TrashCan.create_with_new_body_in_world(
                 world=world,
-                name=PrefixedName("trash_can"),
+                name="trash_can",
```
AbdelrhmanBassiouny: Why are you getting rid of prefixed names?

semantic_digital_twin/src/semantic_digital_twin/world_description/geometry.py
```diff
@@ -372,6 +379,33 @@ def copy_for_world(self, world: World) -> Self:
         }
         return self.__class__(origin=new_origin, **new_props)
 
+    def as_shape_collection(self) -> ShapeCollection:
+        """
+        Wraps this shape in a single-element shape collection anchored to its reference frame.
+        """
+        from semantic_digital_twin.world_description.shape_collection import (
```
AbdelrhmanBassiouny, line 446: cyclic import?

test/semantic_digital_twin_test/test_specifications.py
```diff
@@ -0,0 +1,1139 @@
...
+        body_specification=BodySpecification.box("drawer", Scale(1, 1, 1)),
+        connection_specification=PrismaticConnectionSpecification(axis=Vector3.Z()),
+    )
+    body = spec.spawn(empty_world)
+    assert isinstance(body.parent_connection, PrismaticConnection)
+
+
+def test_child_specification_recursion(empty_world):
+    parent_spec = BodySpecification.box("parent", Scale(1, 1, 1))
+    parent_spec.child_specification.append(
```
AbdelrhmanBassiouny: you added a child without needing to specify a connection?

test/semantic_digital_twin_test/test_specifications.py
```diff
@@ -0,0 +1,1139 @@
...
+@pytest.fixture
+def empty_world() -> World:
+    world = World()
+    with world.modify_world():
+        world.add_body(Body(name=PrefixedName("root", "world")))
+    return world
+
+
+def test_body_specification_spawns_fixed(empty_world):
+    body = BodySpecification.box("box", Scale(1, 1, 1)).spawn(empty_world)
```
AbdelrhmanBassiouny, line 106: I like

test/semantic_digital_twin_test/test_specifications.py
```diff
@@ -0,0 +1,1139 @@
...
+    parent_spec.child_specification.append(
+        BodySpecification.box("child", Scale(1, 1, 1))
+    )
+    parent_body = parent_spec.spawn(empty_world)
+    child = empty_world.get_body_by_name("child")
+    assert child.parent_connection.parent is parent_body
+
+
+def test_fixed_annotation_spawns(empty_world):
+    spec = SemanticAnnotationWithRootSpecification(
```
AbdelrhmanBassiouny: nice

test/semantic_digital_twin_test/test_specifications.py
```diff
@@ -0,0 +1,1139 @@
...
+    assert annotation in empty_world.semantic_annotations
+    assert isinstance(annotation.root.parent_connection, FixedConnection)
+
+
+def test_active_annotation_spawns(empty_world):
+    spec = SemanticAnnotationWithRootSpecification(
+        name="slider",
+        semantic_annotation_type=Slider,
+        root_specification=BodySpecification.box("slider", Scale(0.1, 0.1, 0.1)),
+        axis=Vector3.Z(),
```
AbdelrhmanBassiouny: Does axis appear in the hints of the IDE as an argument that i can or need to set, or do I have to figure that out depending on the semantic annotation type?

test/semantic_digital_twin_test/test_specifications.py
```diff
@@ -0,0 +1,1139 @@
...
+    "..",
+    "semantic_digital_twin",
+    "resources",
+)
+
+
+def test_world_specification_robotless(empty_world):
+    world = WorldSpecification(
+        world=empty_world,
+        starting_objects=[BodySpecification.box("obj", Scale(1, 1, 1))],
```
AbdelrhmanBassiouny: I think if you just call it objects it's fine

AbdelrhmanBassiouny (changes requested):
nice work

semantic_digital_twin/src/semantic_digital_twin/api/specifications.py
LucaKro: they are the default case, both are parameterized via their generic. the rest works in the background. only active connections need special handling

semantic_digital_twin/src/semantic_digital_twin/predetermined_maps/kitchen_environment.py
LucaKro: its just as input for the factories, otherwise it needs to be union str and prefixedname, and most people dont use the prefix anyways

semantic_digital_twin/src/semantic_digital_twin/world_description/geometry.py
LucaKro: yep because shapecollection imports boundingbox from this file

test/semantic_digital_twin_test/test_specifications.py
LucaKro: if you dont specify a connection, it defaults to fixedconnection. but good point actually, i dont think connectionbodyspecification fits the signature yet. ill fix that

test/semantic_digital_twin_test/test_specifications.py
LucaKro: its the default arguments for semanticannotationspecification. rn if its a fixedconnection, the arguments just dont do anything. im not sure if im 100% happy with it yet

test/semantic_digital_twin_test/test_specifications.py
LucaKro: okay i removed connectedbodyspecification

semantic_digital_twin/examples/building_worlds_with_specifications.md
```diff
@@ -0,0 +1,494 @@
...
+by default.
+
+```{code-cell} ipython3
+from semantic_digital_twin.api.specifications import BodySpecification
+from semantic_digital_twin.world_description.geometry import Scale, Color
+
+world = World.create_with_root_body()
+
+table_top = BodySpecification.box(
+    "table_top", Scale(1.2, 0.8, 0.05), color=Color(0.6, 0.4, 0.2, 1.0)
```
ichumuh: write arg names

semantic_digital_twin/examples/building_worlds_with_specifications.md
```diff
@@ -0,0 +1,494 @@
...
+```{code-cell} ipython3
+from semantic_digital_twin.api.specifications import SemanticAnnotationWithRootSpecification
+from semantic_digital_twin.semantic_annotations.semantic_annotations import Milk
+
+world = World.create_with_root_body()
+
+milk = SemanticAnnotationWithRootSpecification(
+    name="milk",
+    semantic_annotation_type=Milk,
+    root_specification=BodySpecification.box("milk", Scale(0.1, 0.1, 0.2)),
```
ichumuh: explicit pose in example

semantic_digital_twin/examples/building_worlds_with_specifications.md
```diff
@@ -0,0 +1,494 @@
...
+assert world is not another_world
+assert len(another_world.get_semantic_annotations_by_type(Milk)) == 1
+print("Materialized two independent worlds, each with one milk and one cup")
+```
+
+### Adding a robot
+
+A world specification can also merge a robot into the environment. Supply the robot's
+[semantic annotation class](adding-robots) through `robot_semantic_annotation`; the robot is
+parsed from its own description and inserted as `world.root -> odom_combined -> drive -> robot`.
```
ichumuh: odom is a better generic name, or 'odometry'

semantic_digital_twin/src/semantic_digital_twin/api/specifications.py
```diff
@@ -0,0 +1,1067 @@
...
+    (incompatible) verbs from it without one masquerading as the other.
+    """
+
+    name: Optional[str]
+    """
+    The name of entities created from this specification, as a plain string. ``None`` defers naming to 
+    materialization.
+    """
+
+    def _resolved_name(self, name: Optional[str] = None) -> Optional[PrefixedName]:
```
ichumuh: luca schreibt fix richtig

semantic_digital_twin/src/semantic_digital_twin/api/specifications.py
```diff
@@ -0,0 +1,1067 @@
...
+    The concrete domain-object type (e.g. ``Body``/``Region``) is bound as the generic
+    parameter by each subclass and resolved at runtime in :meth:`to_domain_object`.
+    """
+
+    shapes: ShapeCollection = field(default_factory=ShapeCollection)
+    """
+    Prototype shapes with origins expressed in the entity frame.
+    """
+
+    child_specification: list[KinematicStructureEntitySpecification] = field(
```
ichumuh: s

semantic_digital_twin/src/semantic_digital_twin/api/specifications.py
```diff
@@ -0,0 +1,1058 @@
...
+            connection_specification=connection_specification,
+        )
+
+    @classmethod
+    def from_3d_points(
+        cls,
+        name: str,
+        points_3d: List[Point3],
+        minimum_thickness: float = 0.005,
+        sv_ratio_tol: float = 1e-7,
```
ichumuh: Scientifically Vibes-based Ratio Tolerance?

semantic_digital_twin/src/semantic_digital_twin/api/specifications.py
```diff
@@ -0,0 +1,1058 @@
...
+    World-independent description of a :class:`~semantic_digital_twin.world_description.world_entity.Region`.
+
+    Carries no fields beyond the base kinematic-structure-entity specification; it only binds the
+    materialized domain-object type to :class:`Region`.
+    """
+
+
+@dataclass
+class SemanticAnnotationWithRootSpecification(
+    SpawnSpecification["HasRootKinematicStructureEntity"]
```
ichumuh: str?

semantic_digital_twin/src/semantic_digital_twin/api/specifications.py
```diff
@@ -0,0 +1,1058 @@
...
+    """
+    The environment world that the robot and starting objects are added to. Its root is ``world.root``.
+    """
+
+    robot_semantic_annotation: Optional[Type[AbstractRobot]] = None
+    """
+    The robot to merge into the environment. If None, no robot is added.
+    """
+
+    drive_connection_type: Type[WheeledDrive] | None = None
```
ichumuh: belongs to robot

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
```diff
@@ -908,28 +960,20 @@ def create_with_new_body_in_world(
...
-            body, container_event
-        ).as_shapes()
-        body.collision = collision_shapes
-        body.visual = collision_shapes
-        return cls._create_with_connection_in_world(
-            name=name,
-            world=world,
-            kinematic_structure_entity=body,
-            world_root_T_self=world_root_T_self,
+        scale = scale if scale is not None else Scale()
```
ichumuh: scale = scale or Scale()

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
```diff
@@ -950,3 +994,25 @@ def _create_container_event(cls, scale: Scale, wall_thickness: float) -> Event:
...
+        """
+        Build the default body specification for a container/case, matching the
+        hollow geometry generated by :meth:`create_with_new_body_in_world`.
+
+        :param name: The name of bodies created from the specification.
+        :param scale: The outer scale of the case. Defaults to a unit :class:`Scale` when omitted.
+        :param wall_thickness: The thickness of the case walls.
+        :return: A body specification with hollow container geometry.
+        """
+        scale = scale if scale is not None else Scale()
```
ichumuh, line 1063: see above

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/semantic_annotations.py
```diff
@@ -77,34 +82,21 @@ class Handle(HasRootBody):
...
-        handle_body = Body(name=name)
-        collision = BoundingBoxCollection.from_event(
-            handle_body, handle_event
-        ).as_shapes()
-        handle_body.collision = collision
-        handle_body.visual = collision
-        return cls._create_with_connection_in_world(
-            name, world, handle_body, world_root_T_self
-        )
+        scale = scale if scale is not None else Scale(0.1, 0.02, 0.02)
```
ichumuh: see above

semantic_digital_twin/src/semantic_digital_twin/world_description/geometry.py
```diff
@@ -372,6 +379,33 @@ def copy_for_world(self, world: World) -> Self:
...
+        Moves the origin so the shape's local-frame bounding box is centered on it.
+
+        The translation is set to the negated bounding-box center, leaving the shape's
+        geometry symmetric about its origin.
+        """
+        bounding_box = self.local_frame_bounding_box
+        center_x = (bounding_box.min_x + bounding_box.max_x) / 2
+        center_y = (bounding_box.min_y + bounding_box.max_y) / 2
+        center_z = (bounding_box.min_z + bounding_box.max_z) / 2
+        self.origin = HomogeneousTransformationMatrix.from_xyz_rpy(
```
ichumuh: reference_frame=self or body it belongs to?

ichumuh (changes requested):
i prefer this interface over the normal way of creating worlds, but it feels a bit like a DRY violation

LucaKro:
> i prefer this interface over the normal way of creating worlds, but it feels a bit like a DRY violation

I agree, im not 100% happy yet, but i think im at the stage where feedback from the outside will be very valuable.
Also im not sure how to solve the DRY violation, any ideas are welcome. Im on vacation until the 14th July now, so im probably not going to answer until then

semantic_digital_twin/src/semantic_digital_twin/world_description/geometry.py
```diff
@@ -372,6 +379,33 @@ def copy_for_world(self, world: World) -> Self:
...
+            -center_x, -center_y, -center_z, 0, 0, 0
```
Narenvasant: why is the number hard coded here? if there is existing rotation, will it might be recentered to wrong orientation

semantic_digital_twin/src/semantic_digital_twin/exceptions.py
```diff
@@ -335,6 +335,60 @@ def suggest_correction(self) -> str:
...
+
+@dataclass
+class MissingConnectionParentError(UsageError):
+    """
+    Raised when a connection is spawned without a parent kinematic structure entity.
+    """
+
+    connection_name: Optional[str]
+    """
+    The name of the connection specification that was spawned without a child.
```
Narenvasant: 'parent' instead of 'child'?

Narenvasant (changes requested):

semantic_digital_twin/src/semantic_digital_twin/world_description/world_entity.py
```diff
@@ -474,9 +461,17 @@ def __post_init__(self):
 
     @classmethod
     def from_shape_collection(
-        cls, name: PrefixedName, shape_collection: ShapeCollection
+        cls,
+        name: PrefixedName,
+        shape_collection: ShapeCollection,
+        *,
+        visuals_shape_collection: ShapeCollection = None,
```
MarvinDufner: optional

semantic_digital_twin/src/semantic_digital_twin/world_description/connections.py
```diff
@@ -459,11 +464,14 @@ def create_with_dofs(
             HomogeneousTransformationMatrix
         ] = None,
         connection_T_child_expression: Optional[HomogeneousTransformationMatrix] = None,
+        **kwargs,
     ) -> Self:
         """
         Creates an instance of the class with automatically generated degrees of freedom (DoFs)
         for the provided parent and child kinematic entities within the specified world.
 
+        The additional kwargs parameters are accepted for interface consistency with other connections and ignored here.
```
MarvinDufner: interface consistency is cool and all but doesnt this now just silently drop any TypeError that would have been raised for unexpected keyword arguments?

MarvinDufner (changes requested):
good work 🥇

semantic_digital_twin/examples/building_worlds_with_specifications.md
LucaKro: yea ill reformulate it. the problem is that connections how i think about them rn are not actually entities (even though they are world entities) but rather relations. but changing that terminology in semdt now would cause confusion

semantic_digital_twin/examples/building_worlds_with_specifications.md
LucaKro: elaborate please

semantic_digital_twin/examples/building_worlds_with_specifications.md
LucaKro: its not, but i see what you mean. its not really relevant, so i removed it again

semantic_digital_twin/src/semantic_digital_twin/api/specifications.py
LucaKro: circular imports otherwise (specifications and mixins, since i provide default specification configurations in mixins, and obv specifications need to import mixins).
The alternative would be local imports everywhere in mixins.py, or no nice default configs for semantic annotations

semantic_digital_twin/src/semantic_digital_twin/robots/robot_parts.py
```diff
@@ -586,6 +627,21 @@ def _get_root_body_name(cls) -> str:
...
+        """
+        The connection type attaching this robot to its ``odom`` frame.
+
+        :return: The mobile base's drive connection type, or :class:`FixedConnection`
+            when the robot has no mobile base and is therefore rigidly attached.
+        """
+        if not issubclass(cls, HasMobileBase):
+            return FixedConnection
+        mobile_base_type = cast(
+            MobileBase, get_generic_type_parameters(cls, HasMobileBase)[0]
```
ichumuh, line 682: what is this cast for? chatgpt says it does nothing in code and is only used by type checkers

semantic_digital_twin/src/semantic_digital_twin/exceptions.py
```diff
@@ -1342,3 +1380,116 @@ def error_message(self) -> str:
...
+    :class:`~semantic_digital_twin.adapters.mujoco_video_recording.MujocoVideoRecorder`
+    exceptions.
+    """
+
+
+@dataclass
+class VideoRecordingAlreadyStartedError(VideoRecordingError):
+    """
+    Raised when :meth:`~semantic_digital_twin.adapters.mujoco_video_recording.MujocoVide
+    oRecorder.start` is called on a recorder that is already recording.
```
ichumuh: formatting fail?

semantic_digital_twin/src/semantic_digital_twin/exceptions.py
```diff
@@ -1342,3 +1380,116 @@ def error_message(self) -> str:
...
+
+    def suggest_correction(self) -> str:
+        return "call stop() before starting a new recording."
+
+
+@dataclass
+class VideoRecordingNotStartedError(VideoRecordingError):
+    """
+    Raised when :meth:`~semantic_digital_twin.adapters.mujoco_video_recording.MujocoVide
+    oRecorder.stop` is called on a recorder that was never started.
```
ichumuh: see above

ichumuh (changes requested):

semantic_digital_twin/src/semantic_digital_twin/robots/robot_parts.py
LucaKro: yes its basically the correct way to tell the type checker that you know it is a specific type.

its basically equivalent to:
```python
mobile_base_type: MobileBase = get_generic_type_parameters(cls, HasMobileBase)[0] 
```

except that in this version, it would underline `get_generic_type_parameters(cls, HasMobileBase)[0] ` yellow, and with cast it doesnt

semantic_digital_twin/src/semantic_digital_twin/exceptions.py
LucaKro: no i dont think so, the docstring looks fine to me, i think it was just a whitespace introduced by docformatter

semantic_digital_twin/src/semantic_digital_twin/exceptions.py
LucaKro: see above

.github/workflows/build_all_docs.yml
Tigul: 👀

semantic_digital_twin/src/semantic_digital_twin/exceptions.py
LucaKro: fixed

semantic_digital_twin/src/semantic_digital_twin/exceptions.py
LucaKro: fixed

krrood/src/krrood/utils.py
```diff
@@ -594,7 +597,7 @@ class needing it, and once per lookup attempt). Every attempt raises the exact s
     :param file_path: The path of the file whose imports were being extracted.
     :param error_message: The message of the ``AttributeError`` that was raised.
     """
-    logger.warning(
+    logger.debug(
```
Tigul, line 600: 🙏

semantic_digital_twin/examples/building_worlds_with_specifications.md
```diff
@@ -0,0 +1,523 @@
...
+
+from semantic_digital_twin.world import World
+from semantic_digital_twin.spatial_computations.raytracer import RayTracer
+from semantic_digital_twin.world_description.connections import (
+    FixedConnection,
+    PrismaticConnection,
+)
+
+world = World.create_with_root_body()
+assert world.root.name.name == "map"
```
Tigul, line 76: grrrrr asserts

semantic_digital_twin/examples/semantic_annotation_part_whole.md
```diff
@@ -45,43 +45,39 @@ Let's build a dresser whose drawer has a handle and a slider. We use the factori
...
-world = World()
-root = Body(name=PrefixedName("root"))
-
-with world.modify_world():
-    world.add_body(root)
+world = World.create_with_root_body()
 
 with world.modify_world():
     dresser = Dresser.create_with_new_body_in_world(
-        name=PrefixedName("dresser"),
```
Tigul, line 64: If you do not remove PrefixedName completly keep this consistent and use it here as well

semantic_digital_twin/src/semantic_digital_twin/adapters/procthor/procthor_parser.py
```diff
@@ -196,7 +192,7 @@ def _add_double_door_to_world(self, world: World) -> DoubleDoor:
             doors.append(door)
 
         double_door = DoubleDoor(
-            name=self.name,
+            name=PrefixedName(self.name),
```
Tigul, line 195: 👀

semantic_digital_twin/src/semantic_digital_twin/adapters/procthor/procthor_parser.py
```diff
@@ -499,24 +495,25 @@ def add_to_world(self, world: World):
...
             )
-        if "Bedroom" in self.name.name:
-            room = Bedroom(name=self.name, floor=floor)
-        elif "LivingRoom" in self.name.name:
-            room = LivingRoom(name=self.name, floor=floor)
-        elif "Kitchen" in self.name.name:
-            room = Kitchen(name=self.name, floor=floor)
-        elif "Bathroom" in self.name.name:
-            room = Bathroom(name=self.name, floor=floor)
+        room_name = PrefixedName(self.name)
```
Tigul, line 506: 👀

semantic_digital_twin/src/semantic_digital_twin/adapters/sage_10k_dataset/schema.py
```diff
@@ -308,7 +306,7 @@ def create_in_world(self, world: World, directory: Path, parent: Body) -> Wall:
 
         with world.modify_world():
             annotation = Wall.create_with_new_body_in_world(
-                name=wall_name,
+                name=self.id,
```
Tigul, line 309: Please no, use something readable for the name

Tigul: Also for the rest of the file

semantic_digital_twin/src/semantic_digital_twin/api.py
```diff
@@ -0,0 +1,1285 @@
...
+class NamedSpecification(ABC):
+    """
+    Base for every specification: it carries a name and normalizes it.
+
+    It deliberately declares no materialization contract, so entity-spawn specs and
+    connection specs can derive their own (incompatible) verbs from it without one
+    masquerading as the other.
+    """
+
+    name: str | None
```
Tigul, line 108: Optional[str]

semantic_digital_twin/src/semantic_digital_twin/api.py
```diff
@@ -0,0 +1,1285 @@
...
+    """
+    The name of entities created from this specification, as a plain string.
+
+    ``None`` defers naming to materialization.
+    """
+
+    def _resolved_name(self, name: str | None = None) -> PrefixedName | None:
```
Tigul, line 115: Optional typing

Tigul: Also PrefixedName and str name used here. Please chose one and stick to it

semantic_digital_twin/src/semantic_digital_twin/api.py
```diff
@@ -0,0 +1,1285 @@
...
+    :class:`SpawnSpecification` (which materializes an entity and its own parent connection). It is
+    materialized via :meth:`connect`, which takes the ``child`` to attach.
+
+    Each connection family is a concrete subclass that binds the connection type as its generic
+    parameter (e.g. ``ConnectionSpecification[FixedConnection]``) and carries exactly the parameters
+    that family uses. Materializing a specification forwards those parameters to the connection type's
+    :meth:`~semantic_digital_twin.world_description.world_entity.Connection.create_with_dofs`.
+    """
+
+    name: str | None = field(default=None, kw_only=True)
```
Tigul, line 183: It even says optional in the doc string

semantic_digital_twin/src/semantic_digital_twin/api.py
```diff
@@ -0,0 +1,1285 @@
...
+            Identity if None.
+        :param name: Overrides the specification's own name. If None, the spec's name is
+            used.
+        :return: The materialized connection.
+        :raises MissingConnectionParentError: If no parent is given and the world has no
+            root.
+        """
+        parent = parent or world.root
+        if parent is None:
+            raise MissingConnectionParentError(connection_name=self.name)
```
Tigul, line 248: No Body in world would be more specific here

semantic_digital_twin/src/semantic_digital_twin/api/specifications.py
```diff
@@ -0,0 +1,1285 @@
...
+        :return: The materialized connection.
+        :raises MissingConnectionParentError: If no parent is given and the world has no
+            root.
+        """
+        parent = parent or world.root
+        if parent is None:
+            raise MissingConnectionParentError(connection_name=self.name)
+
+        parent_T_connection = (
+            deepcopy(parent_T_connection)
```
Tigul: shallow copy would be better I guess since we can't deepcopy connections anyway. This way the code is more readable

Tigul: Also be carfule with copy of connections since this has always the potential for memory leaks

semantic_digital_twin/src/semantic_digital_twin/api.py
```diff
@@ -0,0 +1,1285 @@
...
+        parent_T_connection = (
+            deepcopy(parent_T_connection)
+            if parent_T_connection is not None
+            else HomogeneousTransformationMatrix()
+        )
+        parent_T_connection.reference_frame = parent
+        parent_T_connection.child_frame = child
+
+        with world.modify_world():
+            connection = self.connection_type.create_with_dofs(
```
Tigul, line 261: Why deepcopy when you create a new connection ?

semantic_digital_twin/src/semantic_digital_twin/api.py
```diff
@@ -0,0 +1,1285 @@
...
+            self._resolved_name(name),
+            self.shapes.copy_without_reference_frame(),
+        )
+
+    def attach_and_spawn_children(
+        self,
+        world: World,
+        entity: KinematicStructureEntity,
+        connection_specification: ConnectionSpecification,
+        parent: KinematicStructureEntity | None = None,
```
Tigul, line 458: optional

semantic_digital_twin/src/semantic_digital_twin/api/specifications.py
```diff
@@ -0,0 +1,1285 @@
...
+        :param connection_specification: How the entity attaches to its parent. Defaults
+            to a fixed connection.
+        :return: The created specification.
+        """
+        return cls(
+            name,
+            Sphere(
+                radius=radius,
+                origin=(
+                    origin if origin is not None else HomogeneousTransformationMatrix()
```
Tigul: An empty matrix will have None as reference frame, which could cause problems later on if something is working with the object

semantic_digital_twin/src/semantic_digital_twin/api.py
```diff
@@ -0,0 +1,1285 @@
...
+            Defaults to identity.
+        :param child_specifications: Specifications spawned as kinematic children of the
+            entity. Defaults to none.
+        :param connection_specification: How the entity attaches to its parent. Defaults
+            to a fixed connection.
+        :return: The created specification.
+        """
+        # BoundingBoxCollection requires a reference frame, so the shapes are
+        # built around a throwaway body and unbound again for the specification.
+        anchor = Body(name=PrefixedName("spec_anchor"))
```
Tigul, line 729: The sounds hacky and as if there is a bigger architectural problem here

semantic_digital_twin/src/semantic_digital_twin/api.py
```diff
@@ -0,0 +1,1285 @@
...
+    from semantic_digital_twin.semantic_annotations.mixins import (
+        HasRootKinematicStructureEntity,
+        PartWholeRelationship,
+    )
+    from semantic_digital_twin.robots.robot_parts import AbstractRobot
+    from semantic_digital_twin.adapters.package_resolver import PathResolver
+
+
+# %% specification type parameters
+TWorldEntity = TypeVar("TWorldEntity", bound=WorldEntity)
```
Tigul, line 85: Why use these and not the type directly?

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
```diff
@@ -297,69 +377,58 @@ class HasRootRegion(HasRootKinematicStructureEntity[TRegion], ABC):
...
+
+        :param name: The name of regions created from the specification. ``None`` leaves
+            naming to the spawning annotation, which overrides it with its own name.
+        :param scale: The scale used to generate the region area geometry. ``None`` yields a
+            geometry-less specification, matching the bare region of the base factory.
+        :param connection_specification: Connection attaching the entity to its parent.
+            ``None`` defers to the annotation type's default at spawn time.
+        :return: A region specification.
+        """
+        if scale is None:
```
Tigul, line 421: You can short this to one return statement with scale=scale.to_simple_event().as_composite_set() if scale else None

semantic_digital_twin/src/semantic_digital_twin/world_description/shape_collection.py
```diff
@@ -106,11 +118,17 @@ def __contains__(self, shape: Shape) -> bool:
 
     def append(self, shape: Shape):
         if self.world is not None:
-            self._transform_to_own_frame(
-                shape,
-            )
+            self._transform_shape_to_reference_frame(shape)
         self.shapes.append(shape)
 
+    def copy_without_reference_frame(self) -> ShapeCollection:
```
Tigul, line 124: Hmmm, that sounds like it will make problems down the line

semantic_digital_twin/src/semantic_digital_twin/world.py
```diff
@@ -547,6 +545,20 @@ def __post_init__(self):
         )
         self.collision_manager.add_to_world(self)
 
+    @classmethod
+    def create_with_root_body(
```
Tigul: I like

Tigul (changes requested):
If you start removing PrefixedName then finish it please and remove it completely otherwise we have a weird interface where some things have a prefixed name and some don't

semantic_digital_twin/examples/semantic_annotation_part_whole.md
LucaKro: as discussed during lunch, i disagree. right now specifications and create_with_new_body_in_world receive strings, because they are the userinterfaces, and all three users i reviewed who dont work with semdt that much created their own interfaces for interacting in the world, which also only used strings. this is actually what spawned the PR, because i thoroughly reviewed the three implementations to find out the requirements users seem to have that we dont easily fulfill atm.

semantic_digital_twin/src/semantic_digital_twin/adapters/procthor/procthor_parser.py
LucaKro: no thats correct. if you create a semantic annotation by hand like that, it still holds a prefixed name. just the userinterfaces mentioned above take strings and convert to prefixed names inside

semantic_digital_twin/src/semantic_digital_twin/adapters/procthor/procthor_parser.py
LucaKro: same as above

semantic_digital_twin/src/semantic_digital_twin/adapters/sage_10k_dataset/schema.py
LucaKro: the sage10k id is a human readable string

semantic_digital_twin/src/semantic_digital_twin/api.py
LucaKro: https://docs.python.org/3/library/stdtypes.html#types-union explicitly mentiones this notation as a valid notation for optional. also its used in other cases in the code base as well

semantic_digital_twin/src/semantic_digital_twin/api.py
LucaKro: its str, outputting prefixedname

semantic_digital_twin/src/semantic_digital_twin/api.py
LucaKro: this is officially equivalent to optional as linked above

semantic_digital_twin/src/semantic_digital_twin/api.py
LucaKro: actually i thought about that, and i disagree. I think the more helpful, less generic error is the current one, but ill adjust the error message to reflect that the world must be empty

semantic_digital_twin/src/semantic_digital_twin/api/specifications.py
LucaKro: I added copy_with_new_reference_frames method and used it instead

semantic_digital_twin/src/semantic_digital_twin/api.py
LucaKro: the copy is only on the transform, not the connection

semantic_digital_twin/src/semantic_digital_twin/api/specifications.py
LucaKro: reference frames for shapes are automatically set when a body is created

semantic_digital_twin/src/semantic_digital_twin/api.py
LucaKro: hmm i see what you mean, but i think its not something for this PR. in this case it wont caus any problems becaues its only used once for creation and the resulting shape collection is copied without any reference frames

semantic_digital_twin/src/semantic_digital_twin/api.py
LucaKro: because it together with SubclassSafeGeneric allows me to abstract some interfaces by depending on the Generic Type

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
LucaKro: actually no, one return uses from_event, which is not optional, while the other one constructs it normally

semantic_digital_twin/src/semantic_digital_twin/world_description/shape_collection.py
LucaKro: as discussed at lunch, its fine because reference frames of shapes are set automatically at body creation

semantic_digital_twin/src/semantic_digital_twin/world.py
LucaKro: :innocent:

.github/workflows/build_all_docs.yml
```diff
@@ -24,7 +40,7 @@ jobs:
           - lib: semantic_digital_twin
             doc_path: doc
             artifact_name: semantic-digital-twin-doc
-            use_container: true
+            use_container: False
```
AbdelrhmanBassiouny: why?

krrood/src/krrood/class_diagrams/wrapped_field.py
```diff
@@ -166,16 +173,22 @@ def container_type(self) -> Optional[Type]:
...
         return self.is_container and all(
-            behaves_like_a_built_in_class(field_type)
+            behaves_like_a_built_in_type(field_type)
             for field_type in get_args(self.resolved_type)
         )
 
     @cached_property
     def is_optional(self):
+        """
+        Whether the field holds an optional value.
```
AbdelrhmanBassiouny: use `:return:`

semantic_digital_twin/scripts/new_robot_integration_helper.py
```diff
@@ -21,10 +20,7 @@
 ).parse()
 robot_world.visualize_world_structure().show()
 
-world = World()
-root = Body(name=PrefixedName(name="map"))
-with world.modify_world():
-    world.add_body(root)
+world = World.create_with_root_body(PrefixedName(name="map"))
```
AbdelrhmanBassiouny: I like

semantic_digital_twin/scripts/new_robot_integration_helper.py
ichumuh, line 1: ??

semantic_digital_twin/src/semantic_digital_twin/api.py
```diff
@@ -79,85 +79,289 @@
...
 
 
+# %% specification type parameters
 TWorldEntity = TypeVar("TWorldEntity", bound=WorldEntity)
 TKinematicStructureEntity = TypeVar(
     "TKinematicStructureEntity", bound=KinematicStructureEntity
 )
 TConnection = TypeVar("TConnection", bound=Connection)
+TSemanticAnnotation = TypeVar(
+    "TSemanticAnnotation", bound="HasRootKinematicStructureEntity"
```
AbdelrhmanBassiouny, line 91: why is the bound quoted?

semantic_digital_twin/examples/semantic_annotation_part_whole.md
AbdelrhmanBassiouny: I think it should be completely removed, but necessary to make it in this PR if it touches much more files than what this PR modifies.

semantic_digital_twin/src/semantic_digital_twin/api/specifications.py
```diff
@@ -514,11 +549,19 @@ def box(
...
             ).as_shape_collection(),
-            child_specifications=child_specifications or [],
-            parent_T_self=parent_T_self or HomogeneousTransformationMatrix(),
+            child_specifications=(
+                child_specifications if child_specifications is not None else []
+            ),
+            parent_T_self=(
+                parent_T_self
+                if parent_T_self is not None
+                else HomogeneousTransformationMatrix()
```
ichumuh: or

semantic_digital_twin/src/semantic_digital_twin/api.py
```diff
@@ -168,138 +372,244 @@ class KinematicStructureEntitySpecification(
...
     @classmethod
     def sphere(
         cls,
         name: str,
         radius: float,
-        color: Optional[Color] = None,
-        origin: Optional[HomogeneousTransformationMatrix] = None,
-        parent_T_self: Optional[HomogeneousTransformationMatrix] = None,
-        child_specification: list[KinematicStructureEntitySpecification] | None = None,
+        color: Color | None = None,
```
AbdelrhmanBassiouny, line 583: Starting from color to connection_specification, you have the exact same arguments with box, cylinder, and mesh. For ease of maintenance, you can make a dataclass that holds these and reduce the number of arguments to just one that covers all of these. You could call it maybe `CommonShapeData`.

semantic_digital_twin/src/semantic_digital_twin/api.py
```diff
@@ -665,403 +1068,237 @@ def _mount_part_specifications(
...
-    Each connection family is a concrete subclass that binds the connection type as its generic
-    parameter (e.g. ``ConnectionSpecification[FixedConnection]``) and carries exactly the parameters
-    that family uses. Materializing a specification forwards those parameters to the connection type's
-    :meth:`~semantic_digital_twin.world_description.world_entity.Connection.create_with_dofs`.
+    Materialized via :meth:`spawn`, which merges the robot as ``world.root -> odom ->
+    drive -> robot``.
     """
 
-    name: Optional[str] = field(default=None, kw_only=True)
+    semantic_annotation_type: Type[AbstractRobot]
```
AbdelrhmanBassiouny, line 1071: Not for this PR, but when `Robot`??? I guess this will conflict with Michael's perfect ontology, haha.

semantic_digital_twin/src/semantic_digital_twin/api.py
ichumuh, line 1: rename to api.py and remove the package

semantic_digital_twin/src/semantic_digital_twin/api.py
```diff
@@ -665,403 +1068,237 @@ def _mount_part_specifications(
...
         file_path: str,
         *,
-        prefix: Optional[str] = None,
-        path_resolver: Optional[PathResolver] = None,
-        robot_semantic_annotation: Optional[Type[AbstractRobot]] = None,
-        drive_connection_type: Optional[Type[WheeledDrive]] = None,
-        world_T_odom: Optional[HomogeneousTransformationMatrix] = None,
-        odom_T_robot_start: Optional[HomogeneousTransformationMatrix] = None,
-        starting_objects: Optional[list[SpawnSpecification]] = None,
+        prefix: str | None = None,
```
AbdelrhmanBassiouny, line 1199: same comment on repeated arguments `from_urdf` and `from_mjcf` share prefix, robots, and objects. They can be in for example `CommonParsedEvironmentData`?

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
```diff
@@ -249,10 +297,12 @@ def _kinematic_structure_entities(
 @dataclass(eq=False)
 class HasRootBody(HasRootKinematicStructureEntity[TBody], ABC):
     """
-    Abstract base class for all household objects. Each semantic annotation refers to a single Body.
-    Each subclass automatically derives a MatchRule from its own class name and
-    the names of its HouseholdObject ancestors. This makes specialized subclasses
-    naturally more specific than their bases.
+    Abstract base class for all household objects.
```
ichumuh: ????

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
```diff
@@ -270,22 +320,22 @@ def create_with_new_body_in_world(
...
+            cls.get_default_root_specification(scale=scale),
             parent_connection_specification=parent_connection_specification,
         ).spawn(world, parent_T_self=world_root_T_self)
 
     @classmethod
-    def get_default_body_specification(
+    def get_default_root_specification(
         cls,
-        name: str,
+        name: Optional[str] = None,
```
ichumuh: | None

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
```diff
@@ -343,30 +388,32 @@ def create_with_new_region_in_world(
...
+            cls.get_default_root_specification(scale=scale),
             parent_connection_specification=parent_connection_specification,
         ).spawn(world, parent_T_self=world_root_T_self)
 
     @classmethod
-    def get_default_region_specification(
+    def get_default_root_specification(
         cls,
-        name: str,
+        name: Optional[str] = None,
```
ichumuh: | None

semantic_digital_twin/examples/semantic_annotation_part_whole.md
ichumuh: I think users shouldn't have to use PrefixedName, just because removing it is out of scope for this PR. Make the interface that users want, which is str!!!

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/part_whole.py
```diff
@@ -0,0 +1,79 @@
...
+
+if TYPE_CHECKING:
+    from semantic_digital_twin.semantic_annotations.mixins import PartWholeRelationship
+
+
+# %% relation metadata
+
+
+@dataclass
+class IsPartWholeRelationship(FieldMetadata):
```
AbdelrhmanBassiouny, line 19: Just want to argue about FieldMetadata

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/part_whole.py
```diff
@@ -0,0 +1,79 @@
...
+    """
+
+
+# %% field discovery
+
+
+@lru_cache(maxsize=None)
+def wrapped_part_whole_relationship_fields(
+    cls: Type[PartWholeRelationship],
+) -> list[WrappedField]:
```
ichumuh: shouldn't this be a method of FieldMetadata? also lru_cache is evil

ichumuh (changes requested):

semantic_digital_twin/scripts/new_robot_integration_helper.py
LucaKro: nvm wasnt introduced in this PR

test/semantic_digital_twin_test/test_semantic_annotations/test_factories.py
```diff
@@ -72,21 +77,21 @@
...
+from semantic_digital_twin.api.specifications import (
+    SemanticAnnotationWithRootSpecification,
+)
 
 
 class TestFactories(unittest.TestCase):
     def test_handle_factory(self):
-        world = World()
-        root = Body(name=PrefixedName("root"))
+        world = World.create_with_root_body(PrefixedName("root"))
```
AbdelrhmanBassiouny: Isn't this method part of what this PR touches? If so, why use PrefixedName?

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/part_whole.py
LucaKro: the cache is only considering the type, so i htink its fine. yea maybe it could be a method of fieldmetadata

test/semantic_digital_twin_test/test_semantic_annotations/test_factories.py
```diff
@@ -1103,53 +1014,99 @@ def test_characterize_aperture_region_geometry(self):
...
+                name="microwave_door",
+                scale=Scale(0.03, 0.3, 0.3),
+                world=world,
+            )
+            microwave.add(door)
+
+        semantic_microwave_annotations = world.get_semantic_annotations_by_type(
+            Microwave
+        )
+        self.assertEqual(len(semantic_microwave_annotations), 1)
```
AbdelrhmanBassiouny, line 1087: isn't this a unittest method? vs the pytest way of using Python's built-in assert.

test/semantic_digital_twin_test/test_specifications.py
```diff
@@ -1130,10 +1257,318 @@ def test_nested_composite_matches_manual_construction(empty_world):
...
+    # type and the part-whole fields the parts were bound to come back unchanged.
+    reconstructed = to_dao(_drawer_specification_with_handle()).from_dao()
+
+    assert reconstructed.semantic_annotation_type is Drawer
+    assert [binding.field_name for binding in reconstructed.part_bindings] == ["handle"]
+
+
+@pytest.mark.xfail(
+    strict=True,
+    reason="ormatic maps no specification-valued field, so a persisted specification "
```
AbdelrhmanBassiouny, line 1634: Do you want to fix this in this PR?

AbdelrhmanBassiouny (changes requested):

.github/workflows/build_all_docs.yml
LucaKro: for building the docs on my fork, will be reverted before merge

semantic_digital_twin/src/semantic_digital_twin/api.py
LucaKro: Lazy evaluation to avoid cyclic imports

semantic_digital_twin/src/semantic_digital_twin/api.py
LucaKro: haha sure, actually i think i chose AbstractRobot back then because he was always saying "abstractrobot" xD but yea changing it to robot probably makes sense in the future

semantic_digital_twin/src/semantic_digital_twin/api.py
LucaKro: actually im not sure about that. again the goal of these interfaces is to be as little annoying for users as possible, and so far people always seem to want to use as little complex datastructures as input as possible (see PrefixedName -> str, which is literally just a wrapper around str and people still didnt use it)

semantic_digital_twin/src/semantic_digital_twin/api.py
LucaKro: see my comment below where you mentioned sth similar

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/part_whole.py
LucaKro: xDDD

test/semantic_digital_twin_test/test_semantic_annotations/test_factories.py
LucaKro: you are right

test/semantic_digital_twin_test/test_semantic_annotations/test_factories.py
LucaKro: yes, and these tests are in a unittest class. i guess one can rewrite the whole file, but out of scope for this PR

test/semantic_digital_twin_test/test_specifications.py
LucaKro: im not sure how, and if we even need it. tbh i call yagni, why would we persist a userinterface. in the end only the resulting modifications are whats of interest to us

AbdelrhmanBassiouny: maybe yagni for now

krrood/src/krrood/class_diagrams/class_diagram.py
```diff
@@ -274,6 +276,38 @@ class WrappedClass(Generic[T], SubClassSafeGeneric):
...
+    _fields_by_metadata_type_: Dict[Type[FieldMetadata], List[WrappedField]] = field(
+        init=False, hash=False, compare=False, default_factory=dict, repr=False
+    )
+    """
+    A mapping from a field metadata type to the fields carrying metadata of that type.
+    """
+
+    @classmethod
+    @lru_cache(maxsize=None)
+    def of(cls, clazz: Type[T]) -> Self:
```
AbdelrhmanBassiouny: This is bad because now two different class diagrams would have same instance with same index of graph.

AbdelrhmanBassiouny (changes requested):

krrood/src/krrood/class_diagrams/class_diagram.py
LucaKro: removed, including the dict cache attribute added

AbdelrhmanBassiouny (approved):

semantic_digital_twin/src/semantic_digital_twin/api.py
Tigul: I disagree there is only one case in which this error can be thrown, I think if there is only one way to get here the error should reflect that.

Tigul (approved):

ichumuh (approved):

PR #407, opened 2026-07-02
Fix old motion statechart tasks

ichumuh:
Summary
-------
Revives and hardens several motion statechart tasks/monitors that had bit-rotted or
lacked test coverage (GraspBar, WiggleInsert, TickCounter, Cartesian and feature
monitors), while fixing a handful of correctness bugs uncovered along the way in
node lifecycle management, condition scoping, and QP data filtering.

Key changes
-----------

Node lifecycle / condition scoping
- MotionStatechartNode now creates its observation/life-cycle variables at
  construction (_node_id via uuid.uuid4()) instead of waiting for
  _post_add_to_motion_statechart, so variables exist and are unique before the
  node is added to a statechart.
- Added ConditionScopeError: conditions may now only reference the owning node
  or its siblings; MotionStatechart._validate_condition_scopes() enforces this
  on construction/deserialization.
- TrinaryCondition.__str__ / AST parsing now resolve observation variables via
  each node's unique_name rather than string-munging the ".../observation"
  suffix.
- ThreadPayloadMonitor now logs (instead of silently swallowing) exceptions
  raised while computing an observation.

Tasks
- Removed AlignToPushDoor - dead code referencing an undefined `context`
  (never worked).
- Reworked GraspBar and WiggleInsert implementations for correctness and
  clarity, each backed by new unit tests.
- Refactored TickCounter subclasses and added unit tests.
- Added a build() method to feature monitors, plus tests for feature and
  Cartesian monitors.
- Minor fixes in cartesian_tasks.py, align_planes.py, pointing.py,
  feature_functions.py.

QP
- Removed the unused retries_with_relaxed_constraints config field.
- Fixed slack-variable index arithmetic in QPDataExplicit.apply_filters and
  QPDataTwoSidedInequality (was slicing with -num_slack_variables, now
  computes the non-slack count explicitly and correctly excludes box
  constraints from the inequality matrix).

ROS2 executor
- DebugExpressionPublisher import moved behind TYPE_CHECKING / deferred to
  point-of-use, since it pulls in rclpy unconditionally and previously made
  giskardpy.ros_executor unimportable without ROS2 installed.

Misc
- test_nodes module renamed to nodes_for_testing.
- ormatic_interface.py files regenerated (generated code, not hand-edited).
- Added detailed docstrings across motion statechart classes/methods.

Testing
-------
New/expanded unit tests: test_grasp_bar.py, test_wiggle_insert.py,
test_cartesian_monitors.py, test_feature_monitors.py, test_qp_data.py,
test_debug_expression_publisher.py, plus updates to test_motion_statechart.py,
test_cartesian_tasks.py, test_json_parsing.py.

All added/changed behavior is covered by new tests per the repo's TDD
convention.


giskardpy/src/giskardpy/motion_statechart/monitors/cartesian_monitors.py
AbdelrhmanBassiouny, line 1: Nice, maybe coraplex should use those instead of what it currently has for pose validation

giskardpy/src/giskardpy/motion_statechart/nodes_for_testing/nodes_for_testing.py
```diff
@@ -49,7 +50,7 @@ def build(self, context: MotionStatechartContext) -> NodeArtifacts:
 
 @dataclass(repr=False, eq=False)
 class ChangeStateOnEvents(MotionStatechartNode):
-    state: Optional[str] = None
+    state: str | None = None
```
AbdelrhmanBassiouny, line 53: Why? is this better than optional? or is it a claude hiccup?

giskardpy/src/giskardpy/motion_statechart/tasks/cartesian_tasks.py
```diff
@@ -67,7 +66,9 @@ class CartesianTask(Task, ABC):
     )
     """Transformation matrix from root to goal_reference_frame link."""
 
-    _fk_binding: ForwardKinematicsBinding = field(kw_only=True, init=False)
+    _forward_kinematics_binding: ForwardKinematicsBinding = field(
```
AbdelrhmanBassiouny: what does it mean to be both keyword only and also init=False? I think init=Flase is enough.

giskardpy/src/giskardpy/motion_statechart/tasks/cartesian_tasks.py
```diff
@@ -362,7 +363,7 @@ def _compute_target_point(
 
     def on_tick(
         self, context: MotionStatechartContext
-    ) -> Optional[ObservationStateValues]:
+    ) -> ObservationStateValues | None:
```
AbdelrhmanBassiouny, line 363: same Optional change

giskardpy/src/giskardpy/motion_statechart/tasks/wiggle_insert.py
```diff
@@ -1,223 +1,227 @@
...
+    center_pull_strength_angle: float = field(default=0.1, kw_only=True)
+    """Pull back towards the starting angle during a random walk."""
+    center_pull_strength_vector: float = field(default=0.25, kw_only=True)
+    """Pull back towards the hole point during a random walk."""
+    weight: float = field(default=DefaultWeights.WEIGHT_ABOVE_CA, kw_only=True)
+    """Priority weight relative to other tasks."""
+
+    _hz: float = field(default=0.0, init=False, repr=False)
+    """Control frequency, used to scale the per-cycle noise."""
+    _v1: np.ndarray = field(default=None, init=False, repr=False)
```
AbdelrhmanBassiouny: no abbreviations

giskardpy/src/giskardpy/motion_statechart/tasks/wiggle_insert.py
```diff
@@ -1,223 +1,227 @@
...
+        self._rand_angle = sm.FloatVariable(f"{self.name}_rand_angle")
+        context.float_variable_data.register_expression(self._rand_angle)
 
         tip_V_hole_normal = context.world.transform(
-            target_frame=self.tip_link, spatial_object=self.hole_normal
+            target_frame=self.tip_link, spatial_object=hole_normal
         )
         tip_R_hole_normal = RotationMatrix.from_axis_angle(
-            angle=angle, axis=tip_V_hole_normal
+            angle=self._rand_angle, axis=tip_V_hole_normal
```
AbdelrhmanBassiouny: randy_angle?

giskardpy/src/giskardpy/motion_statechart/tasks/wiggle_insert.py
```diff
@@ -1,223 +1,227 @@
...
+            angle = self._random_sample_angle()
+        context.float_variable_data.set_value(self._rand_translation, translation)
+        context.float_variable_data.set_value(self._rand_angle, angle)
+        return None
+
+    def _random_sample_angle(self) -> float:
+        self._current_angle = ((random.random() - 0.5) * self.noise_angle) / self._hz
+        return self._current_angle
+
+    def _random_walk_angle(self) -> float:
```
AbdelrhmanBassiouny, line 266: docs

AbdelrhmanBassiouny: and for all methods below it in this file

giskardpy/src/giskardpy/motion_statechart/graph_node.py
```diff
@@ -165,16 +167,19 @@ def node_dependencies(self) -> List[MotionStatechartNode]:
...
-            str_representation = str(self.expression.is_const_true())
-        else:
-            str_representation = sm.trinary_logic_to_str(self.expression)
-        str_representation = re.sub(
-            r'"([^"]*?)/observation"', r'"\1"', str_representation
-        )
+            return str(self.expression.is_const_true())
+        str_representation = sm.trinary_logic_to_str(self.expression)
+        for variable in free_symbols:
+            if isinstance(variable, ObservationVariable):
```
AbdelrhmanBassiouny, line 179: could be a hint of open/closed violation?

giskardpy/src/giskardpy/motion_statechart/graph_node.py
```diff
@@ -1010,12 +1040,13 @@ def _worker_loop(self):
...
             try:
                 result = self._compute_observation()
-                # Accept only valid trinary values (floats expected)
                 self._last_result = result
                 self._has_result = True
             except Exception:
-                # On failure, keep previous result and mark as having no new value
-                pass
+                # Keep the previous result, but surface the failure instead of hiding it.
+                logger.exception(
```
AbdelrhmanBassiouny, line 1049: never heard of exception, how is that different from warn?

giskardpy/src/giskardpy/motion_statechart/graph_node.py
```diff
@@ -1086,10 +1117,10 @@ def build(self, context: MotionStatechartContext) -> NodeArtifacts:
     def on_tick(self, context: MotionStatechartContext) -> Optional[float]:
         raise self.exception
 
+    def _json_excluded_fields(self) -> set[str]:
```
AbdelrhmanBassiouny, line 1122: Maybe once the FieldMetaData is here, we could have a JSONFieldMetaData that would include such information at the field itself, and then the JSONSerializer would skip those automatically.

semantic_digital_twin/src/semantic_digital_twin/adapters/ros/visualization/spatial_type_marker_renderer.py
```diff
@@ -48,8 +49,12 @@ class SpatialTypeVisualization:
     color: Color = field(default_factory=Color)
     """The color of the markers. Ignored for the fixed RGB axes of pose-like spatial types."""
 
-    namespace: str = "spatial_type"
-    """The marker namespace. RViz scopes marker ids per namespace, so it must be unique per request."""
+    namespace: str = field(default_factory=lambda: f"spatial_type_{uuid.uuid4().hex}")
```
AbdelrhmanBassiouny, line 52: class SpatialTypePrefixedName(PrefixedName)?

giskardpy/src/giskardpy/motion_statechart/monitors/feature_monitors.py
```diff
@@ -1,28 +1,44 @@
...
+    """
+    Base for monitors that compare a controlled feature (attached to ``tip_link``) with a
+    reference feature (attached to ``root_link``) expressed in the root link frame.
+    """
+
+    tip_link: Body = field(kw_only=True)
+    """Link the controlled feature is attached to."""
+    root_link: Body = field(kw_only=True)
+    """Link the reference feature is attached to."""
     reference_feature: Union[Point3, Vector3] = field(init=False)
```
AbdelrhmanBassiouny: since you use `|` do it here as well instead of Union to stay consistent

giskardpy/src/giskardpy/motion_statechart/monitors/feature_monitors.py
```diff
@@ -1,28 +1,44 @@
...
-from giskardpy.motion_statechart.graph_node import MotionStatechartNode
+from giskardpy.motion_statechart.context import MotionStatechartContext
+from giskardpy.motion_statechart.graph_node import MotionStatechartNode, NodeArtifacts
 from semantic_digital_twin.spatial_types import Point3, Vector3
 from semantic_digital_twin.world_description.world_entity import Body
 
 
-@dataclass
+@dataclass(eq=False, repr=False)
 class FeatureMonitor(MotionStatechartNode):
```
AbdelrhmanBassiouny: Very generic name, when it is actually specific to a reference attached to root_link and something attached to tip_link. Maybe TipLinkFeatureMonitor?

AbdelrhmanBassiouny (changes requested):
looks nice

giskardpy/src/giskardpy/motion_statechart/nodes_for_testing/nodes_for_testing.py
ichumuh: [this is the modern way](https://docs.python.org/3/library/typing.html#typing.Optional)

giskardpy/src/giskardpy/motion_statechart/tasks/cartesian_tasks.py
ichumuh: it means that i copied a kw_only field, added the init=false and forgot about it

giskardpy/src/giskardpy/motion_statechart/graph_node.py
ichumuh: maybe but yagni, can be fixed if i ever add another Variable type to msc nodes

giskardpy/src/giskardpy/motion_statechart/graph_node.py
ichumuh: its a convenience method for logging an error with exception information.

giskardpy/src/giskardpy/motion_statechart/graph_node.py
ichumuh: for sure

semantic_digital_twin/src/semantic_digital_twin/adapters/ros/visualization/spatial_type_marker_renderer.py
ichumuh: creating a whole class just for this one line? i don't like that

giskardpy/src/giskardpy/motion_statechart/monitors/feature_monitors.py
ichumuh: sure, but we still have it in other place, i won't go through all files to change that

giskardpy/src/giskardpy/motion_statechart/monitors/feature_monitors.py
ichumuh: renamed to FeatureFunctionMonitor

giskardpy/src/giskardpy/motion_statechart/monitors/feature_monitors.py
```diff
@@ -1,28 +1,43 @@
...
+    """
+    Base for monitors that compare a controlled feature (attached to ``tip_link``) with a
+    reference feature (attached to ``root_link``) expressed in the root link frame.
+    """
+
+    tip_link: Body = field(kw_only=True)
+    """Link the controlled feature is attached to."""
+    root_link: Body = field(kw_only=True)
+    """Link the reference feature is attached to."""
+    reference_feature: Point3 | Vector3 = field(init=False)
```
LucaKro, line 23: why cant it be a transform?

giskardpy/src/giskardpy/motion_statechart/monitors/payload_monitors.py
```diff
@@ -67,34 +68,71 @@ def on_start(self, context: MotionStatechartContext):
...
-    Keeps track of how many ticks have passed since first True.
+    Number of ticks counted since the last start/reset.
     """
 
-    control_cycles: int = field(kw_only=True)
-    """
-    Turns True after this many control cycles.
-    """
+    def on_start(self, context: MotionStatechartContext):
+        self._counter = 0
```
LucaKro, line 86: why is this needed? for restarts? because the default is already 0 no?

giskardpy/src/giskardpy/motion_statechart/tasks/align_planes.py
```diff
@@ -39,8 +39,19 @@ class AlignPlanes(Task):
...
+    """
+    Angle in radians below which the planes are considered aligned.
+    """
+
     reference_velocity: float = field(default=0.5, kw_only=True)
+    """
+    Reference angular velocity for normalization in rad/s.
+    """
+
     weight: float = field(default=DefaultWeights.WEIGHT_ABOVE_CA, kw_only=True)
```
LucaKro: weight above california?

giskardpy/src/giskardpy/motion_statechart/tasks/cartesian_tasks.py
```diff
@@ -142,7 +139,7 @@ class CartesianPosition(CartesianTask):
     threshold: float = field(default=0.01, kw_only=True)
     """Distance threshold for goal achievement in meters."""
 
-    reference_velocity: Optional[float] = field(
+    reference_velocity: float | None = field(
```
LucaKro: why the optional? can the default factory return none in this case?

giskardpy/src/giskardpy/motion_statechart/graph_node.py
LucaKro: is this relevant now?

LucaKro (changes requested):

giskardpy/src/giskardpy/motion_statechart/tasks/cartesian_tasks.py
ichumuh: claude fail

giskardpy/src/giskardpy/motion_statechart/monitors/feature_monitors.py
ichumuh: it can be, but i can see that the last part of the doc string is misleading, i'll remove it

giskardpy/src/giskardpy/motion_statechart/monitors/payload_monitors.py
ichumuh: yes for restarts. i could also put it into `on_reset`, but idk so far `on_start` always felt cleaner than using `on_reset`

giskardpy/src/giskardpy/motion_statechart/tasks/align_planes.py
ichumuh: no, canada

giskardpy/src/giskardpy/motion_statechart/graph_node.py
ichumuh: i would prefer to change this in the next pr. i suspect it will be a few more changes which are off topic for what i want to fix here

giskardpy/src/giskardpy/motion_statechart/exceptions.py
```diff
@@ -153,6 +153,40 @@ def reason(self) -> str:
...
+    A condition may only reference the owning node itself or nodes sharing the same
+    parent.
+    """
+
+    dependency: MotionStatechartNode
+    """
+    The referenced node that lives in a different scope than the condition's owner.
+    """
+
+    def reason(self) -> str:
```
AbdelrhmanBassiouny, line 207: is a new method we implement in exceptions? Or is this just an addition from you here only?

giskardpy/src/giskardpy/motion_statechart/exceptions.py
```diff
@@ -153,6 +153,40 @@ def reason(self) -> str:
...
+            f'References "{self.dependency.unique_name}" from scope "{dependency_scope}", '
+            f'but the condition\'s owner lives in scope "{owner_scope}". '
+            f"Conditions may only reference the node itself or its siblings."
+        )
+
+    def suggest_correction(self) -> str:
+        return "Reference a sibling of the owning node instead, e.g. the template node that contains the dependency."
+
+    @staticmethod
+    def _scope_name(node: MotionStatechartNode) -> str:
```
AbdelrhmanBassiouny, line 220: docs

giskardpy/src/giskardpy/motion_statechart/graph_node.py
```diff
@@ -389,6 +393,14 @@ class MotionStatechartNode(SubclassJSONSerializer):
     The index of this node in the motion statechart.
     """
 
+    _node_id: str = field(init=False, default=None)
```
AbdelrhmanBassiouny, line 396: Why is this a string?

AbdelrhmanBassiouny: I guess it should be UUID?

giskardpy/src/giskardpy/motion_statechart/graph_node.py
AbdelrhmanBassiouny: cool

giskardpy/src/giskardpy/motion_statechart/graph_node.py
AbdelrhmanBassiouny: ok

semantic_digital_twin/src/semantic_digital_twin/adapters/ros/visualization/spatial_type_marker_renderer.py
AbdelrhmanBassiouny: me neither

test/giskardpy_test/test_motion_statechart/test_motion_statechart.py
```diff
@@ -428,6 +531,40 @@ def boom():
...
+    assert monitor._thread.is_alive()
+
+    monitor.cleanup(context=MotionStatechartContext.empty())
+
+    monitor._thread.join(timeout=1.0)
+    assert not monitor._thread.is_alive()
+
+
+def test_thread_payload_monitor_surfaces_compute_exception():
+    import giskardpy.motion_statechart.graph_node as graph_node_module
```
AbdelrhmanBassiouny: inline import

AbdelrhmanBassiouny (changes requested):

giskardpy/src/giskardpy/motion_statechart/monitors/feature_monitors.py
LucaKro: but then the typing is wrong as well no?

giskardpy/src/giskardpy/motion_statechart/graph_node.py
ichumuh: it's a string because i have some ugly chicken egg problem. index is the actuall UUID like thing and _node_id is used for a human readable unique name

giskardpy/src/giskardpy/motion_statechart/monitors/payload_monitors.py
LucaKro: do you also call it on the actual first time start? if no then on reset would be more fitting

LucaKro (changes requested):

giskardpy/src/giskardpy/motion_statechart/monitors/feature_monitors.py
ichumuh: no, why? it can be both a point and a vector

giskardpy/src/giskardpy/motion_statechart/monitors/payload_monitors.py
ichumuh: in the end, it doesn't really matter. currently, the default for this field is 0 too, so it also works if i move it to `on_reset`, but i prefer to use `on_reset` only for stuff that is specifically needed for resetting and not initialization. another advantage of using `on_start` is that if a node is reset, but hasn't started again, we can still see the `_counter` in the debugger.

giskardpy/src/giskardpy/motion_statechart/monitors/feature_monitors.py
LucaKro: > why cant it be a transform?

> it can be

ichumuh: oh i misunderstood your first comment. i thought you meant why the point3/vector3 has to be in a specific frame and can't be transformed instead.
in theory there could be cases where a Pose would make sense, but there are none implemented, so Point3 | Vector3 is correct for now

giskardpy/src/giskardpy/motion_statechart/exceptions.py
ichumuh: its not a new method, just something inside of giskard. anyway, i removed most reasons and replaced them with proper exceptions instead

AbdelrhmanBassiouny (approved):

giskardpy/src/giskardpy/motion_statechart/graph_node.py
AbdelrhmanBassiouny: I see

LucaKro (approved):

PR #439, opened 2026-07-13
Added support for custom rectangular camera resolutions in RayTracer

Sanic:
This allows renderings by RayTracer to be of shape AxB instead of only AxA. This is a prerequisite for rendering in the same aspect ratios as the employed sensors on our robots to enable more realistic simulation-based perception as well as imagistic reasoning.


semantic_digital_twin/src/semantic_digital_twin/spatial_computations/raytracer.py
```diff
@@ -1,19 +1,85 @@
...
-from semantic_digital_twin.datastructures.types import NpMatrix4x4
 from semantic_digital_twin.world_description.world_entity import Body
 from semantic_digital_twin.spatial_types.spatial_types import GenericSpatialType
+from semantic_digital_twin.exceptions import InvalidCameraResolutionError
 
 if TYPE_CHECKING:
     from semantic_digital_twin.world import World
 
 
+@dataclass(frozen=True)
```
Tigul: Do you need frozen? I remember this could cause issues with sqlalchemy

semantic_digital_twin/src/semantic_digital_twin/spatial_computations/raytracer.py
```diff
@@ -1,19 +1,85 @@
...
+                resolution=resolution,
+                reason="resolution must be an integer, list, or tuple.",
+            )
+
+        if len(resolution) != 2:
+            raise InvalidCameraResolutionError(
+                resolution=resolution,
+                reason="resolution pair must contain exactly two values.",
+            )
+        return cls(width=resolution[0], height=resolution[1])
```
Tigul: you can also write this as cls(*resolution)

semantic_digital_twin/src/semantic_digital_twin/spatial_computations/raytracer.py
```diff
@@ -110,7 +176,7 @@ def update_transforms(self):
     def create_segmentation_mask(
         self,
         camera_pose: GenericSpatialType,
-        resolution: int = 512,
+        resolution: CameraResolutionValue = 512,
```
Tigul: If this is a CameraResolutionValue the default here is wrong

semantic_digital_twin/src/semantic_digital_twin/spatial_computations/raytracer.py
```diff
@@ -158,7 +225,7 @@ def create_segmentation_mask(
     def create_depth_map(
         self,
         camera_pose: GenericSpatialType,
-        resolution: int = 512,
+        resolution: CameraResolutionValue = 512,
```
Tigul: Same here, wrong default

semantic_digital_twin/src/semantic_digital_twin/spatial_computations/raytracer.py
```diff
@@ -211,20 +279,24 @@ def create_depth_map(
         return a
 
     def create_camera_rays(
-        self, camera_pose: GenericSpatialType, resolution: int = 512, fov=90
-    ) -> Tuple[np.ndarray, np.ndarray, np.ndarray]:
+        self,
+        camera_pose: GenericSpatialType,
+        resolution: CameraResolutionValue = 512,
```
Tigul: default

semantic_digital_twin/src/semantic_digital_twin/spatial_computations/raytracer.py
```diff
@@ -119,7 +185,8 @@ def create_segmentation_mask(
         in the mask corresponds to the index of a body in the scene or -1 if no body is hit at that pixel.
 
         :param camera_pose: The position of the camera.
-        :param resolution: The resolution of the segmentation mask.
+        :param resolution: The camera resolution. An integer creates a square image; a width-height pair creates a
```
Tigul: Hmm I see what you are doing here but I would keep this consistent. When the typing says camera resolution value then only use that

semantic_digital_twin/src/semantic_digital_twin/exceptions.py
```diff
@@ -241,6 +241,29 @@ class UsageError(LogicalError):
     """
 
 
+@dataclass
+class InvalidCameraResolutionError(UsageError):
+    """
+    Raised when a camera resolution cannot describe a ray tracing image.
+    """
+
+    resolution: Any
```
Tigul, line 255: Camera Resolution Value ?

semantic_digital_twin/src/semantic_digital_twin/exceptions.py
```diff
@@ -241,6 +241,29 @@ class UsageError(LogicalError):
...
+    """
+    The invalid camera resolution value.
+    """
+
+    reason: str
```
Tigul, line 260: Isn't this inherited?

Tigul (changes requested):
good work

semantic_digital_twin/src/semantic_digital_twin/spatial_computations/raytracer.py
Sanic: No. I'll change that. Thanks for the hint!

semantic_digital_twin/src/semantic_digital_twin/spatial_computations/raytracer.py
Sanic: int is intentionally a part of CameraResolutionValue to preserve backwards compatibility. But i'll update the value to a tuple to clearly indicate that we can now accept tuples.

semantic_digital_twin/src/semantic_digital_twin/exceptions.py
Sanic: Not always in this case. It will hold the (invalid/bad) value that was passed to the resolution parameter, which could theoretically be anything.

semantic_digital_twin/src/semantic_digital_twin/exceptions.py
Sanic: No, reason is not inherited. The parent classes don't have such an attribute. Some other exceptions have it, but they don't fit as super class.

semantic_digital_twin/src/semantic_digital_twin/spatial_computations/raytracer.py
Sanic: Same as comment above. Will update the default to tuple.

semantic_digital_twin/src/semantic_digital_twin/spatial_computations/raytracer.py
Sanic: Same as comment above. Will update the default to tuple.

semantic_digital_twin/src/semantic_digital_twin/spatial_computations/raytracer.py
Sanic: Updated comment

semantic_digital_twin/src/semantic_digital_twin/spatial_computations/raytracer.py
```diff
@@ -1,19 +1,85 @@
...
+            )
+
+        if len(resolution) != 2:
+            raise InvalidCameraResolutionError(
+                resolution=resolution,
+                reason="resolution pair must contain exactly two values.",
+            )
+        return cls(*resolution)
+
+    def to_shape(self) -> tuple[int, int]:
```
tomsch420: just make this a property shape?

semantic_digital_twin/src/semantic_digital_twin/spatial_computations/raytracer.py
```diff
@@ -1,19 +1,85 @@
...
+        """
+        Returns the resolution in the axis order used by :mod:`trimesh`.
+        """
+        return self.width, self.height
+
+
+CameraResolutionValue = int | list[int] | tuple[int, int] | CameraResolution
```
tomsch420: is this something like a typedef? if so, use typdef

semantic_digital_twin/src/semantic_digital_twin/exceptions.py
```diff
@@ -241,6 +241,29 @@ class UsageError(LogicalError):
     """
 
 
+@dataclass
+class InvalidCameraResolutionError(UsageError):
```
tomsch420, line 250: this class looks like it should be a subclass of type error and dataclass exception. Also its kind of pointless since its just a type error or am i missing something?

tomsch420 (changes requested):

semantic_digital_twin/src/semantic_digital_twin/exceptions.py
Sanic: It will also be raised when a corret type is supplied, but the values are bad. For example, if you try to pass negative width or height values.
Would you still recommend inheriting from `TypeError` as well?

I could also change the raised exceptions in Raytracer to raise `TypeError` if it's only a matter of wrong types and only raise InvalidCameraResolutionError if the (correctly typed values are bad.

What do you think?

Sanic:
> good work

Thanks! I've addressed your comments.

semantic_digital_twin/src/semantic_digital_twin/spatial_computations/raytracer.py
```diff
@@ -1,19 +1,86 @@
...
+    height: int
+    """
+    Number of pixels along the image height.
+    """
+
+    def __post_init__(self) -> None:
+        """
+        Validates that the resolution can describe a camera image.
+        """
+        if not isinstance(self.width, Integral) or not isinstance(
```
LucaKro, line 39: i dont like these checks, in my opinion the typehints should be enough in python, we cannot always check that the user adheres to the explicit requirements of the class

semantic_digital_twin/src/semantic_digital_twin/spatial_computations/raytracer.py
```diff
@@ -1,19 +1,86 @@
...
+            self.height, Integral
+        ):
+            raise InvalidCameraResolutionError(
+                resolution=(self.width, self.height),
+                reason="width and height must be integers.",
+            )
+        if self.width <= 0 or self.height <= 0:
```
LucaKro, line 46: this check on the other hand i think makes sense in general, since its some implicit condition made explicit.

semantic_digital_twin/src/semantic_digital_twin/spatial_computations/raytracer.py
```diff
@@ -1,19 +1,86 @@
...
+            raise InvalidCameraResolutionError(
+                resolution=(self.width, self.height),
+                reason="width and height must be positive.",
+            )
+
+    @classmethod
+    def from_value(cls, resolution: CameraResolutionValue) -> CameraResolution:
```
LucaKro, line 72: why do we need this abstraction? cant we just have a from_integral, from_iterable etc, and then pass CameraResolution around (for example to the raytracer) instead of CameraResolutionValue?

LucaKro (changes requested):
nice, this could also be used to describe the camera resolution in the robot semantic annotations no?

semantic_digital_twin/src/semantic_digital_twin/spatial_computations/raytracer.py
Sanic: Understandable.
In this case i actually like them, because we provide explicit feedback if users try to pass float values, which are not supported.

semantic_digital_twin/src/semantic_digital_twin/spatial_computations/raytracer.py
Sanic: I've implemented the from_integral and from_iterable methods.
The CameraResolutionValue ensures that we don't break the API of RayTracer

Sanic:
> nice, this could also be used to describe the camera resolution in the robot semantic annotations no?

Jeah, good point.

semantic_digital_twin/src/semantic_digital_twin/spatial_computations/raytracer.py
LucaKro: but one can always make that argument, these types of explicit checks always give you more freedom to provide feedback to a user. I dont think its unique to here. 
why should we do it here, but not everywhere else?
Imo this should be removed

semantic_digital_twin/src/semantic_digital_twin/spatial_computations/raytracer.py
LucaKro: where is CameraResolutionValue coming from? and why would we break the api? I think there is an argument to be made to just change the api for the raytracer, as its probably outdated then

LucaKro: oh just saw it below.
1. if you make type declarations like CameraResolutionValue below, please if possible move them to the top of the file
2. i think we should adjust the api of raytracer to always take "CameraResolution" as input

LucaKro: also cant from_value now be removed that we have the other two methods?

LucaKro (changes requested):

Sanic:
Hi @Tigul @tomsch420 @LucaKro 
After reading the comments and talking to @Tigul , i've changed the PR such that:
1) We don't have the CameraResolutionValue type hint any more
2) We don't have to handle multiple types of data to instantiate a valid camera resolution
3) CameraResolution is now a dataclass living in semantic_digital_twin/datastructures
4) Raytracer API had to be changed, but the callers in the monorepo are adapted to comply with the new API.

Please re-review :heart_hands:

LucaKro (approved):

semantic_digital_twin/src/semantic_digital_twin/exceptions.py
tomsch420: hmm normally we only provide such exceptions when there are non trivial errors. But its fine

semantic_digital_twin/src/semantic_digital_twin/datastructures/camera_resolution.py
```diff
@@ -0,0 +1,32 @@
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+
+from semantic_digital_twin.exceptions import InvalidCameraResolutionError
+
+
+@dataclass
+class CameraResolution:
```
tomsch420, line 9: i think there should be an association from Camera to this

tomsch420 (changes requested):

semantic_digital_twin/src/semantic_digital_twin/datastructures/camera_resolution.py
Sanic: Good point. One of the next points on my agenda was to make a PR to extend related SemanticAnnotations such that they cover all the different perception use cases and sensors we encountered in RoboKudo.
I can add it in the upcoming PR

tomsch420 (approved):

Tigul (approved):

PR #440, opened 2026-07-14
Add xArm5 robot

JTShuai:
## Summary

Adds support for the UFACTORY xArm 5: the semantic annotation class, its self-collision
config, and the robot description itself (URDF + meshes), so the robot loads from a plain
checkout without any ROS installation or environment setup.

## What is included

- `semantic_digital_twin/robots/xarm5.py` — `XArm5` semantic annotation, mirroring the
  structure of the other robots in this package (`pr2.py` / `stretch.py`): leaf-first part
  classes (`XArm5Arm`, `XArm5Gripper`, `XArm5Camera`) tied together by the robot class.
- `semantic_digital_twin/resources/collision_configs/xarm5.srdf` — self-collision matrix,
  loaded via `importlib.resources` like the other shipped configs.
- `semantic_digital_twin/resources/xarm_description/` — the robot description as a standard
  ROS package layout (`urdf/xarm5.urdf` + 10 mesh files, ~3 MB total).
- `test/semantic_digital_twin_test/test_robots/test_world_xarm5.py` — load test.

## Design notes

- The description packages of the existing robots (e.g. `iai_pr2_description`) are provided
  by the environment. No public package ships this exact flat-URDF layout for the xArm 5, so
  the description is bundled in `resources/` instead; otherwise the class would not be
  loadable from a checkout.
- `get_ros_file_path()` registers the bundled `xarm_description` on `ROS_PACKAGE_PATH`
  (appended, not prepended) and returns the usual `package://` URI. An `xarm_description`
  already present in the environment therefore takes precedence over the bundled copy.
- The xArm 5 is a table-mounted 5-DoF arm with a bare end-effector flange: the end effector
  has no fingers (`root` and `tool_frame` are both `link_eef`), and the camera is a
  placeholder at the flange so that robots with sensors validate; both are documented in the
  class docstrings and can be replaced once real hardware is mounted.

## Testing

- `test_xarm5_description_ships_with_package`: parses the URDF via
  `XArm5.get_ros_file_path()` and applies the semantic annotation in an environment with no
  ROS installation and no `ROS_PACKAGE_PATH` set — passes (8 bodies, root `link_base`, one
  arm).


AbdelrhmanBassiouny (approved):
Looks good, thanks 👍

tomsch420 (approved):

semantic_digital_twin/resources/xarm_description/urdf/xarm5.urdf
Tigul, line 1: Don't add the urdf directly to the Repo. Instead add the xArm repo here: https://github.com/cram2/cognitive_robot_abstract_machine/blob/main/.github/docker/setup_workspace.py#L269 and remove (with list provided there) all other packages except the xarm_description so we do not inroduce dependencies for the hardware controller

Tigul (changes requested):
Looks good

semantic_digital_twin/resources/xarm_description/urdf/xarm5.urdf
JTShuai: > Don't add the urdf directly to the Repo. Instead add the xArm repo here: https://github.com/cram2/cognitive_robot_abstract_machine/blob/main/.github/docker/setup_workspace.py#L269 and remove (with list provided there) all other packages except the xarm_description so we do not inroduce dependencies for the hardware controller

Done. The URDF/meshes are gone from the repo; the workspace setup now clones xarm_ros2 (jazzy) stripped to xarm_description only. Verified in the CI image: the stripped clone colcon-builds and the device xacro (dof:=5) loads into the same structure as before.

semantic_digital_twin/src/semantic_digital_twin/robots/xarm5.py
```diff
@@ -0,0 +1,173 @@
+"""
+Semantic annotation for the UFACTORY xArm 5.
+
+A table-mounted 5-DoF serial arm with a bare end-effector flange and a placeholder
+camera. Mirrors the structure of the other robots in this package (see ``pr2.py`` /
```
LucaKro, line 5: i dont like descriptions like this:

> Mirrors the structure of the other robots in this package (see ``pr2.py`` /
``stretch.py``): each part is a leaf-first class implementing the three abstract
methods, and the robot class ties them together.

as they are reasonings done by the LLM that may not hold, and that dont, in my opinion, add any value to someone reading the code.

semantic_digital_twin/src/semantic_digital_twin/robots/xarm5.py
LucaKro, line 1: could you post a picture of this robot so that i can judge if the semantic annotations are constructed correctly?

semantic_digital_twin/src/semantic_digital_twin/robots/xarm5.py
```diff
@@ -0,0 +1,173 @@
...
+    and a placeholder camera. It owns the arm and the camera directly (no MobileBase,
+    Torso or Neck).
+    """
+
+    @classmethod
+    def get_ros_file_path(cls) -> str:
+        return "package://xarm_description/urdf/xarm_device.urdf.xacro"
+
+    @classmethod
+    def get_xacro_mappings(cls) -> Dict[str, str]:
```
LucaKro, line 148: please make this a method of AbstractRobot which can be overridden if needed

LucaKro (changes requested):
nice, very exciting having someone external contribute 💪 thank you!

were there any troubles in setting everything up/understanding the code/figuring out what to extend?

Tigul (approved):

PR #443, opened 2026-07-15
Enforce docformatter repo-wide via pre-commit, with an automatic black fallback

AbdelrhmanBassiouny:
Replaces the standalone black pre-commit hook with a local hook that runs black then docformatter, keeping docformatter changes only when a second black pass is a no-op (works around real docformatter bugs). Reformats ~740 files accordingly. Full detail: https://github.com/AbdelrhmanBassiouny/cognitive_robot_abstract_machine/pull/69


scripts/format_docstrings.py
```diff
@@ -0,0 +1,169 @@
...
+    """
+    Format each given file, writing back changes and reporting skipped docstrings.
+
+    :param file_arguments: File paths to format, as passed by pre-commit.
+    :return: 1 if any file was modified (matching pre-commit's convention for
+        autoformatting hooks), 0 otherwise.
+    """
+    any_file_changed = False
+
+    for file_argument in file_arguments:
```
tomsch420: tqdm.tqdm

tomsch420 (changes requested):

tomsch420 (approved):

scripts/format_docstrings.py
AbdelrhmanBassiouny: Done

drawer_explanation.pdf
Tigul, line 1: Whats with the pdfs?

Tigul (approved):

drawer_explanation.pdf
AbdelrhmanBassiouny: not from this pr

PR #448, opened 2026-07-15
Fix subscriber-count discovery race causing flaky giskardpy self-collision test

AbdelrhmanBassiouny:
Fixes a subscriber-count discovery race in Synchronizer.publish() that could let a synchronous world-modify publish return before a remote subscriber actually applied the update, plausibly causing the flaky test_attached_self_collision_avoid_stick failure. Full detail: https://github.com/AbdelrhmanBassiouny/cognitive_robot_abstract_machine/pull/72


tomsch420 (approved):

davidprueser (approved):

semantic_digital_twin/src/semantic_digital_twin/adapters/ros/world_synchronizer.py
```diff
@@ -286,7 +342,9 @@ def publish(self, msg: Message):
 
         with self._publish_lock, self._acknowledge_condition_variable:
             self._current_publication_event_id = msg.publication_event_id
-            self._expected_acknowledgment_count = self._snapshot_subscribers()
+            self._expected_acknowledgment_count = (
```
LucaKro, line 345: you are sometimes expecting the highest count seen...wouldnt that result in a deadlock if we wait for 10 acknowledgements, even though only 6 are active because 10 was briefly the highest count?

LucaKro (changes requested):

semantic_digital_twin/src/semantic_digital_twin/adapters/ros/world_synchronizer.py
AbdelrhmanBassiouny: I think what will happen in that case is a timeout not a deadlock.

AbdelrhmanBassiouny: at least that what claude said haha, can you verify that?

AbdelrhmanBassiouny: or maybe I can make a test for that

AbdelrhmanBassiouny: done, check out the new tests

LucaKro (approved):

PR #450, opened 2026-07-15
Generative match .where() without an explicit resolve()

AbdelrhmanBassiouny:
Removes the need for an explicit resolve() before .where() on a match — Match.variable was overloaded as both the subject and a not-lowered-yet sentinel; this decouples the two responsibilities so the subject is created eagerly and resolved is the sole lowering signal. Backward compatible. Full krrood test suite: 944 passed, 6 skipped.

Full detail: https://github.com/AbdelrhmanBassiouny/cognitive_robot_abstract_machine/pull/4


krrood/src/krrood/entity_query_language/factories.py
```diff
@@ -101,6 +101,28 @@ def set_of(*selected_variables: Union[Selectable[T], Any]) -> SetOf:
 # %% Variable Declaration
 
 
+def _resolve_domain(type_: Type[T], domain: Optional[DomainType]) -> DomainType:
```
davidprueser: Maybe put this into a utils or helper file? This doesn't look like it's limited to this module

tomsch420: does this really belong in factories?

tomsch420 (changes requested):

davidprueser (changes requested):

krrood/src/krrood/entity_query_language/factories.py
AbdelrhmanBassiouny: moved to a new helpers module

AbdelrhmanBassiouny: moved to a new helpers module

tomsch420 (approved):

davidprueser (approved):

PR #452, opened 2026-07-16
Refactor the EQL Query: spec/product lifecycle, subquery caching, and a composable result pipeline

AbdelrhmanBassiouny:
Untangles EQL Query's spec/product/embeddable-value conflation into a dirty-flag lifecycle with a single compile path, caches uncorrelated subquery results instead of re-evaluating per outer row, and expresses ordering/quantification as composable ResultTransformer stages instead of expression-tree wrapper nodes. 48 files, +1,665/-832. Full krrood suite green (855 passed), EQL+ORM suites green (406 passed).

Full detail: https://github.com/AbdelrhmanBassiouny/cognitive_robot_abstract_machine/pull/5

PR #453, opened 2026-07-17
EQL core semantics (foundation for the RDR engine)

AbdelrhmanBassiouny:
EQL-core behaviour the RDR/EQL engine builds on: core expression / Match / factory tweaks, operator & rule adjustments, scope handling, verbalization tidy-ups, patterns (field metadata / role predicates), and SDT reasoning glue, with tests. Parent query-class-refactor has landed in main (cram2 #452). Full detail: https://github.com/AbdelrhmanBassiouny/cognitive_robot_abstract_machine/pull/28


krrood/src/krrood/class_diagrams/utils.py
```diff
@@ -407,3 +428,18 @@ def get_object_by_name_from_another_object_in_same_module(
...
+
+
+def get_class_unique_name(cls: type) -> str:
+    """
+    Returns a unique identifier for a class based on its source file and qualified name.
+
+    :param cls: The class.
+    :return: The unique name of the class
+    """
+    try:
```
tomsch420, line 440: why is this not class.module + class.name? Thats already unqieu

tomsch420: get_full_class_name is what i mean

krrood/src/krrood/entity_query_language/core/base_expressions.py
```diff
@@ -775,6 +923,21 @@ def collect(node: Optional[OperationResult]) -> None:
     def has_value(self) -> bool:
         return self.operand is not None and self.operand._id_ in self.bindings
 
+    @property
+    def is_condition_false(self) -> bool:
+        """Canonical condition-truth rule used by :meth:`~krrood.entity_query_language.core.base_expressions.TruthValueOperator._evaluate_child_as_condition_`
```
tomsch420: docformatter where

krrood/src/krrood/entity_query_language/core/base_expressions.py
```diff
@@ -775,6 +923,21 @@ def collect(node: Optional[OperationResult]) -> None:
...
+        and :meth:`~krrood.entity_query_language.core.base_expressions.SymbolicExpression._evaluate_conclusions_and_update_bindings_`.
+
+        For expressions that bind a direct value (``Attribute``, ``Comparator``, ``Variable``, …)
+        truth is derived from the value's boolean content.  For logical operators
+        that manage their own ``is_false`` flag (``NOT``, ``AND``, ``OR``, …) and produce no
+        direct value, the flag is used as-is.
+        """
+        if self.has_value:
+            v = self.value
```
tomsch420: why are you assigning v here?

krrood/src/krrood/entity_query_language/rules/conclusion.py
```diff
@@ -20,7 +20,12 @@
     Selectable,
     BinaryExpression,
 )
-from krrood.entity_query_language.core.variable import Variable
+from krrood.entity_query_language.core.variable import Literal, Variable
+
+
+def _unwrap(value: Any) -> Any:
+    """Unwrap a :class:`Literal` wrapper to get the raw value."""
```
tomsch420: utils?

krrood/src/krrood/entity_query_language/_stack.py
```diff
@@ -53,6 +53,14 @@ class StackFrame:
     """
     Dotted module name (string, not ``ModuleType``) to avoid reference leaks.
     """
+    global_ns: Optional[Dict[str, Any]] = None
```
tomsch420: no abbrivations

krrood/src/krrood/entity_query_language/exceptions.py
```diff
@@ -692,6 +692,23 @@ def suggest_correction(self) -> str:
         return ""
 
 
+@dataclass
+class SelfReferentialInsertionError(DataclassException):
+    """Raised when insert_at would create a self-referential selector node."""
+
+    anchor: "SymbolicExpression" = field(kw_only=True)
```
tomsch420: haram typing

krrood/src/krrood/entity_query_language/scope.py
```diff
@@ -0,0 +1,170 @@
...
+
+from krrood.entity_query_language import factories as eql
+from krrood.entity_query_language.rules.conclusion import Add
+from krrood.entity_query_language.rules.conclusion_selector import (
+    Alternative,
+    Next,
+    Refinement,
+)
+
+EQL_PACKAGE = "krrood.entity_query_language"
```
tomsch420: why?

krrood/src/krrood/entity_query_language/scope.py
```diff
@@ -0,0 +1,170 @@
...
+        return True
+    module = inspect.getmodule(frame_info.frame)
+    module_name = module.__name__ if module else ""
+    return any(
+        module_name == pkg or module_name.startswith(pkg + ".") for pkg in skip_packages
+    )
+
+
+def capture_caller_scope(
+    skip_packages=(EQL_PACKAGE,), max_depth: int = 60
```
tomsch420: this signature is not nice. Also type hints. When you want to give more skip packages here, youll have to remember EQL_Package yourself

krrood/src/krrood/entity_query_language/scope.py
```diff
@@ -0,0 +1,170 @@
...
+            if _is_internal_frame(frame_info, skip_packages):
+                continue
+            frame = frame_info.frame
+            return {**frame.f_globals, **frame.f_locals}
+        return {}
+    finally:
+        del frames
+
+
+def eql_factory_namespace() -> Dict[str, Any]:
```
tomsch420, line 94: why do we need such a function?

tomsch420 (changes requested):

krrood/src/krrood/class_diagrams/utils.py
AbdelrhmanBassiouny: It's not unique because you can run into a situation where you have two versions of the library, one installed via pip or something, and another that is the current source, so this solves this issue by using the absolute path.

krrood/src/krrood/entity_query_language/core/base_expressions.py
AbdelrhmanBassiouny: probably a file where black and docformatter disagreed, and thus the script reverts to just black.

krrood/src/krrood/entity_query_language/core/base_expressions.py
AbdelrhmanBassiouny: good question idk

krrood/src/krrood/entity_query_language/rules/conclusion.py
AbdelrhmanBassiouny: I guess this should be moved to a helpers.py

krrood/src/krrood/entity_query_language/scope.py
AbdelrhmanBassiouny: Maybe one can import entity_query_package and compare its name or import path to the file or module or object in question, and it can be done locally without needing this global variable.

krrood/src/krrood/entity_query_language/scope.py
AbdelrhmanBassiouny: yeah true the EQL_PACKAGE should be done locally inside the function and not be in the args at all , and the skip packages be an optional keyword argument that is an empty tuple or None by default.

krrood/src/krrood/entity_query_language/scope.py
AbdelrhmanBassiouny: as a scope (namespace) that is injected and used to write rules or eql queries anywhere programatically, like in an IpythonInterface, for example.

krrood/src/krrood/entity_query_language/core/base_expressions.py
AbdelrhmanBassiouny: removed

krrood/src/krrood/entity_query_language/rules/conclusion.py
AbdelrhmanBassiouny: done

tomsch420 (approved):

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
```diff
@@ -7,17 +7,6 @@
```
sunava, line 7: Almost this entire diff is an IDE reformat away from Black style (4-space continuation → 8-space) plus stray double blank lines, are we not using black anymore?

krrood/src/krrood/entity_query_language/rules/conclusion_selector.py
```diff
@@ -51,23 +52,63 @@ def create_and_update_rule_tree(
...
-        new_condition = chained_logic(AND, *conditions)
+        return cls.insert_at(cls._get_current_context_condition(), *conditions)
 
-        current_context = cls._get_current_context_condition()
+    @classmethod
+    def insert_at(
+        cls,
+        anchor: SymbolicExpression,
+        *conditions: ConditionType,
+    ) -> Self:
```
sunava: `insert_at` is annotated `-> Self` but returns `new_condition` is this the correct behavior?

krrood/src/krrood/entity_query_language/_stack.py
```diff
@@ -53,6 +53,17 @@ class StackFrame:
     """
     Dotted module name (string, not ``ModuleType``) to avoid reference leaks.
     """
+    global_namespace: Optional[Dict[str, Any]] = None
+    """
+    Shallow snapshot of the frame's ``f_globals``, or ``None`` when scope was not
```
sunava, line 58: just confused why not f_locals?

krrood/src/krrood/entity_query_language/exceptions.py
```diff
@@ -692,11 +688,29 @@ def suggest_correction(self) -> str:
         return ""
 
 
+@dataclass
+class SelfReferentialInsertionError(DataclassException):
+    """
+    Raised when insert_at would create a self-referential selector node.
+    """
+
+    anchor: SymbolicExpression = field(kw_only=True)
```
sunava, line 697: docstring

krrood/src/krrood/entity_query_language/scope.py
```diff
@@ -0,0 +1,180 @@
...
+    :param obj: The object (typically an underspecified query/variable) to annotate.
+    :param scope: An explicit scope to attach; if ``None``, the caller scope is
+        captured.
+    :return:``obj`` (for chaining).
+    """
+    if scope is None:
+        scope = capture_caller_scope()
+    try:
+        setattr(obj, _SCOPE_ATTRIBUTE, scope)
+    except (AttributeError, TypeError):
```
sunava, line 142: `setattr` in a try/except that silently `pass`es hides an illegal state.A `__slots__` object that can't carry the scope would be silently dropped. Prefer an explicit guard (e.g. check for `__dict__`) or let it raise.

sunava (changes requested):

krrood/src/krrood/entity_query_language/rules/conclusion_selector.py
AbdelrhmanBassiouny: No you are right I think

krrood/src/krrood/entity_query_language/_stack.py
AbdelrhmanBassiouny: It is gotten feom f_locals, but later I also take a shallow copy from it because I don't want all of it

krrood/src/krrood/entity_query_language/scope.py
AbdelrhmanBassiouny: Yeah it's fine if this happens it's not an illegal state, it's an explanation of inference , if we cannot put the scope then we cannot put in the explanation where this inference happened but that's not a big issue and it rarely happens, at least it doesn't happen in cram at all till now that we use slots.

sunava (approved):

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
AbdelrhmanBassiouny: Reformatted it

PR #457, opened 2026-07-19
[EQLVerbalization] Extract reusable verbalization surface-verification framework

AbdelrhmanBassiouny:
Extracts the verbalization surface-verification framework out of krrood's surface test into a reusable module on main, so any package defining SymbolicCallable subclasses can verify its surfaces with three one-line tests.

Full detail: https://github.com/AbdelrhmanBassiouny/cognitive_robot_abstract_machine/pull/86


krrood/src/krrood/entity_query_language/testing/surface_verification.py
```diff
@@ -0,0 +1,222 @@
...
+    """
+
+    value: Any
+    """
+    The concrete value to pass for that field.
+    """
+
+
+@dataclass(frozen=True)
+class SymbolicSurfaceSnapshot:
```
tomsch420, line 80: sounds like a singleton to me

tomsch420 (changes requested):

krrood/src/krrood/entity_query_language/testing/surface_verification.py
AbdelrhmanBassiouny: actually no, this will have an instance for every package that has SymbolicCallable implementations and want to maintain and test their surface verbalizations.

tomsch420 (approved):

krrood/src/krrood/entity_query_language/verbalization/surface_verification.py
```diff
@@ -0,0 +1,222 @@
...
+        """
+        fragment_less = sorted(
+            module_and_class_name(cls)
+            for cls in self.discovered_callables()
+            if not self.has_fragment(cls)
+        )
+        assert not fragment_less, (
+            "These symbolic callables have no verbalization fragment and must implement "
+            f"_verbalization_fragment_: {fragment_less}."
+        )
```
davidprueser: I feel like these assert functions should not be in the source code. Either they are meant for testing and dont belong here, or the source code should use them, but why not use custom exceptions that get raised, as is done everywhere else? I dont like these kinds of verifications here

krrood/src/krrood/entity_query_language/testing/surface_verification.py
```diff
@@ -0,0 +1,222 @@
...
+            module_and_class_name(cls)
+            for cls in self.discovered_callables()
+            if not self.has_fragment(cls)
+        )
+        assert not fragment_less, (
+            "These symbolic callables have no verbalization fragment and must implement "
+            f"_verbalization_fragment_: {fragment_less}."
+        )
+
+    def assert_surfaces_cover_every_callable(self) -> None:
```
davidprueser, line 166: same here

krrood/src/krrood/entity_query_language/testing/surface_verification.py
```diff
@@ -0,0 +1,222 @@
...
+            module_and_class_name(surface.callable_class) for surface in self.surfaces
+        }
+        missing = sorted(discovered - declared)
+        stale = sorted(declared - discovered)
+        assert discovered == declared, (
+            f"Declared surfaces are out of sync. Discovered classes with no entry (add one): "
+            f"{missing}. Entries whose class is no longer discovered (remove them): {stale}."
+        )
+
+    def assert_declared_surfaces_render_as_stated(self) -> None:
```
davidprueser, line 186: same here

krrood/src/krrood/utils.py
```diff
@@ -73,6 +73,10 @@ def get_full_class_name(cls):
 
 
 def module_and_class_name(t: Union[Type, _SpecialForm]) -> str:
```
davidprueser: Abbreviation spotted 👀

test/krrood_test/test_class_diagram/test_class_implements_own_method.py
```diff
@@ -0,0 +1,85 @@
+from __future__ import annotations
+
+from krrood.class_diagrams.utils import class_implements_own_method
+
+
+class MethodOwner:
```
davidprueser, line 6: So where is the distinction between defining classes here and defining classes in the example_classes.py?

davidprueser (changes requested):
Good stuff like always

test/krrood_test/test_class_diagram/test_class_implements_own_method.py
AbdelrhmanBassiouny: So we talked and decided that we would put them in an external file if there's a need for that (e.g. you are parsing that file in your tests so you won't have to parse the file where the test actually exists).

davidprueser:
What about my comments regarding the assert_... functions above?

krrood/src/krrood/entity_query_language/verbalization/surface_verification.py
AbdelrhmanBassiouny: well we discussed this and it is fine as it meant to be imported by other packages, what I did though now was put it in a dedicating testing subpackage in eql source code similar to how numpy does it.

krrood/src/krrood/entity_query_language/testing/surface_verification.py
AbdelrhmanBassiouny: addressed

krrood/src/krrood/entity_query_language/testing/surface_verification.py
AbdelrhmanBassiouny: addressed

krrood/src/krrood/entity_query_language/verbalization/surface_verification.py
AbdelrhmanBassiouny: and this specific method was actually removed because abstractmethod already does that job.

davidprueser (approved):

PR #463, opened 2026-07-20
Lift me up

Tigul:
# Description 
Adds an elevator semantic annotation and its create_in_world_with_bodies method


semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/semantic_annotations.py
```diff
@@ -436,6 +439,199 @@ def hole_direction(self) -> Vector3:
...
+    sits opposite the doors); ``scale.y`` is the cabin width (the two side walls); ``scale.z`` is
+    the cabin height, from floor level up to the ceiling.
+    """
+
+    WALL_THICKNESS: ClassVar[float] = 0.05
+    """
+    Thickness used for the three cabin walls and the two door leaves.
+    """
+
+    doors: DoubleDoor = field(kw_only=True, default=None)
```
AbdelrhmanBassiouny: Why double door?

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/semantic_annotations.py
```diff
@@ -436,6 +439,199 @@ def hole_direction(self) -> Vector3:
...
+            )
+            world.add_semantic_annotation(double_door)
+
+            for wall in walls:
+                world.move_branch(wall.root, floor.root)
+            for door in (door1, door2):
+                world.move_branch(door.root, floor.root)
+
+            door_travel = door_scale.y
+            door_slider_configs = (
```
AbdelrhmanBassiouny: Can you make a config dataclass for this instead of the funny tuples?

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/semantic_annotations.py
```diff
@@ -436,6 +439,199 @@ def hole_direction(self) -> Vector3:
...
+            elevator.floor = floor
+            elevator.anker_point = anker
+
+        return elevator
+
+    def open(self):
+        """
+        Opens the elevator doors
+        """
+        self.doors.door_0.mechanical_joint.root.parent_connection.position = (
```
AbdelrhmanBassiouny: I don't like how we need to write this train wreck of dots to reach something we want, can this be made easier? Looks cryptic

semantic_digital_twin/src/semantic_digital_twin/world_description/geometry.py
```diff
@@ -300,6 +300,10 @@ def to_bounding_box(self) -> BoundingBox:
     def to_np(self) -> np.ndarray:
         return np.array([self.x, self.y, self.z])
 
+    @property
+    def xy(self):
```
AbdelrhmanBassiouny, line 304: Docs and type hints

AbdelrhmanBassiouny (changes requested):

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/semantic_annotations.py
Tigul: Have you ever seen an elevator with a single door?

davidprueser: Actually, the ones sliding from right to left and vice versa could be considered a single door, but I think double door is the more general approach

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/semantic_annotations.py
Tigul: There are the specifications from @LucaKro that are for this use-case but otherwise I find making a whole dataclass for this a bit overkill

davidprueser (approved):
looks good

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/semantic_annotations.py
Tigul: Yes that is a bit a question of definition but just datastructure-wise a single door only has one door body and I need two here, hence a DoubleDoor

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/semantic_annotations.py
```diff
@@ -436,6 +448,196 @@ def hole_direction(self) -> Vector3:
...
 
 
+@dataclass(eq=False)
+class Elevator(HasRootBody):
+    """
+    An elevator in the world, consists of three walls a floor, double doors for entering and a prismatic drive that moves
+    the elevator to other floors.
+
+    ``scale`` (passed to :meth:`create_with_new_bodies_in_world`) defines the cabin as follows:
+    ``scale.x`` is the cabin depth, along the axis the double doors face and open (the back wall
```
tomsch420: the scale docstring is redundant

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/semantic_annotations.py
```diff
@@ -436,6 +448,196 @@ def hole_direction(self) -> Vector3:
...
+    sits opposite the doors); ``scale.y`` is the cabin width (the two side walls); ``scale.z`` is
+    the cabin height, from floor level up to the ceiling.
+    """
+
+    WALL_THICKNESS: ClassVar[float] = 0.05
```
tomsch420: why is this a class variable?

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/semantic_annotations.py
```diff
@@ -436,6 +448,196 @@ def hole_direction(self) -> Vector3:
...
+                scale=scale.xy,
+            )
+
+            slider = Slider.create_with_new_body_in_world(
+                name=PrefixedName(f"{name.name}_drive", name.prefix),
+                world=world,
+                active_axis=Vector3.Z(),
+            )
+
+            world.move_branch(slider.root, anker)
```
tomsch420: checkout lucas add semantic, it kind of does it already

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/semantic_annotations.py
```diff
@@ -436,6 +448,196 @@ def hole_direction(self) -> Vector3:
...
+    """
+    Floor of the elevator cabin
+    """
+
+    anker_point: Body = field(kw_only=True, default=None)
+    """
+    root point below the elevator from which the elevator is driving up
+    """
+
+    floor_positions: Dict[str, float] = field(
```
tomsch420: why is the key str and not floor?

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/semantic_annotations.py
tomsch420: code is read more often then written, so its not overkill

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/semantic_annotations.py
```diff
@@ -436,6 +448,196 @@ def hole_direction(self) -> Vector3:
...
+            elevator.floor = floor
+            elevator.anker_point = anker
+
+        return elevator
+
+    def open(self):
+        """
+        Opens the elevator doors
+        """
+        self.doors.door_0.mechanical_joint.position = self.doors.door_0.scale.y
```
tomsch420: shouldnt these use the limit of the DOFs

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/semantic_annotations.py
```diff
@@ -436,6 +448,196 @@ def hole_direction(self) -> Vector3:
...
+        self.doors.door_1.mechanical_joint.position = self.doors.door_1.scale.y
+
+    def close(self):
+        """
+        Closes the elevator doors
+        """
+        self.doors.door_0.mechanical_joint.position = 0
+        self.doors.door_1.mechanical_joint.position = 0
+
+    def add_floor(self, floor_name: str, floor_position: float):
```
tomsch420: floor name instead of floor grrrrrr

tomsch420 (changes requested):

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/semantic_annotations.py
Tigul: True but for exactly this case we have the specifications

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/semantic_annotations.py
Tigul: Because the Floor semantic annotation is a floor as on the gorund and not a floor as in the level of a house

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/semantic_annotations.py
Tigul: Also possible but I guess it doesn't really matter since both do the same thing

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/semantic_annotations.py
Tigul: That does not work in this case, from what I understand of the add semantic

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/semantic_annotations.py
Tigul: Maybe we also need a floor region to address the whole level of a building

tomsch420: but technically you are driving to the Floor which is the Floor of the floor (level)

Tigul: But a floor (level) can have multiple floors (ground) also a floor (level) can contain other rooms

semantic_digital_twin/src/semantic_digital_twin/regions/regions.py
```diff
@@ -0,0 +1,22 @@
...
+
+from semantic_digital_twin.world_description.world_entity import Region
+
+
+@dataclass(eq=False)
+class Level(Region): ...
+
+
+@dataclass(eq=False)
+class GroundFloor(Region):
```
LucaKro: Regions are world entities, GroundFloor seems to be a semantic annotation type thing, not a subclass of region. I think its a semantic annotation inheriting from the `Floor` semantic annotation

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/semantic_annotations.py
LucaKro: This whole class is not how semantic annotation design is intended. Its intended that you inherit from mixins so that you dont need to define all the parameters like doors etc, so that external interfaces can also just depend on the definition of HasDoors.

first of all, I would say that it should inherit from HasCaseAsRootBody, so that you get the basic geometry of the elevator (a hollow box with one open side) for free. then I would say iy should inherit from HasDoors, as this can mean one door, or multiple doors. I dont think doubledoors is a good fit here. then for the "drive", you should also inherit from HasMechanicalJoint (if you want to be more accurate and only allow prismatic joints, make a subclass HasSlider(HasMechanicalJoint).

walls is unnecessary now because we inherit from HasCaseAsRootBody, which generates all of those on its own, same with floor.

I dont get the point of "anker_point"? is it the root that is connected to the prismatic joint? if yes just remove it, that should just be the existing root in my opinion.

Floor positions im not sure of yet, I see the point and maybe it can stay like this for now. but ill think on it some more

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/semantic_annotations.py
```diff
@@ -436,6 +449,205 @@ def hole_direction(self) -> Vector3:
...
+
+    floor_positions: Dict[Level, float] = field(
+        kw_only=True, default_factory=dict, init=False
+    )
+    """
+    Positions of the drive which corresponds to the floor the elevator can reach.
+    """
+
+    @classmethod
+    def create_with_new_bodies_in_world(
```
LucaKro: I dont like this. for one you are defining an interface that nobody else has instead of relying on existing interfaces and workflows we use for all other semantic annotations.

What you are doing here is basically what we had in the very beginning with semantic annotation factories. We discarded that idea because, while it was okay if you just wanted to test something out without an explicit real world reference, as soon as one tried to model something from the real world (as was also necessary in suturo) this approach turned out to be completely useless. thats why I decided to rewrite the workflow for semantic annotations. you have mixins, which define the partwholerelationships like HasDoors. for each semantic annotation with a root you have create_with_new_body_in_world, with which you create just the minimal geometry, without any geometry of the partwholerelationships. then you add those afterwards using elevator.add(door), and it gets remounted automatically. this allows us to one by one build up our complex semantic annotation, while keeping us flexible enough to actually accurately model real life geometry, and while ensuring the kinematic structure is mounted in a way that works with our assumptions (for example that the doors of an elevator are the children of the elevator, with a slider or hinge in between.

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/semantic_annotations.py
LucaKro: i think it should use the dof limits idk

semantic_digital_twin/src/semantic_digital_twin/world_description/geometry.py
```diff
@@ -300,6 +300,15 @@ def to_bounding_box(self) -> BoundingBox:
...
```
LucaKro, line 304: i dont get the point of this

LucaKro: i found it, I dont think we should add properties for cases like that, but also  i think this will be unnecessary once you fix the elevator semantic annotation as i described

LucaKro (changes requested):

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/semantic_annotations.py
tomsch420: but your elevator will stop at definitely one floor (ground)

tomsch420: anyways, just dont use string here

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/semantic_annotations.py
tomsch420: i also think double doors is too static for a definition

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/semantic_annotations.py
tomsch420: they are not nescessarily the same thing

semantic_digital_twin/src/semantic_digital_twin/regions/regions.py
```diff
@@ -0,0 +1,22 @@
+from dataclasses import dataclass
+
+from semantic_digital_twin.world_description.world_entity import Region
+
+
+@dataclass(eq=False)
```
tomsch420: associate with floors, not inheriting from region

semantic_digital_twin/src/semantic_digital_twin/regions/regions.py
```diff
@@ -0,0 +1,22 @@
...
+class Level(Region): ...
+
+
+@dataclass(eq=False)
+class GroundFloor(Region):
+    pass
+
+
+@dataclass(eq=False)
+class FirstFloor(Region):
```
tomsch420: we definitely do not want these classes

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/semantic_annotations.py
```diff
@@ -490,7 +481,7 @@ class Elevator(HasRootBody):
     root point below the elevator from which the elevator is driving up
     """
 
-    floor_positions: Dict[str, float] = field(
+    floor_positions: Dict[Level, float] = field(
```
tomsch420: i think this should be known by the geometry of the Level

semantic_digital_twin/src/semantic_digital_twin/world_description/geometry.py
tomsch420: if you only use this once, remove it

tomsch420 (changes requested):

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/semantic_annotations.py
Tigul: In this case they do

semantic_digital_twin/src/semantic_digital_twin/regions/regions.py
Tigul: Why not? there is defenitly a case to have them

semantic_digital_twin/src/semantic_digital_twin/regions/regions.py
Tigul: I would argue against that, a ground floor of a building is the whole region of that building not just the semantic annotation for that

LucaKro: I do not understand what you mean. what does "just the semantic annotation" mean...

 we would also not make a subclass of body called "milk". 
instead we have a semantic annotation milk to say that this specific body  *is* a milk, just like a semantic annotation would say that this region *is* the ground floor

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/semantic_annotations.py
LucaKro: but the degree(s) of freedom will always correctly describe the behavior of the door, while in this case its mostly accidental. So please use the degree of freedom instead

semantic_digital_twin/src/semantic_digital_twin/world_description/geometry.py
LucaKro: please dont resolve comments you have not addressed without an explanation

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/semantic_annotations.py
LucaKro: yea after thinking about it more i think so as well. you can just have the association with the levels you want to reach, and then calculate the correct position from there. no need for this kind of dict

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/semantic_annotations.py
LucaKro: i think when you apply the rest of my comments it will work, because as tom said thats exactly what the add is for

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/semantic_annotations.py
LucaKro: here as well, the problem is more that this whole code does not belong here, so i think it will be more manageable once my other comments are addressed

semantic_digital_twin/src/semantic_digital_twin/world_description/geometry.py
Tigul: I want to keep it

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/semantic_annotations.py
```diff
@@ -436,6 +448,53 @@ def hole_direction(self) -> Vector3:
...
+            door.mechanical_joint.position = (
+                door.mechanical_joint.root.parent_connection.dof.limits.upper.position
+            )
+
+    def close(self):
+        """
+        Closes the elevator doors
+        """
+        for door in self.doors:
+            door.mechanical_joint.position = 0.0
```
Narenvasant: i feel this is hardcoded, maybe use same as 'Open'?

test/conftest.py
```diff
@@ -570,6 +582,80 @@ def _apartment_world_setup():
...
+def _elevator_world_setup():
+
+    world = World()
+
+    with world.modify_world():
+        world.add_body(Body(name=PrefixedName("root")))
+
+        wall_thickness = 0.05
+        scale = Scale(1, 1, 1)
+        name = PrefixedName("elevaotr")
```
Narenvasant: typo elevator

test/conftest.py
```diff
@@ -570,6 +582,80 @@ def _apartment_world_setup():
     return apartment_world
 
 
+@pytest.fixture(scope="session")
+def _elevator_world_setup():
```
Narenvasant, line 586: are you planning to use this in any setup, right now it is not used anywhere, would be cool to see this in action

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/semantic_annotations.py
```diff
@@ -1,13 +1,15 @@
 from __future__ import annotations
 
+import math
```
Narenvasant: check on some dead imports which are currently not being used

Narenvasant (changes requested):

test/conftest.py
Tigul: Yes there will be a follow up PR

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
```diff
@@ -593,7 +593,7 @@ def get_objects_of_type(
 
 
 @dataclass(eq=False)
-class HasSupportingSurface(IsStorageSpace, ABC):
+class HasSupportingSurface(IsStorageSpace):
```
AbdelrhmanBassiouny, line 596: shouldn't it be the other way around? IsStorageSpace is subclass of HasSupportingSurface?

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/semantic_annotations.py
```diff
@@ -760,6 +610,51 @@ class Bathroom(Room): ...
 class LivingRoom(Room): ...
 
 
+@dataclass(eq=False)
+class Level(SemanticAnnotation):
```
AbdelrhmanBassiouny: don't we also have a floor semantic annotation? maybe a Level has a Floor so there would be a floor attribute in this class as well? and then floor_plan would be an attribute of floor?

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/semantic_annotations.py
```diff
@@ -436,6 +444,44 @@ def hole_direction(self) -> Vector3:
...
+        for door in self.doors:
+            door.mechanical_joint.position = door.mechanical_joint.position = (
+                door.mechanical_joint.root.parent_connection.dof.limits.lower.position
+            )
+
+    def drive_to_floor(self, floor: Level):
+        """
+        Drives the elevator to the floor given
+        """
+        floor_plane_height = (floor.floor_plane[0].z + floor.floor_plane[1].z) / 2
```
LucaKro: arent both z always the same here, and then you basically do *2 and /2?

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/semantic_annotations.py
```diff
@@ -564,6 +610,51 @@ class Bathroom(Room): ...
 class LivingRoom(Room): ...
 
 
+@dataclass(eq=False)
+class Level(SemanticAnnotation):
```
LucaKro: HasRootRegion instead of semantic annotation. 

also i think this should have an association with rooms, but maybe its out of scope for this PR

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
LucaKro: no every supporting surface can be used as a storage space no? but potentially we may have storage spaces that arent supporting surfaces

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/semantic_annotations.py
LucaKro: im writing sth similar rn. I think it would have associations with rooms, which has an association with floor

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/semantic_annotations.py
```diff
@@ -564,6 +610,51 @@ class Bathroom(Room): ...
...
+    A level of a building
+    """
+
+    root: Region = field(kw_only=True)
+    """
+    The region in space which makes up the level
+    """
+
+    @property
+    def floor_plane(self) -> Tuple[Point3, Point3]:
```
LucaKro, line 619: i dont fully get why we need this. cant we just use the bounding box of the region, and then any consumer can extract the infos it needs from that? or the consumer themselves extract the bounding box of the region?

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/semantic_annotations.py
```diff
@@ -564,6 +610,51 @@ class Bathroom(Room): ...
...
+        return min_point_global, Point3(
+            max_point_global.x,
+            max_point_global.y,
+            min_z,
+            reference_frame=self._world.root,
+        )
+
+
+@dataclass(eq=False)
+class GroundFloor(Level): ...
```
LucaKro, line 641: im not sure if these are necessary imo. I feel like they are computed properties maybe instead of individual annotations, but i dont have a strong opinion on that yet

LucaKro (changes requested):

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/semantic_annotations.py
Tigul: Yes, I also think it is out-of-scope and we don't really have a use-case for this. So I call yagni

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/semantic_annotations.py
Tigul: Coudl also be, but we would need the concept of a building to compute that and as of now this is also yagni

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/semantic_annotations.py
Tigul: I like it the way it is

LucaKro (approved):

AbdelrhmanBassiouny (approved):

Narenvasant (approved):

PR #459, opened 2026-07-20
Tool-based actions: path-driven cutting, mixing, wiping and pouring

sunava:
## Summary

This PR reworks the tool-based composite actions (cutting, mixing, wiping, pouring) from hard-coded pose sequences into a clean, path-driven architecture. Tool motions are now described as sampled geometric tool paths that are executed as aligned TCP waypoint trajectories, with the tool kept correctly oriented towards its target throughout the motion.
This is the first iteration without the big experiment. 
### coraplex — actions & motions
- **New `tool_paths` module** (`robot_plans/actions/composite/tool_paths.py`): builds parameterized tool paths from object geometry (bounding boxes / mesh top-surface hulls). Provides path primitives (approach, descend, saw, spiral, stir, shear, raster, sweep) plus builders for container mixing paths, cutting paths and surface wiping paths, including constraints that clamp the motion to the container cylinder or the surface's convex hull.
- **Rewritten `tool_based` actions** (`robot_plans/actions/composite/tool_based.py`):
  - `FullBodyControlledAction` base: temporarily enables full-body control on mobile-base robots so the base supports the arm motion, and restores the previous state afterwards.
  - `ToolMotionAction` base: builds a tool path, samples it into world-frame waypoints and executes it as a single aligned waypoint motion.
  - `MixingAction`: spiral or timed stir pattern sized to the container.
  - `CuttingAction`: slice / saw / halving techniques with configurable slice thickness, number of cuts and a `SlicingPriority` to resolve conflicts between the two.
  - `WipingAction`: wipes a surface body or a patch around a target pose (wipe / shear / spread), with a distance tolerance to accept nearly-finished motions.
  - `PouringAction`: pours from a held container into a target container, computing the pour pose from the source container's mouth height and an approach direction snapped to the target's nearest local axis.
- **New motion `MoveTCPWaypointsAlignedMotion`** (`robot_plans/motions/gripper.py`): follows a Cartesian position trajectory through waypoints while holding `AlignPlanes` constraints for the given alignment pairs; supports full-body control, gripper-only collision rules and an upright-torso task for Justin.
- **New enums** (`CuttingTechnique`, `SlicingPriority`, `ToolPathSegmentKind`, `WipingTechnique`, `MixingPattern`) and **new exceptions** (`MissingWaypoints`, `WipingTargetMissing`, `MissingToolFrame`).
- `BaseMotion` gained a helper to build collision rules that only allow gripper–environment collisions.

### semantic_digital_twin — tool semantics
[...]

PR #464, opened 2026-07-21
Underspecified tool-action demos and a scaled-up tool-based action epxeriment

sunava:
### Summary

This PR adds two things on top of the tool-based actions from #459:

1. **Underspecified demos** — one per tool action (cutting, mixing, pouring, wiping), showing plans where parameters are left as ellipses and resolved by the probabilistic backend, bounded only by `where`-conditions.
2. **A reproducible experiment campaign** that measures how reliably the PR2 performs the four tool-based actions on randomly generated, cluttered scenes in the apartment.

It also contains two small upstream fixes in `semantic_digital_twin` that the experiment surfaced.

### Underspecified demos (`experiments/.../tool_based_actions/underspecified_demo/`)

Each demo builds the same plan shape as the corresponding simple demo, but leaves parameters open e.g. the base pose (`x=..., y=...`), `keep_joint_states`, the cutting `technique`, or the `pour_height` — and constrains them with `where`-conditions instead of concrete values. Resolution happens through `ProbabilisticBackend`. 

### Tool-based action experiment (`experiments/.../tool_based_actions/experiment/`)

An experiment is a grid of trials (task × seed), run by `run_experiment.py` with one subprocess per trial, wall-clock timeout, and resume support: trials with recorded results are skipped, so a killed campaign continues where it left off. See the package `README.md` for details.

**Scene generation** (`scene.py`) is seeded and produces dense, realistic scenes:
- target count from surface area (`targets_per_square_meter`, clamped, best-effort placement down to a minimum),
- per-target random uniform scale, with the footprint radius measured from the mesh so placements never overlap each other or hang off surface edges,
- placements keep clear of the bounding boxes of all other world bodies (nothing spawns inside sinks or furniture), and surfaces above a height cap are excluded.

**Trial execution** (`single_trial.py`):
- the base pose per target is *underspecified* — a variable over a costmap location (ring at arm's reach ∧ collision-free occupancy); the tool actions themselves are fully specified,
- `full_body_motion` (default on): the base drives along during tool motions, so targets at the edge of reach stay achievable,
- `collision_avoidance` (default on): motions avoid the environment, with the acting arm's manipulator *including the mounted tool* exempted so the tool may touch its target; real violations abort the action and are recorded as failures,
- the robot is parked once per trial *without* collision avoidance first, because it spawns with its outstretched arms in contact with the kitchen island,
- the currently approached target is dyed blue in RViz and restored afterwards; the world, tf tree, and closest-point collision results are published for RViz.

**Results** (`results.py`) go to a JSON-lines file, one record per target (position, yaw, scale, surface, success, duration, failure reason). Loading validates the schema and fails with a clear `IncompatibleResultRecord` on records from an older version.

### Upstream fixes (`semantic_digital_twin`)

- **`CollisionManager` serialization crashed with runtime consumers:** registering a collision consumer holding a ROS node (e.g. the collision visualization publisher) broke every subsequent `modify_world` with `BrokenWorldModificationHistoryError`, because the generic serializer walked the `collision_consumers` field — which `_from_json` never restores anyway. The field is now `init=False` and thereby excluded from serialization (same convention as `collision_matrix`).
- **`Mesh.scale` assignments were silently ignored:** `Mesh.mesh` is a `cached_property` that bakes the scale in at load time, so setting `shape.scale` after parsing changed the RViz marker but not the geometry. Cutting paths on scaled breads were computed on the unscaled bounding box and visibly started mid-object. Assigning `scale` now invalidates the cached mesh, so bounding boxes, tool paths, and collision geometry match what RViz shows.

### Testing

- `test/experiments_test/test_tool_based_experiment.py`: 30 tests covering the sampler (determinism, footprint clearance, obstacle rejection, height cap, best-effort placement, density counts), obstacle discovery, footprint measurement, scaled spawning, result recording/resume, schema validation, configuration round-trip, and the target highlight.
- `test/semantic_digital_twin_test/...`: regression tests for both upstream fixes.
- End-to-end verified in simulation: cutting trials succeed with full-body motion, collision avoidance, and highlighting active.


experiments/src/experiments/tool_based_actions/experiment/configuration.py
```diff
@@ -0,0 +1,302 @@
...
+
+    spawn_height_offset: float = 0.05
+    """
+    Height in meters above a surface top at which targets are spawned.
+    """
+
+    results_file: Path = field(
+        default_factory=lambda: Path(__file__).parent
+        / "records"
+        / "tool_based_experiment_results.jsonl"
```
Narenvasant, line 140: does the json file also needs to be committed?

Narenvasant (approved):
Nice!!!

experiments/src/experiments/tool_based_actions/experiment/records/tool_based_experiment_results.jsonl
AbdelrhmanBassiouny, line 1: Do you want that committed?

experiments/src/experiments/tool_based_actions/experiment/configuration.py
```diff
@@ -0,0 +1,302 @@
...
+
+import enum
+import json
+from dataclasses import asdict, dataclass, field
+from pathlib import Path
+
+from typing_extensions import List, Tuple
+
+
+class ToolBasedTask(enum.Enum):
```
AbdelrhmanBassiouny: You can get this information by getting all subclasses of ToolBasedTasks, which you already have. There's a way to get subclasses of a class programmatically; check krrood utils or class_diagram utils, or ask Claude to make this util if missing.

AbdelrhmanBassiouny: And if you want to indicate which task it is, just directly use the task class itself instead of this enum because this is a duplicate of the classes themselves.

experiments/src/experiments/tool_based_actions/experiment/configuration.py
AbdelrhmanBassiouny: I did the same comment above haha

experiments/src/experiments/tool_based_actions/experiment/configuration.py
```diff
@@ -0,0 +1,302 @@
...
+    results_file: Path = field(
+        default_factory=lambda: Path(__file__).parent
+        / "records"
+        / "tool_based_experiment_results.jsonl"
+    )
+    """
+    File the trial results are appended to, one JSON object per line.
+    """
+
+    trial_timeout: float = 3600.0
```
AbdelrhmanBassiouny: use timedelta instead of float

experiments/src/experiments/tool_based_actions/experiment/configuration.py
```diff
@@ -0,0 +1,302 @@
...
+                task=task,
+                seed=seed,
+                robot_name=self.robot_name,
+                environment_name=self.environment_name,
+            )
+            for task in self.tasks
+            for seed in self.seeds
+        ]
+
+    def to_json(self) -> str:
```
AbdelrhmanBassiouny: Use SubClassJSONSerializer if you want to do that. Also, maybe your class is automatically serializable such that you don't need all of that?

experiments/src/experiments/tool_based_actions/experiment/records/tool_based_experiment_results.jsonl
tomsch420, line 1: looks like a file we dont want. Also if this is neems, why not use the db?

experiments/src/experiments/tool_based_actions/experiment/configuration.py
sunava: not really :D you are right

experiments/src/experiments/tool_based_actions/experiment/configuration.py
```diff
@@ -0,0 +1,302 @@
...
+    """
+
+    CUTTING = "cutting"
+    MIXING = "mixing"
+    POURING = "pouring"
+    WIPING = "wiping"
+
+
+@dataclass(frozen=True)
+class SpawnRegion:
```
tomsch420: is this like a 2d bounding box? I thought we have something for that already

experiments/src/experiments/tool_based_actions/experiment/configuration.py
```diff
@@ -0,0 +1,302 @@
...
+    """
+    The tasks to run.
+    """
+
+    seeds: Tuple[int, ...] = (910001, 910002, 910003)
+    """
+    The seeds to run every task with.
+    """
+
+    robot_name: str = "pr2"
```
tomsch420: you can use the semantic annotation here

experiments/src/experiments/tool_based_actions/experiment/configuration.py
tomsch420: use subclass json serializer here

experiments/src/experiments/tool_based_actions/experiment/configuration.py
tomsch420, line 1: if some of these are like experiment tables, use the superclass

experiments/src/experiments/tool_based_actions/experiment/README.md
```diff
@@ -0,0 +1,139 @@
...
+  stay achievable.
+- `collision_avoidance` (default True): motions avoid collisions with the
+  environment. The tool motions exempt the acting arm's manipulator **including
+  the mounted tool**, so the tool can touch its target. A real collision of any
+  other body part aborts the action and is recorded as a failure
+  (`CollisionViolatedError`).
+
+## Results
+
+Results are appended to `records/tool_based_experiment_results.jsonl`, one JSON
```
tomsch420: ORMatic >:(

experiments/src/experiments/tool_based_actions/experiment/configuration.py
LucaKro: we have sth ike this already i think. if you use semantic annotations, if your supporting surface (like a table) correctly inherits from HasSupportingSurface you can do table.calculate_supporting_surface() and then table.supporting_surface will be populated with a region representing the surface, which you can also randomly sample from

experiments/src/experiments/tool_based_actions/experiment/README.md
tomsch420, line 1: is this readme really nescessary? Why not just write a parameterizable main like the other experiments?

experiments/src/experiments/tool_based_actions/experiment/results.py
```diff
@@ -0,0 +1,196 @@
...
+
+from typing_extensions import Any, Dict, List, Optional, Set
+
+from experiments.tool_based_actions.experiment.configuration import (
+    ToolBasedTask,
+    TrialSpecification,
+)
+
+
+class IncompatibleResultRecord(Exception):
```
tomsch420: dataclass exception

experiments/src/experiments/tool_based_actions/experiment/results.py
```diff
@@ -0,0 +1,196 @@
...
+    """
+    Wall-clock duration of the action in seconds.
+    """
+
+    failure_reason: Optional[str] = None
+    """
+    Compact description of the failure, or None on success.
+    """
+
+    def to_json_line(self) -> str:
```
tomsch420: subclass json serializer

experiments/src/experiments/tool_based_actions/experiment/results.py
```diff
@@ -0,0 +1,196 @@
...
+            if field.default is MISSING and field.default_factory is MISSING
+        }
+        missing_fields = sorted(required_field_names - record.keys())
+        unexpected_fields = sorted(record.keys() - field_names)
+        if missing_fields or unexpected_fields:
+            raise IncompatibleResultRecord(missing_fields, unexpected_fields)
+
+
+@dataclass
+class ResultRecorder:
```
tomsch420, line 196: not needed if you use ormatic

experiments/src/experiments/tool_based_actions/experiment/run_experiment.py
```diff
@@ -0,0 +1,143 @@
...
+            logger.info(
+                "%s: %d/%d targets succeeded",
+                task_name,
+                successes_by_task.get(task_name, 0),
+                totals_by_task[task_name],
+            )
+        logger.info("Results file: %s", self.configuration.results_file)
+
+
+def main() -> None:
```
tomsch420, line 186: i like this class, but why does this now have a main?

experiments/src/experiments/tool_based_actions/experiment/scene.py
```diff
@@ -0,0 +1,531 @@
...
+import random
+from dataclasses import dataclass, field
+
+from semantic_digital_twin.world import World
+from typing_extensions import ClassVar, List, Optional, Set, Tuple
+
+from experiments.tool_based_actions.experiment.configuration import SpawnRegion
+
+
+class MissingSpawnSurfaces(Exception):
```
tomsch420: dataclass exception

experiments/src/experiments/tool_based_actions/experiment/scene.py
```diff
@@ -0,0 +1,531 @@
...
+
+    def largest_radius(self) -> float:
+        """
+        :return: The inflated footprint radius in meters at the largest scale.
+        """
+        return self.radius_for_scale(max(self.scale_choices))
+
+
+@dataclass(frozen=True)
+class ObstacleBox:
```
tomsch420: whats this if not a bounding box?

experiments/src/experiments/tool_based_actions/experiment/scene.py
```diff
@@ -0,0 +1,531 @@
...
+            return False
+        if z >= self.maximum_z - self.top_epsilon:
+            return False
+        return (
+            self.minimum_x - radius <= x <= self.maximum_x + radius
+            and self.minimum_y - radius <= y <= self.maximum_y + radius
+        )
+
+
+def discover_obstacles(
```
tomsch420: looks like GCS

experiments/src/experiments/tool_based_actions/experiment/scene.py
```diff
@@ -0,0 +1,531 @@
...
+                ),
+            )
+        )
+    if not surfaces:
+        raise MissingSpawnSurfaces(surface_names)
+    return surfaces
+
+
+@dataclass(frozen=True)
+class TargetPlacement:
```
tomsch420, line 132: looks like the 2d pose we have already

experiments/src/experiments/tool_based_actions/experiment/scene.py
```diff
@@ -0,0 +1,531 @@
...
+        :param targets_per_square_meter: Desired target density on the surfaces.
+        :param minimum: Smallest allowed number of targets.
+        :param maximum: Largest allowed number of targets.
+        :return: The density-based number of targets, clamped to
+            ``[minimum, maximum]``.
+        """
+        area = sum(surface.region.area() for surface in self._usable_surfaces())
+        return max(minimum, min(maximum, round(area * targets_per_square_meter)))
+
+    def sample_placements(
```
tomsch420: use eql with generative backend here

experiments/src/experiments/tool_based_actions/experiment/single_trial.py
```diff
@@ -0,0 +1,351 @@
...
+        * 0.66,
+        world=context.world,
+        origin=target_pose,
+    )
+    costmap = OccupancyCostmap.default_map(context, target_pose) & ring
+    return Location(context, target_pose, costmap, [])
+
+
+@dataclass
+class TrialRunner:
```
tomsch420, line 99: sounds like an experiment runner from before

experiments/src/experiments/tool_based_actions/experiment/task_definitions.py
tomsch420, line 1: this looks very duplicated with the existing actions, why is this here?

experiments/src/experiments/tool_based_actions/underspecified_demo/demo_cutting.py
```diff
@@ -0,0 +1,95 @@
...
+
+    knife = CuttingKnife(root=knife_body)
+    with world.modify_world():
+        world.add_semantic_annotations([Bread(root=bread_body), knife])
+
+    context.evaluate_conditions = False
+
+    navigate = build_underspecified_navigation(world)
+
+    cutting = a(CuttingAction)(
```
tomsch420, line 71: guccigang

experiments/src/experiments/tool_based_actions/underspecified_demo/demo_mixing.py
```diff
@@ -0,0 +1,108 @@
...
+
+    navigate = build_underspecified_navigation(world)
+
+    mixing = a(MixingAction)(
+        container=bowl_body,
+        arm=Arms.RIGHT,
+        tool=whisk,
+        mix_duration=...,
+    )
+    mixing.where(
```
tomsch420, line 86: epic

experiments/src/experiments/tool_based_actions/underspecified_demo/demo_pouring.py
```diff
@@ -0,0 +1,113 @@
...
+
+    navigate = build_underspecified_navigation(world)
+
+    pouring = a(PouringAction)(
+        target_container=bowl_body,
+        source_container=cup,
+        arm=Arms.RIGHT,
+        pour_height=...,
+    )
+    pouring.where(
```
tomsch420, line 91: epic

experiments/src/experiments/tool_based_actions/underspecified_demo/demo_setup.py
```diff
@@ -0,0 +1,110 @@
...
+from coraplex.robot_plans.actions.core.navigation import NavigateAction
+
+from experiments.tool_based_actions.simple_demo.demo_world import (
+    TARGET_POSITION_XYZ,
+    parse_object,
+)
+
+
+@dataclass(frozen=True)
+class SampledRegion:
```
tomsch420: i have seen this class like 3 times in this pr already

experiments/src/experiments/tool_based_actions/underspecified_demo/demo_setup.py
```diff
@@ -0,0 +1,110 @@
...
+    Lower Y bound of the region in the world frame.
+    """
+
+    maximum_y: float
+    """
+    Upper Y bound of the region in the world frame.
+    """
+
+
+BASE_POSE_REGION = SampledRegion(
```
tomsch420: globals grrr

experiments/src/experiments/tool_based_actions/underspecified_demo/test_demos.py
```diff
@@ -0,0 +1,37 @@
...
+run is a full simulation taking minutes.
+"""
+
+import subprocess
+import sys
+
+import pytest
+
+DEMO_MODULES = (
+    "experiments.tool_based_actions.underspecified_demo.demo_cutting",
```
tomsch420: no strings. Also experiments have their own tests

tomsch420 (changes requested):
I think general argparse is your friend here.

I love that you are using a/an here

experiments/src/experiments/tool_based_actions/experiment/configuration.py
```diff
@@ -0,0 +1,302 @@
...
+        :param x: X coordinate in the world frame.
+        :param y: Y coordinate in the world frame.
+        :return: True if the point lies inside the region.
+        """
+        return (
+            self.minimum_x <= x <= self.maximum_x
+            and self.minimum_y <= y <= self.maximum_y
+        )
+
+    def grid_capacity(self, clearance: float) -> int:
```
LucaKro: hmm i find this to be a bit inflexible, because this depends on the size of your objects

experiments/src/experiments/tool_based_actions/experiment/configuration.py
```diff
@@ -0,0 +1,302 @@
...
+            minimum_y=self.minimum_y + margin,
+            maximum_y=self.maximum_y - margin,
+            height=self.height,
+        )
+
+    def is_empty(self) -> bool:
+        """
+        :return: True if the region contains no points.
+        """
+        return self.maximum_x <= self.minimum_x or self.maximum_y <= self.minimum_y
```
LucaKro: i feel like this should be a post init check, a region that violates this doesnt make sense imo

experiments/src/experiments/tool_based_actions/experiment/scene.py
```diff
@@ -0,0 +1,531 @@
...
+    def __init__(self, surfaces: List[SpawnSurface], clearance: float, count: int):
+        surface_names = [surface.name for surface in surfaces]
+        super().__init__(
+            f"Could not place {count} targets with clearance {clearance} m on the "
+            f"surfaces {surface_names}."
+        )
+
+
+@dataclass(frozen=True)
+class SpawnSurface:
```
LucaKro: i think this is basically thre HasSupportingSurface semantic annotation mixin

experiments/src/experiments/tool_based_actions/experiment/scene.py
```diff
@@ -0,0 +1,531 @@
...
+    """
+
+    region: SpawnRegion
+    """
+    The spawnable rectangle on top of the surface, in the world frame.
+    """
+
+
+@dataclass(frozen=True)
+class ObjectFootprint:
```
LucaKro: i dont understand what this is for

experiments/src/experiments/tool_based_actions/experiment/scene.py
LucaKro: this is  3d tho, so id say just a pose no? with less  rotational dimensions, but still, we cant make a new class for all of those cases

experiments/src/experiments/tool_based_actions/experiment/scene.py
LucaKro: or just sample from the supporting surface with functionality we already have

experiments/src/experiments/tool_based_actions/experiment/scene.py
sunava: `Pose2D` is a symbolic (casadi) spatial type — its `x`/`y`/`yaw` are `sm.Scalar` bound to a `reference_frame`, not plain floats, and it is not value-equal/serializable — so it cannot back the world-free, deterministic sampler and its unit tests. Kept `TargetPlacement` as a lightweight float record.

experiments/src/experiments/tool_based_actions/experiment/scene.py
sunava: It wraps a `BoundingBox` now (`ObstacleBox.from_bounding_box(...)`); the remainder is placement semantics — `vertical_margin`/`top_epsilon` and `blocks()`, i.e. "resting on the top face does not count as a collision". Same reason as `SpawnRegion` for not storing the `BoundingBox` itself: it is `eq=False` and `contains()` needs a live world.

semantic_digital_twin/src/semantic_digital_twin/collision_checking/collision_manager.py
```diff
@@ -147,9 +147,14 @@ class CollisionManager(ModelChangeCallback):
     between two bodies.
     """
 
-    collision_consumers: list[CollisionConsumer] = field(default_factory=list)
+    collision_consumers: list[CollisionConsumer] = field(
+        default_factory=list, init=False
```
LucaKro, line 151: @ichumuh dont we need to synchronize collision consumers?

experiments/src/experiments/tool_based_actions/underspecified_demo/demo_pouring.py
sunava: <3

LucaKro (changes requested):

experiments/src/experiments/tool_based_actions/experiment/configuration.py
```diff
@@ -0,0 +1,347 @@
...
+            TrialSpecification(
+                task=task,
+                seed=seed,
+                environment_name=self.environment_name,
+            )
+            for task in self.tasks
+            for seed in self.seeds
+        ]
+
+    def to_json(self) -> Dict[str, Any]:
```
AbdelrhmanBassiouny: same json comment

experiments/src/experiments/tool_based_actions/experiment/results.py
```diff
@@ -0,0 +1,293 @@
...
+    """
+    Name of the surface the target was spawned on.
+    """
+
+    success: bool
+    """
+    True if the action completed without an error.
+    """
+
+    duration: float
```
AbdelrhmanBassiouny: Use a proper duration type like timedelta or similar

experiments/src/experiments/tool_based_actions/experiment/results.py
```diff
@@ -0,0 +1,293 @@
...
+    """
+    Name of the environment the action ran in.
+    """
+
+    target_name: str
+    """
+    Name of the target the action acted on.
+    """
+
+    target_x: float
```
AbdelrhmanBassiouny: Why aren't x,y, and yaw combined in one dataclass like PlanePose or Pose2D, for example?

experiments/src/experiments/tool_based_actions/experiment/results.py
```diff
@@ -0,0 +1,293 @@
...
+    """
+    Wall-clock duration of the action in seconds.
+    """
+
+    failure_reason: Optional[str] = None
+    """
+    Compact description of the failure, or None on success.
+    """
+
+    def to_json(self) -> Dict[str, Any]:
```
AbdelrhmanBassiouny: json

experiments/src/experiments/tool_based_actions/experiment/results.py
```diff
@@ -0,0 +1,293 @@
...
+    """
+    True if the action completed without an error.
+    """
+
+    duration: float
+    """
+    Wall-clock duration of the action in seconds.
+    """
+
+    failure_reason: Optional[str] = None
```
AbdelrhmanBassiouny: Can this be an Exception type or an Enum?

experiments/src/experiments/tool_based_actions/experiment/scene.py
```diff
@@ -0,0 +1,567 @@
...
+from krrood.exceptions import DataclassException
+from semantic_digital_twin.world import World
+from semantic_digital_twin.world_description.geometry import BoundingBox
+from typing_extensions import ClassVar, List, Optional, Set, Tuple
+
+from experiments.tool_based_actions.experiment.configuration import SpawnRegion
+
+
+@dataclass
+class MissingSpawnSurfaces(DataclassException):
```
AbdelrhmanBassiouny: Can you move all exceptions to an exceptions.py module?

experiments/src/experiments/tool_based_actions/experiment/scene.py
```diff
@@ -0,0 +1,567 @@
...
+    Band in meters below an obstacle's top within which a placement counts as resting on
+    the obstacle instead of colliding with it.
+    """
+
+    name: str
+    """
+    Name of the obstacle body in the world.
+    """
+
+    minimum_x: float
```
AbdelrhmanBassiouny: looks like bounding box parameters (minx,y,z, max x,y,z) you can just use a BoundingBox instead of all these fields.

experiments/src/experiments/tool_based_actions/experiment/scene.py
```diff
@@ -0,0 +1,567 @@
...
+            return False
+        if z >= self.maximum_z - self.top_epsilon:
+            return False
+        return (
+            self.minimum_x - radius <= x <= self.maximum_x + radius
+            and self.minimum_y - radius <= y <= self.maximum_y + radius
+        )
+
+    @classmethod
+    def from_bounding_box(cls, name: str, bounding_box: BoundingBox) -> ObstacleBox:
```
AbdelrhmanBassiouny: Why not have the bounding box as a field instead of duplicating similar fields in this class?

experiments/src/experiments/tool_based_actions/experiment/scene.py
```diff
@@ -0,0 +1,567 @@
...
+    """
+    Unique name of the target within its trial.
+    """
+
+    surface_name: str
+    """
+    Name of the surface the target is spawned on.
+    """
+
+    x: float
```
AbdelrhmanBassiouny: same comment on Pose2D or PlanarPose

AbdelrhmanBassiouny: even if you make your own PlanarPose it would be better than recreating these fields in multiple classes

experiments/src/experiments/tool_based_actions/experiment/task_definitions.py
```diff
@@ -0,0 +1,381 @@
...
+            pointer_stride=self.pointer_stride,
+        )
+
+    def target_footprint(
+        self, scale_choices: Tuple[float, ...], safety_factor: float
+    ) -> ObjectFootprint:
+        return ObjectFootprint.point()
+
+
+def definition_for_task(
```
AbdelrhmanBassiouny: This is a task to definition, not definition to task

AbdelrhmanBassiouny: And why not directly take the definition you want to create instead of the task?

experiments/src/experiments/tool_based_actions/experiment/visualization.py
```diff
@@ -0,0 +1,93 @@
...
+    The body to highlight, or None for targets that are pure poses (e.g. wiping
+    patches), which are left untouched.
+    """
+
+    color: Color = field(default_factory=lambda: Color(R=0.1, G=0.4, B=1.0))
+    """
+    The color the target is dyed with while it is highlighted.
+    """
+
+    _original_colors: List[Tuple[Shape, Color]] = field(
```
AbdelrhmanBassiouny: Make a dataclass with two fields, one for shape and one for color, instead of the funny tuple.

experiments/src/experiments/tool_based_actions/experiment/visualization.py
```diff
@@ -0,0 +1,93 @@
...
+        init=False, default_factory=list
+    )
+    """
+    The shape colors to restore when the highlight ends.
+    """
+
+    def __enter__(self) -> TargetHighlight:
```
AbdelrhmanBassiouny, line 93: docs

experiments/src/experiments/tool_based_actions/experiment/visualization.py
```diff
@@ -0,0 +1,93 @@
...
+        if self.body is None:
+            return self
+        with self.world.modify_world():
+            for shape in self.body.visual.shapes:
+                self._original_colors.append((shape, shape.color))
+                shape.color = self.color
+        return self
+
+    def __exit__(self, exception_type, exception_value, exception_traceback) -> None:
```
AbdelrhmanBassiouny, line 109: docs

experiments/src/experiments/tool_based_actions/simple_demo/demo_world.py
```diff
@@ -90,6 +90,39 @@ def parse_object(stl_file_name: str, color: Optional[Color] = None) -> World:
     return object_world
 
 
+def spawn_mesh_body(
+    world: World,
+    stl_file_name: str,
```
AbdelrhmanBassiouny: Why necessarily STL?

experiments/src/experiments/tool_based_actions/underspecified_demo/demo_cutting.py
```diff
@@ -0,0 +1,95 @@
...
+from coraplex.testing import attach_tool, setup_world, start_visualization
+
+
+def main() -> None:
+    """
+    Build the demo world and run the underspecified plan on the simulated robot.
+    """
+    # Importing the ORM interface registers the data access objects the
+    # probabilistic backend needs to condition on the given literal values.
+    import coraplex.orm.ormatic_interface
```
AbdelrhmanBassiouny: Move import to the top of the module

experiments/src/experiments/tool_based_actions/underspecified_demo/demo_cutting.py
AbdelrhmanBassiouny: How do you like the new interface? @sunava, you helped shape it

AbdelrhmanBassiouny (changes requested):

experiments/src/experiments/tool_based_actions/experiment/results.py
```diff
@@ -2,51 +2,79 @@
...
+)
+from experiments.tool_based_actions.experiment.scene import PlanarPose
+from experiments.tool_based_actions.experiment.task_definitions import (
+    ToolTaskDefinition,
 )
 
 
-class IncompatibleResultRecord(Exception):
+@dataclass(frozen=True)
+class FailureDescription:
```
tomsch420, line 35: dataclass exception are ormaticable if thats a problem here

experiments/src/experiments/tool_based_actions/experiment/results.py
```diff
@@ -139,18 +140,58 @@ def _validate_record_matches_schema(cls, record: Dict[str, Any]) -> None:
...
+        )
+        totals = Counter(result.task.task_name() for result in results)
+        return [
+            cls(task=name, successes=successes.get(name, 0), total=totals[name])
+            for name in sorted(totals)
+        ]
+
+
 @dataclass
 class ResultRecorder:
```
tomsch420, line 196: soooo, ormatic when?

experiments/src/experiments/tool_based_actions/experiment/scene.py
```diff
@@ -13,249 +15,117 @@
...
-    top_epsilon: ClassVar[float] = 0.01
-    """
-    Band in meters below an obstacle's top within which a placement counts as resting on
-    the obstacle instead of colliding with it.
-    """
-
-    name: str
-    """
-    Name of the obstacle body in the world.
+class PlanarPose:
```
tomsch420, line 100: this class already exists

tomsch420: its called Pose2D

experiments/src/experiments/tool_based_actions/experiment/scene.py
```diff
@@ -271,103 +141,131 @@ class TargetPlacement:
...
+                continue
+            surfaces.append(surface)
+    if not surfaces:
+        raise MissingSpawnSurfaces(tuple(surface_names))
+    return surfaces
+
 
-    footprint: ObjectFootprint = field(default_factory=ObjectFootprint.point)
+@dataclass
+class SceneSpawner:
```
tomsch420, line 222: @davidprueser look at this and perhaps this is another RSPN application?

test/experiments_test/test_underspecified_demos.py
tomsch420, line 1: do you really want to run all demos in the tests? This sounds like it takes a while

AbdelrhmanBassiouny (approved):

tomsch420 (changes requested):

semantic_digital_twin/src/semantic_digital_twin/collision_checking/collision_manager.py
ichumuh: no, at least the ones we have should not be synchronized.

experiments/src/experiments/tool_based_actions/experiment/README.md
LucaKro: can you write the reasoning for keeping it please?

experiments/src/experiments/tool_based_actions/experiment/scene.py
```diff
@@ -0,0 +1,429 @@
...
+)
+
+if TYPE_CHECKING:
+    from experiments.tool_based_actions.experiment.task_definitions import (
+        ToolTaskDefinition,
+    )
+
+
+@dataclass(frozen=True)
+class SpawnRegion:
```
LucaKro, line 47: why not just use Regions here?

experiments/src/experiments/tool_based_actions/experiment/scene.py
LucaKro: im pretty sure thats not correct. you can instantiate poses without reference frames, and spatial types are serializable

LucaKro (changes requested):

PR #465, opened 2026-07-21
Import joint dynamics and link inertials from URDF

AbdElRahmanFarhan:
## Problem

`URDFParser` never read the `<dynamics>` or `<inertial>` elements, so everything it imported reached physics with placeholder values:

- every joint arrived as `JointDynamics(armature=0.0, dry_friction=0.0, damping=0.0)` — frictionless and undamped;
- every body arrived at `mass=1.0` with a unit inertia tensor.

This is an importer gap only: `Connection.dynamics` and `Body.inertial` already exist, `MujocoBuilder` already writes both into the MJCF, and `adapters/mjcf.py` already reads them back on the MJCF→world path.

## Change

- `parse_dynamics` reads `<dynamics damping=... friction=.../>` into `JointDynamics`. URDF has no armature (that is a simulator-side rotor inertia), so it stays at its default; undeclared attributes stay zero.
- `parse_inertial` reads `<inertial>` into `Inertial`. URDF gives the tensor in the inertial frame, so it is rotated into the link frame, which is the frame `Inertial` documents. Links that declare no `<inertial>` keep the `Body` default, so nothing regresses for descriptions without inertia.

## Tests

Five tests in `test_urdf.py`, on models the suite already uses — no new assets:

- Tiago — `arm_left_1_joint` declares `damping="40.0" friction="1.0"`, so it covers both attributes;
- PR2 — a movable joint that declares no `<dynamics>` stays at zero, and `l_elbow_flex_link` comes through with its real mass, centre of mass and off-diagonal inertia products;
- the table model — a description with no `<inertial>` at all keeps the `Body` default;
- the frame rotation is covered by calling `parse_inertial` with a `urdf_parser_py` link whose inertial frame is rotated 90° about z, since no description in the resources declares a rotated inertial frame.

### DIFF
--- a/semantic_digital_twin/src/semantic_digital_twin/adapters/urdf.py
+++ b/semantic_digital_twin/src/semantic_digital_twin/adapters/urdf.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 from dataclasses import dataclass, field
 
-from scipy.spatial.transform import Rotation
 from typing_extensions import Dict, Optional, Tuple, Union, List
 from urdf_parser_py import urdf as urdfpy
 from xacro import process_file
@@ -17,6 +16,7 @@ from semantic_digital_twin.spatial_types.derivatives import Derivatives, Derivat
 from semantic_digital_twin.spatial_types.spatial_types import (
     HomogeneousTransformationMatrix,
     Point3,
+    RotationMatrix,
     Vector3,
 )
 from semantic_digital_twin.utils import (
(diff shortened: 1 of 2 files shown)


tomsch420 (approved):

semantic_digital_twin/src/semantic_digital_twin/adapters/urdf.py
```diff
@@ -293,8 +322,51 @@ def parse_link(self, link: urdfpy.Link, parent_frame: PrefixedName) -> Body:
...
+        urdf_inertia = link.inertial.inertia
+        inertia_in_inertial_frame = InertiaTensor.from_values(
+            ixx=urdf_inertia.ixx,
+            iyy=urdf_inertia.iyy,
+            izz=urdf_inertia.izz,
+            ixy=urdf_inertia.ixy,
+            ixz=urdf_inertia.ixz,
+            iyz=urdf_inertia.iyz,
+        )
+        link_R_inertial = Rotation.from_euler("xyz", roll_pitch_yaw).as_matrix()
```
davidprueser: There is a RotationMatrix in SemDT you could use

davidprueser (changes requested):
Looks cool :)

davidprueser (approved):

PR #468, opened 2026-07-21
Fix flaky zoo-dataset tests when the UCI ML repo server is unavailable

AbdelrhmanBassiouny:
get_dataset()/load_zoo_dataset() now degrade gracefully (return empty) on DatasetNotFoundError, not just ConnectionError, so a transient UCI ML repo outage skips affected tests instead of breaking collection. Full detail: https://github.com/AbdelrhmanBassiouny/cognitive_robot_abstract_machine/pull/95


test/krrood_test/test_ripple_down_rules/test_dataset_availability.py
```diff
@@ -0,0 +1,35 @@
+import unittest
+from unittest.mock import patch
+
+from ucimlrepo import DatasetNotFoundError
+
+from . import datasets
+
+
+class DatasetServerUnavailableTestCase(unittest.TestCase):
```
tomsch420: grrr unittest

tomsch420 (changes requested):

test/krrood_test/test_ripple_down_rules/test_dataset_availability.py
AbdelrhmanBassiouny: Refactored to use unittest.

tomsch420 (approved):

davidprueser (approved):

PR #472, opened 2026-07-24
ripple_down_rules reduction (reconciled to code_generation)

AbdelrhmanBassiouny:
Mostly just removals.
Fork PR: https://github.com/AbdelrhmanBassiouny/cognitive_robot_abstract_machine/pull/53


tomsch420 (approved):

test/krrood_test/test_ripple_down_rules/test_expert_answers/mcrdr_expert_answers_fit_no_targets.py
```diff
@@ -829,6 +854,7 @@ def animal_species_of_type_species(case: DataFrame) -> Species:
 '===New Answer==='
 
 
+from types import NoneType
```
ichumuh: why are you adding these imports everywhere without using them?

LucaKro (approved):

test/krrood_test/test_ripple_down_rules/test_expert_answers/mcrdr_expert_answers_fit_no_targets.py
AbdelrhmanBassiouny: This is from automatic import generation from the scope and type hints of the objects relevant to the rdr interaction. They are not necessarily used but are there for convenience. 

I reverted them anyway, they will only be auto generated again if someone manually interacts with the rdr tests by setting a flag of regenerate answers to True which noone will do anyway.

ichumuh (approved):

PR #473, opened 2026-07-24
Luca is always right

ichumuh:
- Changed FieldMetaData to @LucaKro design
- DataclassJsonSerializer uses custom MetaData to skip/include fields
- MotionStatechartNode doesn't inherit from SubClassJsonSerializer anymore


LucaKro (approved):
beautiful

krrood/src/krrood/patterns/field_metadata.py
```diff
@@ -1,67 +1,45 @@
...
-        return field_metadata.get_metadata_by_type(cls)
+        return field_.metadata.get(cls)
 
 
-MetadataType = TypeVar("MetadataType", bound=FieldMetadata)
-"""
-A type that is a subclass of :class:`FieldMetadata`.
-"""
+@dataclass
+class JSONMetadata(FieldMetadata):
```
tomsch420, line 41: hmm, why is a json serialize flag not port of the base fieldmetadata? Is there a case where this information shouldnt be there?

tomsch420 (changes requested):

krrood/src/krrood/patterns/field_metadata.py
LucaKro: ~hmm fair point, could make this a general thing that you can say which metadata should be serialized :+1:~

LucaKro: actually it says for the whole field if it should be serialized or not, so having it in here makes sense nvm. because otherwise you may have contradicting flags whether to serialize or not

tomsch420: yeah, i would only make subclasses if its domain specfic, e. g. something inside the semantic annotations that tells how this should be treated or so. But general KRROOD related stuff can just go in there

ichumuh: hmm, but that might get complicated when we add multiple fieldmatadata to a field and they conflict. i prefer it like this

tomsch420: why would you add multiple field metadata objects to one field? Every field has exactly one constructor in the entire code

LucaKro: I think there should be a KRROODMetadata subclass for that. and maybe that class can also contain the info for serialization. but the superclass everyone inherits from should stay clean, so if you have two metadata classes you implemented and both want to add to a field, there arent any conflicts in there, which you would have if you put  the krrood information in the superclass

krrood/src/krrood/patterns/field_metadata.py
```diff
@@ -1,67 +1,45 @@
...
 
 @dataclass
 class FieldMetadata:
     """
     Krrood-specific metadata carried inside a dataclass field's ``metadata`` mapping.
 
-    A field carries a single :class:`FieldMetadata` under :attr:`METADATA_KEY` (attach
-    it with :meth:`as_dict`). Specific metadata classes are held as sub-metadatas in
-    :attr:`other_metadata` and read back by type via :meth:`get_metadata_by_type`.
+    A field carries at most one instance of a given :class:`FieldMetadata` subclass,
```
AbdelrhmanBassiouny, line 13: Why at most one? Isn't the point of it to allow multiple types of metadatas?

krrood/src/krrood/patterns/field_metadata.py
```diff
@@ -1,67 +1,45 @@
...
-        """:return: the first sub-metadata in :attr:`other_metadata` that is an instance of
-        *metadata_type*, or ``None`` when none is present.
+        :return: a dataclass-field ``metadata`` mapping carrying this metadata under its own
+            type, ready to pass to ``field(metadata=...)``.
         """
-        for metadata in self.other_metadata:
-            if isinstance(metadata, metadata_type):
-                return metadata
-        return None
+        return {type(self): self}
```
AbdelrhmanBassiouny, line 23: Ok you can make the metada data keys types? I actually thought only string keys are allowed. 

I like this. Then we just combine dicts of the different metadatas using **

AbdelrhmanBassiouny (changes requested):

krrood/src/krrood/patterns/field_metadata.py
ichumuh: only one of the given same type, but you can have multiple of different types

krrood/src/krrood/patterns/field_metadata.py
ichumuh: after a long discussion in person, @tomsch420 said that we can keep it like it is for now

tomsch420 (approved):

krrood/src/krrood/patterns/field_metadata.py
AbdelrhmanBassiouny: Ok I see

AbdelrhmanBassiouny (approved):

PR #474, opened 2026-07-26
[Verbalization] Auto-generate the verbalization surfaces snapshot module

AbdelrhmanBassiouny:
Generates `verbalization_surfaces.py` via `VerbalizationSurfaceGenerator` + Jinja, keeping it fresh on every test run (same pattern as `ormatic_interface.py`). See fork PR #97 for full details: https://github.com/AbdelrhmanBassiouny/cognitive_robot_abstract_machine/pull/97


krrood/src/krrood/entity_query_language/testing/templates/__init__.py
```diff
@@ -0,0 +1 @@
+"""Template package for verbalization-testing Jinja2 templates."""
```
LucaKro: i feel like these are so pointless 😅

krrood/src/krrood/entity_query_language/testing/surface_generation.py
```diff
@@ -0,0 +1,130 @@
...
+from krrood.entity_query_language.testing.surface_verification import (
+    SymbolicSurfaceSnapshot,
+    VerbalizationSurface,
+)
+
+
+@dataclass
+class VerbalizationSurfaceGenerator:
+    """
+    Renders the Python source of a ``SURFACES`` snapshot module for a
```
LucaKro: why "surface"?

LucaKro (changes requested):

krrood/src/krrood/entity_query_language/testing/surface_generation.py
tomsch420: yeah i also thinks its too viby.

krrood/src/krrood/entity_query_language/testing/surface_generation.py
```diff
@@ -0,0 +1,130 @@
...
+    """
+    Renderer bound to this package's templates directory.
+    """
+
+    def __post_init__(self):
+        templates = importlib.resources.files(__package__) / "templates"
+        self.code_generator = CodeGenerator(template_directory=str(templates))
+
+    def covered_callables(self) -> Tuple[Type[SymbolicCallable], ...]:
+        """:return: the discovered callables that implement their own verbalization
```
tomsch420: formatting

krrood/src/krrood/entity_query_language/predicate.py
```diff
@@ -203,6 +204,28 @@ def _verbalization_fragment_(cls, fields):
         """
 
 
+@dataclass(frozen=True)
+class OverriddenOperand:
+    """
+    One dataclass field's concrete VALUE for a symbolic callable whose fragment reads
```
tomsch420: stop screaming

tomsch420 (changes requested):

krrood/src/krrood/entity_query_language/testing/surface_generation.py
AbdelrhmanBassiouny: This should have been a comment on on the previous merged pr 😅

krrood/src/krrood/patterns/role_predicates.py
```diff
@@ -33,18 +33,18 @@ class IsSameSemanticEntity(Predicate):
     symbolically inside an entity-query-language ``where`` clause.
     """
 
-    entity_1: Any
+    first_entity: Any
```
tomsch420, line 36: you dont have to fix this but i thought about this more. I think using the field name is still correct, as in the semantic sense, this is what this field means in this context.  In this instance, this should be a list of Objects thats compared, and hence it would read weirdly if you just assume two entities and do it with these field names. Does that make sense?

tomsch420 (approved):

krrood/src/krrood/entity_query_language/predicate.py
AbdelrhmanBassiouny: Done

krrood/src/krrood/entity_query_language/testing/surface_generation.py
AbdelrhmanBassiouny: Done

krrood/src/krrood/entity_query_language/testing/surface_generation.py
AbdelrhmanBassiouny: Done changed to result instead of surface

krrood/src/krrood/entity_query_language/testing/templates/__init__.py
AbdelrhmanBassiouny: Removed

krrood/src/krrood/patterns/role_predicates.py
AbdelrhmanBassiouny: Yeah I know what you mean, currently the default isnusing the type unless the the type in Any,  or object, otherwise it used the field name. 

The type if it is not generic should always give a correct meaning, but perhaps not the most specific meaning to this context. 

I decided to leave it as it is, and we will see from the coming prs that migrate semdt callables and coraplex callables how the current one fares and see if we want to keep it or change it.

LucaKro (approved):

PR #476, opened 2026-07-27
Translate any propositional where expression into a random event

tomsch420:
## What

`WhereExpressionToRandomEventTranslator` required its input to be in disjunctive normal form and rejected everything else. That precondition is not needed: instead of flattening the condition by hand, the translator now recurses over the condition tree and maps `AND`, `OR` and `Not` onto the corresponding operations of the product algebra, which normalizes the result itself. A `where` expression may therefore have any propositional shape — disjunctions nested inside conjunctions, negations, and combinations of both.

Every leaf is built over all variables of the query rather than over the compared one alone, because the product algebra combines events by the variables they already carry and would otherwise drop the variables an operand does not mention.

## Refused conditions

Quantified conditions (`for_all`, `exists`) are the only genuinely unmappable ones: a product algebra is propositional and constrains a fixed set of variables, so it states nothing about the objects a quantifier ranges over. They raise the new `WhereExpressionIsFirstOrder`. A part of a condition that constrains nothing a random event variable stands for, such as a comparison between two variables, raises the new `WhereExpressionHasNoRandomEventRepresentation`.

Both replace `WhereExpressionNotInDisjunctiveNormalForm`, and the now unused disjunctive normal form checks (`is_disjunctive_normal_form` and friends) are removed with it. `is_literal_comparator` stays, since it is what decides whether a leaf is representable.

## Tests

Every test of the translator now lives in `test_random_events_translator.py`, including `test_underspecification_with_where`, which moved out of `test_probable_variable.py`. The new ones were written before the implementation and fail against the old translator:

- a `where` condition outside disjunctive normal form, and a disjunction nested in a conjunction sharing a variable with it,
- `not_(x > 0)` equalling `x <= 0`, and De Morgan on a negated conjunction,
- a quantified `where` condition and a comparison between two variables, both refused.

Full `krrood` suite passes (1928 passed, 6 skipped; `test_code_generation` and `test_eql/test_typing` do not collect locally because `inflection` is missing, unrelated to this change).

https://claude.ai/code/session_01SCjykZP4vnSY462RQpbnQZ

### DIFF
--- a/krrood/src/krrood/parametrization/random_events_translator.py
+++ b/krrood/src/krrood/parametrization/random_events_translator.py
@@ -2,10 +2,9 @@ from __future__ import annotations
 
 import itertools
 import operator
-from collections import deque
 from dataclasses import dataclass
 from functools import cached_property
-from typing import assert_never, List, Dict
+from typing import assert_never, Dict
 
 import numpy as np
 
@@ -16,12 +15,19 @@ from krrood.entity_query_language.core.mapped_variable import MappedVariable
 from krrood.entity_query_language.core.variable import Literal
 from krrood.entity_query_language.factories import ConditionType
 from krrood.entity_query_language.operators.comparator import Comparator
-from krrood.entity_query_language.operators.core_logical_operators import OR, AND
+from krrood.entity_query_language.operators.core_logical_operators import OR, AND, Not
+from krrood.entity_query_language.operators.logical_quantifiers import (
+    QuantifiedConditional,
+)
 from krrood.parametrization.exceptions import (
-    WhereExpressionNotInDisjunctiveNormalForm,
+    WhereExpressionHasNoRandomEventRepresentation,
+    WhereExpressionIsFirstOrder,
 )
 from random_events.interval import closed_open, closed, open
 from random_events.product_algebra import Event, SimpleEvent
+
+
+# %% translating a where expression into a random event
 
 
 @dataclass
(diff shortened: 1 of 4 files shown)


krrood/src/krrood/parametrization/random_events_translator.py
```diff
@@ -74,110 +76,103 @@ def _get_variable_from_comparator(
...
-        self, expression: SymbolicExpression
-    ) -> Dict[random_events.variable.Variable, List[Comparator]]:
+        result = self.unconstrained_simple_event()
+        variable = self._get_variable_from_comparator(comparator)
+
+        if isinstance(comparator.right._value_, type(Ellipsis)):
+            return result.as_composite_set()
+
+        match comparator.operation:
+            case operator.eq:
```
LucaKro, line 143: no abbreviations

LucaKro (changes requested):

krrood/src/krrood/parametrization/random_events_translator.py
tomsch420: what do you mean? Operator is a builtin of python

krrood/src/krrood/parametrization/random_events_translator.py
```diff
@@ -74,110 +76,103 @@ def _get_variable_from_comparator(
...
+            return self._translate_expression(
+                expression.left
+            ) & self._translate_expression(expression.right)
+
+        if isinstance(expression, OR):
+            return self._translate_expression(
+                expression.left
+            ) | self._translate_expression(expression.right)
+
+        if isinstance(expression, Not):
```
ichumuh, line 109: why `Not` and not `NOT`? like the other two

krrood/src/krrood/parametrization/random_events_translator.py
```diff
@@ -74,110 +76,103 @@ def _get_variable_from_comparator(
     def translate(self) -> Event:
         """
         :return: The random event that corresponds to the query.
+        :raises WhereExpressionIsFirstOrder: When the query quantifies over a variable.
+        :raises WhereExpressionHasNoRandomEventRepresentation: When a part of the query
+            constrains nothing that a random event variable stands for.
```
ichumuh, line 81: why does this method have the `raises` and not `_translate_expression`?

ichumuh (changes requested):
use github "KRROOD" label instead of writing "[KRROOD]" in the title grrrrr

krrood/src/krrood/parametrization/random_events_translator.py
tomsch420: It is the EQL class. Which NOT are you referring to?

krrood/src/krrood/parametrization/random_events_translator.py
tomsch420: It is the public entry point. The others are protected. Anyways I will update the docstring

krrood/src/krrood/parametrization/random_events_translator.py
ichumuh: i mean, why is it not all caps, but if its not your not, then whatever i guess

krrood/src/krrood/parametrization/random_events_translator.py
tomsch420: renamed abbrivated methods

ichumuh (approved):

LucaKro (approved):

AbdelrhmanBassiouny (approved):

PR #477, opened 2026-07-27
End motion slowdown

ichumuh:
- EndMotion waits for a local minimum before turning true, if there are node in the statechart that cause motions. **this changes behavior**
- CancelMotion behavior stays the same. motions will still stop abruptly
- fixed a bug where plotting for trajectories didn't work in giskardpy_ros

before (test_cart_goal_1eef):
[goal_1.pdf](https://github.com/user-attachments/files/30410937/goal_1.pdf)

after:
[after.pdf](https://github.com/user-attachments/files/30410899/after.pdf)

notice how the trajectory is now a little longer and all velocities are 0


giskardpy/src/giskardpy/middleware/ros2/behavior_tree_config.py
```diff
@@ -14,6 +14,11 @@ class BehaviorTreeConfig:
     tree_tick_rate: float = 0.05
     debug_mode: bool = False
 
+    plot_output_directory: str = "/tmp/"
```
LucaKro: why dont you use https://docs.python.org/3/library/tempfile.html

giskardpy/src/giskardpy/motion_statechart/graph_node.py
```diff
@@ -886,6 +886,47 @@ def __repr__(self) -> str:
...
+        velocity_limit = dof.limits.upper.velocity * joint_convergence_threshold
+        velocity_limit = min(max(minimum_threshold, velocity_limit), maximum_threshold)
+        ref.append(velocity_limit)
+        symbols.append(dof.variables.velocity)
+
+    dt = (
+        context.qp_controller_config.control_dt
+        or context.qp_controller_config.model_predictive_control_time_step
+    )
+    trajectory_longer_than_one_second = context.control_cycle_variable * dt > 1
```
maltehue, line 923: so no shorter trjectories are allowed?

LucaKro (changes requested):

giskardpy/src/giskardpy/middleware/ros2/behavior_tree_config.py
ichumuh: done

giskardpy/src/giskardpy/middleware/ros2/behavior_tree_config.py
```diff
@@ -1,3 +1,4 @@
+import tempfile
```
LucaKro: a lone import?

LucaKro (approved):

maltehue (approved):

giskardpy/src/giskardpy/motion_statechart/graph_node.py
ichumuh: currently yes, i need some threshold to prevent it from triggering before the motion even started. maybe 1s is too long though idk.

maltehue: ok, we will see once we encounter problems. in that case we might have to expose it. but fine for me for now

PR #478, opened 2026-07-27
Add motion statechart scale tests; fix O(N²) tick cost on large graphs

ichumuh:
**Summary**
                                                                                                                                                                                                                            
  - Add stress/scale tests for MotionStatechart compilation and tick performance, covering two topologies that keep the number of concurrently active nodes at exactly one while growing total graph size: a long linear Sequence chain, and  
  many alternative branches where only one is ever traveled (the rest are gated permanently NOT_STARTED).                                                                                                                                     
  - Register a slow pytest marker, excluded by default via addopts = -sv -m "not slow", so these large-graph tests don't add to normal test runs.                                                                                             
  - Fix an O(N) per-tick cost in MotionStatechart.tick() that the scale tests surfaced: three full for node in self.nodes Python loops were rescanning every node on every tick — reconstructing LifeCycleValues enums per node — just to     
  rediscover the one or two nodes that actually changed. That made a full N-node sequence sweep O(N²).                                                                                                                                        
                                                                                                                                                                                                                                              
**Why**

Profiling a 3,000-node chain showed real CasADi/numpy state computation was only ~10% of tick cost; the remaining ~90% was pure Python-interpreter loop/enum-construction overhead with no dirty-tracking. At N=10,000 nodes, a full        
  sequence sweep took 204 seconds.                                                                                                                                                                                                            
                                                                                                                                                                                                                                              
**Fix**

In giskardpy/src/giskardpy/motion_statechart/motion_statechart.py:                                                                                                                                                                          
  - Cache _nodes, _cancel_motion_nodes, _end_motion_nodes on MotionStatechart, populated incrementally in add_node (nodes are only ever appended, never removed, so this stays index-aligned with the rustworkx graph for free).              
  - _update_observation_state and _trigger_life_cycle_callbacks now use np.flatnonzero to read changed/RUNNING node indices directly off the numpy state arrays, instead of looping over all N nodes and enum-constructing each one to check. 
  - is_end_motion / _raise_if_cancel_motion iterate the cached type-specific lists instead of an O(N) isinstance scan via get_nodes_by_type.                                                                                                  
                                                                                                                                                                                                                                              
**Results (long-sequence topology, tick-loop time)**
before:
<img width="430" height="395" alt="image" src="https://github.com/user-attachments/assets/c86536f4-ec7b-49e0-a4e2-c0d846136f1c" />

after:
<img width="445" height="388" alt="image" src="https://github.com/user-attachments/assets/81510c15-01c7-45dc-96a0-25ffff419c14" />


test/giskardpy_test/test_motion_statechart/test_motion_statechart_scale.py
```diff
@@ -0,0 +1,111 @@
...
+        if previous is not None:
+            node.start_condition = previous.observation_variable
+        node.end_condition = node.observation_variable
+        chain.append(node)
+        previous = node
+    return chain
+
+
+@pytest.mark.slow
+@pytest.mark.parametrize("node_count", [100, 1_000, 10_000])
```
MarvinDufner, line 38: isnt it enough to run the highest node count or do you need the comparison?

MarvinDufner (approved):

test/giskardpy_test/test_motion_statechart/test_motion_statechart_scale.py
ichumuh: the purpose of this test is to see how well it scales. it actually prints smt like this:
```
=================== 6 passed, 5 warnings in 75.24s (0:01:15) ===================
[long_sequence] N=100 compile=0.2045s tick=0.0304s (303.99 us/node)
PASSED[long_sequence] N=1000 compile=2.1511s tick=0.5000s (499.98 us/node)
PASSED[long_sequence] N=10000 compile=22.2601s tick=27.0736s (2707.36 us/node)
PASSED[alternative_branches] branches=10 length=10 total_nodes=102 compile=0.1213s tick=0.0034s (33.52 us/node)
PASSED[alternative_branches] branches=10 length=100 total_nodes=1002 compile=1.1382s tick=0.0498s (49.72 us/node)
PASSED[alternative_branches] branches=10 length=1000 total_nodes=10002 compile=11.6522s tick=2.5547s (255.42 us/node)
```
the tests are skipped by default

maltehue (approved):

AbdelrhmanBassiouny (approved):

PR #484, opened 2026-07-28
Add ScrewConnection to model screw mechanisms as a single dof

maltehue:
A screw pair couples rotation about an axis with translation along it via a pitch (meters per radian), so unscrewing a bottle cap becomes a plain joint goal for the existing Open/Close goals in giskardpy.

- ScrewConnection(ActiveConnection1DOF) with required pitch field; rotation_angle_for_travel_distance converts travel goals to rotation
- ScrewJoint mechanical joint annotation and BottleCap (renamed from the unused LiquidCap, now with HasMechanicalJoint) so caps mount like door hinges; generic connection_kwargs passthrough in the annotation factories forwards extra connection parameters without type branching
- Regenerated ORM interfaces
- Tests: symbolic/numeric FK coupling, JSON and database round-trips, annotation factory and mount regression, and a PR2 statechart test that unscrews a cap by rotation and tightens it back by travel distance using the unchanged Open/Close goals

Made with the help of Claude
[screw-motion.webm](https://github.com/user-attachments/assets/18618dbb-70e8-4ed1-9fbc-c7308df4de5a)


test/giskardpy_test/test_motion_statechart/test_other_tasks.py
```diff
@@ -848,3 +852,159 @@ def test_open(self, pr2_world_copy, tmp_path):
...
+                            ),
+                        ),
+                        Parallel(
+                            [
+                                Open(
+                                    tip_link=r_tip,
+                                    environment_link=cap_body,
+                                    goal_joint_state=unscrew_goal,
+                                ),
+                                unscrewed := JointPositionReached(
```
ichumuh: this monitor should be unnecessary, the open goal should be true at the same time.

semantic_digital_twin/src/semantic_digital_twin/world_description/connections.py
```diff
@@ -386,6 +386,168 @@ def add_to_world(self, world: World):
...
+    single degree of freedom.
+
+    Increasing the degree of freedom's position rotates the child counterclockwise about
+    ``axis`` (right-hand rule) and translates it along ``axis`` by ``pitch * position``.
+    A right-handed thread whose ``axis`` points from the parent toward the child
+    therefore has a positive ``pitch``: driving the degree of freedom toward its upper
+    limit unscrews the child.
+    """
+
+    pitch: float = field(kw_only=True)
```
ichumuh: so this is the pitch of the thread? i find this name confusing. would have expected something like number of revolutions until the bottle is closed, but i'm not deep into bottles 🤷

semantic_digital_twin/src/semantic_digital_twin/world_description/connections.py
```diff
@@ -386,6 +386,168 @@ def add_to_world(self, world: World):
...
+    ``axis`` during one full revolution (for common single-start threads this equals the
+    distance between neighbouring thread crests). Convert it via ``pitch = lead / (2 *
+    math.pi)``. Negative values model left-handed threads.
+    """
+
+    def add_to_world(self, world: World):
+        super().add_to_world(world)
+
+        angle = self.dof.variables.position
+        translation_axis = self.axis * (self.pitch * angle)
```
ichumuh: we can decide between the dof representing the translation and the angle. i guess the current angle choice is better, because its the same unit most robot joints use too

ichumuh (changes requested):
cool stuff. i wonder if we can represent somehow, that the cup is disconnected once the cap is unscrewed.

semantic_digital_twin/src/semantic_digital_twin/world_description/connections.py
LucaKro: I agree, as discussed maybe call it scrwe_pitch and link an wikipedia article where it explains why its called that

LucaKro (changes requested):
nice good job, would be cool to see a video of it!

maltehue:
> cool stuff. i wonder if we can represent somehow, that the cup is disconnected once the cap is unscrewed.

yes, i would  put that into a follow up. Similar to either an Action or Motionstatechart for automatic regrasping to handle wrist joint limits.

LucaKro:
wooow did someone not use pycharms refactor functionality? wooooow

LucaKro:
okay sorry that was mean :cry: <3

LucaKro (approved):

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/semantic_annotations.py
```diff
@@ -302,6 +303,81 @@ def _parent_connection_type(self) -> Type[Connection]:
         return PrismaticConnection
 
 
+@dataclass(eq=False)
+class ScrewJoint(MechanicalJoint):
```
ichumuh: isn't it confusing to have ScrewJoint and ScrewConnection?

test/giskardpy_test/test_motion_statechart/test_other_tasks.py
```diff
@@ -848,3 +855,148 @@ def test_open(self, pr2_world_copy, tmp_path):
 
         assert opened.observation_state == ObservationStateValues.TRUE
         assert closed.observation_state == ObservationStateValues.TRUE
+
+    def test_unscrew_and_tighten_bottle_cap(self, pr2_world_copy, rclpy_node):
+        screw_pitch = 0.03
+        unscrew_goal = 2 * np.pi
+        VizMarkerPublisher(_world=pr2_world_copy, node=rclpy_node).with_tf_publisher()
```
ichumuh: delete vis marker publisher

test/giskardpy_test/test_motion_statechart/test_other_tasks.py
```diff
@@ -848,3 +855,148 @@ def test_open(self, pr2_world_copy, tmp_path):
...
+                        # Grasp with the roll axis of the gripper collinear with the
+                        # screw axis, so the wrist can follow the cap's rotation.
+                        CartesianPose(
+                            root_link=pr2_world_copy.root,
+                            tip_link=r_tip,
+                            goal_pose=HomogeneousTransformationMatrix.from_xyz_rpy(
+                                reference_frame=cap_body
+                            ),
+                        ),
+                        Parallel(
```
ichumuh: useless parallel

test/giskardpy_test/test_motion_statechart/test_other_tasks.py
```diff
@@ -848,3 +855,148 @@ def test_open(self, pr2_world_copy, tmp_path):
...
+                                    environment_link=cap_body,
+                                    goal_joint_state=unscrew_goal,
+                                ),
+                            ]
+                        ),
+                    ]
+                ),
+            ]
+        )
+        unscrew_statechart.add_node(EndMotion.when_true(unscrew_statechart.nodes[0]))
```
ichumuh: assign node with walrus to variable and use the variable here

test/giskardpy_test/test_motion_statechart/test_other_tasks.py
```diff
@@ -848,3 +855,148 @@ def test_open(self, pr2_world_copy, tmp_path):
...
+        unscrew_travel_distance = float(np.linalg.norm(cap_translation))
+        tightened_goal_joint_state = (
+            unscrew_goal
+            - root_C_screw.rotation_angle_for_travel_distance(unscrew_travel_distance)
+        )
+
+        tighten_statechart = MotionStatechart()
+        tighten_statechart.add_nodes(
+            [
+                Parallel(
```
ichumuh: see above

test/giskardpy_test/test_motion_statechart/test_other_tasks.py
```diff
@@ -848,3 +855,148 @@ def test_open(self, pr2_world_copy, tmp_path):
...
+                        close := Close(
+                            tip_link=r_tip,
+                            environment_link=cap_body,
+                            goal_joint_state=tightened_goal_joint_state,
+                        ),
+                    ]
+                ),
+            ]
+        )
+        tighten_statechart.add_node(EndMotion.when_true(tighten_statechart.nodes[0]))
```
ichumuh: see above

test/semantic_digital_twin_test/test_adapters/test_json_parsing.py
```diff
@@ -260,6 +263,40 @@ def test_connection_json_serialization_with_world():
...
 
+def test_screw_connection_json_serialization_with_world():
+    world = World()
+    body = Body(name=PrefixedName("body"))
+    body2 = Body(name=PrefixedName("body2"))
+    screw_pitch = 0.005
+    with world.modify_world():
+        world.add_kinematic_structure_entity(body)
+        world.add_kinematic_structure_entity(body2)
+        connection = ScrewConnection.create_with_dofs(
```
ichumuh, line 274: why manual creation instead of using ScreJoint again?

ichumuh (changes requested):

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/semantic_annotations.py
LucaKro: hmm fair, i think the technical name is  helical joint? so maybe  helical Connection, and Screw(MechanicalJoint)? tho i already see the nameclash with the actual physical screw

LucaKro: ScrewMechanism?

test/giskardpy_test/test_motion_statechart/test_other_tasks.py
maltehue: :scissors:

test/semantic_digital_twin_test/test_adapters/test_json_parsing.py
maltehue: i want to test the serializer not the factory from the semantic annotation. Do you prefer to change it?

ichumuh: just wondering, doesn't really matter

ichumuh (approved):

PR #507, opened 2026-08-04
Claude optimization

Tigul:
Some small optimizations for creating collision in the PyBullet collision detector and the re-building of the class diagram in conftest. 

Measured performance in test_multi_robot_action_designator: 306s -> 187s

### DIFF
--- a/coraplex/src/coraplex/datastructures/dataclasses.py
+++ b/coraplex/src/coraplex/datastructures/dataclasses.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 import logging
 from dataclasses import dataclass, field
-from functools import cached_property
 
 from typing_extensions import (
     Optional,
@@ -18,6 +17,7 @@ from krrood.entity_query_language.backends import (
     EntityQueryLanguageGenerativeBackend,
 )
 from krrood.class_diagrams.mocking import MockedClass, MockedModule
+from krrood.utils import memoize
 from coraplex.plans.plan import Plan
 from coraplex.plans.plan_entity import PlanEntity
 from semantic_digital_twin.robots.robot_parts import AbstractRobot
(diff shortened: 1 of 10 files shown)


giskardpy/src/giskardpy/motion_statechart/exceptions.py
```diff
@@ -251,6 +251,28 @@ def suggest_correction(self) -> str:
...
 
 
+@dataclass
+class ActionClientTypeMismatchError(MotionStatechartError):
+    """
+    Raised when an action topic is requested with a different message type than the
+    one its cached action client was created with.
+    """
+
+    action_topic: str
```
ichumuh, line 274: param doc

giskardpy/src/giskardpy/motion_statechart/ros_context.py
```diff
@@ -1,13 +1,54 @@
...
+                raise ActionClientTypeMismatchError(
+                    action_topic=action_topic,
+                    existing_message_type=action_client._action_type,
+                    requested_message_type=message_type,
+                )
+            return action_client
+
+        action_client = ActionClient(self.ros_node, message_type, action_topic)
+        self._action_clients[action_topic] = action_client
+        return action_client
```
ichumuh, line 60: @claude does this work if we have two nodes in the same motion statechart that use the same action client?

ichumuh (changes requested):

giskardpy/src/giskardpy/motion_statechart/ros_context.py
Tigul: As long as they are not working in parallel I don't see a problem there but will investigate

coraplex/src/coraplex/datastructures/dataclasses.py
```diff
@@ -107,6 +108,13 @@ def debug(self, value):
             logging.DEBUG if self.debug else logging.INFO
         )
 
+    @cached_property
```
LucaKro: memoize just to be sure, as the cached giskardwrapper holds an active world reference

LucaKro (changes requested):

ichumuh (approved):

LucaKro (approved):

PR #515, opened 2026-08-05
[DevTools] Fix a dashboard rendering issue.

AbdelrhmanBassiouny:
- Root cause: a plan item's `blockers` field was a YAML prose block (a string), and `list(data.get("blockers") or [])` silently iterated it character-by-character, rendering one bullet per character on the dashboard.
- Underlying data already fixed directly on `personal-notes` (plan data never lives on a reviewed branch, per convention).
- This PR is the code-side hardening: adds `InvalidBlockers`, mirroring the existing `InvalidDependsOn` check, so a non-list `blockers` now fails validation loudly instead of corrupting the render.
- New regression test mirrors the existing `depends_on` test.

Fork PR: https://github.com/AbdelrhmanBassiouny/cognitive_robot_abstract_machine/pull/136


.claude/skills/plan-dashboard/build_dashboard.py
```diff
@@ -251,6 +251,26 @@ def describe(self) -> str:
...
+    blocker per character instead of failing loudly.
+    """
+
+    item_identifier: str
+    """The offending item's effective id."""
+
+    actual_type: type
+    """The type ``blockers`` actually held, instead of ``list``."""
+
+    def describe(self) -> str:
```
LucaKro: please dont start to introduce variations of our existing dataclass exception...

LucaKro (changes requested):

ichumuh (approved):
nothing to add beyond what luca said, so i'll approve already

.claude/skills/plan-dashboard/build_dashboard.py
AbdelrhmanBassiouny: Since I can't depend on DataclassException, I think a Protocol that mimics DataclassException would be a good intermediate solution here.

LucaKro (approved):

PR #516, opened 2026-08-05
[RDR] First Slice: backward inference + condition suggestion

AbdelrhmanBassiouny:
- First slice of the RDR engine (`entity_query_language/rdr/`): the backtracking + condition-suggestion foundation the rest of the engine builds on.
- `backward_inference.py` — the `BackwardInferenceIndex` + backward-inference query engine (the `%knows` backend logic).
- `condition_resolver.py` — the condition-suggestion resolvers used during fitting.
- `utils.py` — the shared `rdr` helpers those two build on.
- `__init__.py` intentionally empty (no re-exports); consumers import submodules directly.
- Imports only the EQL core, so no `code_generation`/Match-API reconciliation here — it's the clean base the coupled slices stack on.
- Stacked on `code-generation-extract` (https://github.com/AbdelrhmanBassiouny/cognitive_robot_abstract_machine/pull/39) → `ripple-down-rules-refactor` (https://github.com/AbdelrhmanBassiouny/cognitive_robot_abstract_machine/pull/40) → this PR.

Fork PR: https://github.com/AbdelrhmanBassiouny/cognitive_robot_abstract_machine/pull/41


krrood/src/krrood/entity_query_language/rdr/__init__.py
tomsch420, line 1: The whole package is an abbreviation. I actually don't mind here but technically: no abbreviations

krrood/src/krrood/entity_query_language/rdr/backward_inference.py
```diff
@@ -0,0 +1,336 @@
+"""
+Backward inference for EQL-RDR rule trees.
+
+Given a target conclusion value (e.g. ``Species.molusc``), traverse the rule-tree DAG
```
tomsch420: Is it a tree or a dag?

krrood/src/krrood/entity_query_language/rdr/backward_inference.py
```diff
@@ -0,0 +1,336 @@
...
+backwards to enumerate every rule path that could produce it. Each path accumulates
+*guard conditions* from the ``Refinement``/``Alternative``/``Next`` selectors and wraps
+the result as a :class:`SufficientConditionSet`. The full answer is the disjunction of all
+such sets (a DNF formula).
```
tomsch420, line 10: If it's a DNF formula it has natural synergies with random events. Just saying, perhaps you can use that at some point to detect disjointedness of formulas

krrood/src/krrood/entity_query_language/rdr/backward_inference.py
```diff
@@ -0,0 +1,336 @@
...
+
+
+# %%
+# Data structures
+
+
+@dataclass(frozen=True)
+class GuardCondition:
+    """A condition that must hold for a rule to fire.
+
```
tomsch420: Fire is jargon. Use a neutral word

krrood/src/krrood/entity_query_language/rdr/backward_inference.py
```diff
@@ -0,0 +1,336 @@
...
+        :param shared_variable: The EQL variable the conditions range over
+            (the rule tree's ``case_variable``).
+        :param case: The concrete case object to evaluate against.
+        :return: ``True`` if every guard condition is satisfied.
+        """
+        return all(guard.holds_for(shared_variable, case) for guard in self.conditions)
+
+
+@dataclass(frozen=True)
+class ConclusionKnowledge:
```
tomsch420: Knowledge is a very generic class name. Can you specify this more?

krrood/src/krrood/entity_query_language/rdr/backward_inference.py
```diff
@@ -0,0 +1,336 @@
...
+            conclusion_value,
+            ConclusionKnowledge(conclusion_value, ()),
+        )
+
+
+# %%
+# Public API
+
+
+def what_do_we_know_about(
```
tomsch420: I think that's a very ambiguous name

krrood/src/krrood/entity_query_language/rdr/condition_resolver.py
```diff
@@ -0,0 +1,244 @@
...
+if TYPE_CHECKING:
+    from krrood.entity_query_language.core.base_expressions import SymbolicExpression
+    from krrood.entity_query_language.core.variable import Variable
+    from krrood.entity_query_language.rdr.backward_inference import (
+        GuardCondition,
+        SufficientConditionSet,
+    )
+
+
+class ResolutionMode(Enum):
```
tomsch420: Looks like a str enum to me

krrood/src/krrood/entity_query_language/rdr/condition_resolver.py
```diff
@@ -0,0 +1,244 @@
...
+        :param corner_case: The case that triggered the firing rule's creation.
+        :param target_knowledge: Backward-inference knowledge for ``target_conclusion``.
+        :param current_knowledge: Backward-inference knowledge for ``current_conclusion``.
+        :param firing_anchor: Condition expression of the rule that fired, used to
+            identify the active path without re-evaluating the rule tree.
+        :return: A :class:`ResolvedCondition`, or ``None`` if resolution is not possible.
+        """
+
+
+def _materialize(guard: GuardCondition) -> SymbolicExpression:
```
tomsch420: The method name doesn't really makes sense to me

krrood/src/krrood/entity_query_language/rdr/condition_resolver.py
```diff
@@ -0,0 +1,244 @@
...
+    """Produce the EQL expression for a new rule's condition, applying negation.
+
+    A negated guard is wrapped with ``not_`` so the new rule fires when the guard's
+    expression is False.
+    """
+    return not_(guard.expression) if guard.negated else guard.expression
+
+
+class TargetKnowledgeResolver(ConditionResolver):
```
tomsch420: Also here the name is ambiguous and vibzy. Why not: SufficientConditionResolver

krrood/src/krrood/entity_query_language/rdr/utils.py
```diff
@@ -0,0 +1,27 @@
...
+
+    def __repr__(self) -> str:
+        return "UNSET"
+
+    def __str__(self) -> str:
+        return "UNSET"
+
+
+UNSET: _Unset = _Unset.UNSET
+"""Sentinel for "no current/target conclusion was supplied" (e.g. the ask-for-rule path)."""
```
tomsch420: Why is this called UNSET instead of something more self described?

tomsch420 (changes requested):

krrood/src/krrood/entity_query_language/rdr/__init__.py
AbdelrhmanBassiouny: I think it is better to do this at the end of the refactor not in this splice as it will have a ripple effect on all downward PRs which I don't want to deal with now.

krrood/src/krrood/entity_query_language/rdr/backward_inference.py
AbdelrhmanBassiouny: The rules are a tree view off the whole expression dag.

AbdelrhmanBassiouny: if you consider each rule as a box and the branching between them the edges, you will get a tree, but if you include the expressions in the rules, it will be a DAG.

AbdelrhmanBassiouny: I clarified the comment to include this explanation

krrood/src/krrood/entity_query_language/rdr/backward_inference.py
```diff
@@ -0,0 +1,336 @@
...
+@dataclass(frozen=True)
+class SufficientConditionSet:
+    """One rule path's complete conditions for a specific conclusion value to fire.
+
+    The conditions are stored as :class:`GuardCondition` tuples. Use
+    :meth:`evaluate_against` to check them against a concrete case without mutating the
+    original rule tree.
+    """
+
+    conditions: Tuple[GuardCondition, ...]
```
LucaKro, line 56: docs

krrood/src/krrood/entity_query_language/rdr/backward_inference.py
```diff
@@ -0,0 +1,336 @@
...
+class _RulePath:
+    """An internal value object for one discovered rule path during traversal."""
+
+    conditions: Tuple[GuardCondition, ...]
+    """Guard conditions accumulated along the path to these add nodes."""
+    add_nodes: Tuple[Add, ...]
+    """Conclusion nodes at the leaf of this rule path."""
+
+
+def _leaf_guards(
```
LucaKro, line 110: hmm not very openclosed no?

krrood/src/krrood/entity_query_language/rdr/backward_inference.py
```diff
@@ -0,0 +1,336 @@
...
+        for child in expr._operation_children_:
+            result.extend(_leaf_guards(child, negated))
+        return result
+    if isinstance(expr, Not) and isinstance(expr._child_, ConclusionSelector):
+        # Push negation through the selector — refines NOT(Refinement), NOT(Alternative), NOT(Next)
+        return _leaf_guards(expr._child_, not negated)
+    return [GuardCondition(expr, negated)]
+
+
+def _collect_rule_paths(
```
LucaKro, line 140: same here

krrood/src/krrood/entity_query_language/rdr/backward_inference.py
LucaKro: i agree

LucaKro (changes requested):
i feel like naming in this PR is a bit off here

krrood/src/krrood/entity_query_language/rdr/backward_inference.py
AbdelrhmanBassiouny: Yeah I think that would be a natural follow up to make a subsumption classification algorithm over the definitions of concepts (their DNF).

krrood/src/krrood/entity_query_language/rdr/backward_inference.py
AbdelrhmanBassiouny: replaced with applies/is applied

krrood/src/krrood/entity_query_language/rdr/backward_inference.py
AbdelrhmanBassiouny: replaced with ConclusionSufficientConditionSets

krrood/src/krrood/entity_query_language/rdr/backward_inference.py
AbdelrhmanBassiouny: replaced with `get_conclusion_sufficient_conditions_from_a_rule_tree`

krrood/src/krrood/entity_query_language/rdr/condition_resolver.py
AbdelrhmanBassiouny: Done

krrood/src/krrood/entity_query_language/rdr/condition_resolver.py
AbdelrhmanBassiouny: Renamed to `as_expression` and moved to the GuardCondition class. Also renamed the existing expression field to `original_expression`

krrood/src/krrood/entity_query_language/rdr/condition_resolver.py
AbdelrhmanBassiouny: Renamed to `TargetSufficientConditionsBasedResolver`

krrood/src/krrood/entity_query_language/rdr/backward_inference.py
AbdelrhmanBassiouny: Done

krrood/src/krrood/entity_query_language/rdr/backward_inference.py
AbdelrhmanBassiouny: It is because these are the only conclusion selectors that exist and will not be extended; it is the well-known branching logic (nested if, else if, adjacent if). That was my assumption when I left this as is.

But I think it would be cleaner to move it, so I will do it.

krrood/src/krrood/entity_query_language/rdr/utils.py
AbdelrhmanBassiouny: Removed from this pr, will refactor it in another coming pr as it doesn't have it's usage here actually

krrood/src/krrood/entity_query_language/rdr/backward_inference.py
AbdelrhmanBassiouny: Done, used external classes that link to the original classes instead of adding backward inference concern to the eql expressions which is an rdr concern

krrood/src/krrood/entity_query_language/rdr/backward_inference.py
AbdelrhmanBassiouny: Done , same comment as above

tomsch420 (approved):

LucaKro (approved):

PR #519, opened 2026-08-06
Improve GraphOfBoundingBoxes path finding: weighted Dijkstra + shortc…

tomsch420:
…utting

path_from_to previously ran an unweighted (hop-count) Dijkstra search, so a detour through a few huge boxes could beat a longer chain of small boxes even when it was far longer in real distance. Edges are now weighted by the Euclidean distance between adjacent boxes' centers, and Dijkstra minimizes that instead of hop count.

The resulting waypoints are also shortcut afterwards: any waypoint that a straight line can bypass without leaving free space is dropped, removing unnecessary zigzags through the box partition.

Made with the help of Claude.

## GCS Path-Finding: Before/After Across CRAM's URDF Environments

_Old = unweighted Dijkstra, no shortcutting. New = weighted Dijkstra + shortcutting (this PR). 20 randomly sampled free-space point pairs per environment, 2-D navigation map over each environment's full XY footprint at a fixed 1.2 m height band._

| Environment | Nodes / Edges | Time (mean, old→new) | Hops (mean, old→new) | Distance (mean, old→new) |
|---|---|---|---|---|
| Apartment | 487 / 494 | 0.070s → 0.162s | 10.85 → 1.90 (**−82%**) | 9.06m → 7.54m (**−17%**) |
| Apartment (no walls) | 294 / 300 | 0.230s → 0.264s | 11.00 → 1.55 (**−86%**) | 6.19m → 5.98m (**−3%**) |
| Kitchen | 128 / 129 | 0.043s → 0.063s | 6.25 → 2.05 (**−67%**) | 4.60m → 3.84m (**−17%**) |
| Kitchen (small) | 63 / 65 | 0.008s → 0.019s | 5.05 → 2.30 (**−54%**) | 5.19m → 4.43m (**−15%**) |
| Broken Kitchen | 175 / 177 | 0.046s → 0.082s | 10.55 → 2.25 (**−79%**) | 6.60m → 4.56m (**−31%**) |
| IAI Kitchen | — | Failed: no collision geometry (all-visual URDF, unrelated to this PR) | — | — |
| **Average (5 envs)** | — | 0.079s → 0.118s (**1.5×**) | 8.74 → 2.01 (**−77%**) | 6.33m → 5.27m (**−17%**) |

Consistent story across every real environment: **~77% fewer waypoints and ~17% shorter paths, at a ~1.5× time cost** (all still sub-second).

**Caveats:**
- IAI Kitchen has no `has_collision()` bodies at all in this checkout — a pre-existing property of the asset, not something this PR touches.
- The height band (`[min_z, min_z+1.2m]`) is auto-derived from each environment's lowest collision geometry, not hand-tuned per room — e.g. Apartment's band came out as `[-0.70, 0.50]`, which may include some below-floor structure. Good enough for a relative before/after comparison, not a claim of an ideal navigation slice.
- Profiling shows the dominant remaining cost in `path_from_to` (both old and new) is `node_of_point`'s O(n) linear scan with uncached symbolic bounding-box checks — pre-existing, not introduced by this PR.

### DIFF
--- a/semantic_digital_twin/src/semantic_digital_twin/world_description/geometry.py
+++ b/semantic_digital_twin/src/semantic_digital_twin/world_description/geometry.py
@@ -840,9 +840,9 @@ class Mesh(Shape):
         """
         points = np.asarray([point.to_np()[:3] for point in points_3d], dtype=float)
         points = np.unique(points, axis=0)
-        assert (
-            len(points) >= 3
-        ), "At least 4 unique points are required to define a 3D region."
+        assert len(points) >= 3, (
+            "At least 4 unique points are required to define a 3D region."
+        )
 
         centered_points = points - points.mean(axis=0, keepdims=True)
         assert np.any(centered_points), "Points must not be all identical."
@@ -1126,6 +1126,39 @@ class Bounds(Generic[T], SubClassSafeGeneric):
     The corner with the largest coordinate on every axis.
     """
 
+    def clip_segment(
+        self, start: np.ndarray, direction: np.ndarray
+    ) -> Optional[SimpleInterval]:
+        """
+        Clip the parametrized segment ``start + t * direction`` (``t`` in ``[0, 1]``)
+        against this region, using the slab method.
+
+        Assumes ``lower``/``upper`` are plain numeric arrays, as returned by
+        :meth:`BoundingBox.to_array_bounds`.
+
+        :param start: The segment's start point.
+        :param direction: The vector from the segment's start to its end.
+        :return: The sub-interval of ``t`` for which the segment lies inside this
+            region, or None if the segment misses it entirely.
+        """
+        t_min, t_max = 0.0, 1.0
+        for coordinate, delta, lower, upper in zip(
+            start, direction, self.lower, self.upper
+        ):
+            if abs(delta) < 1e-12:
+                if coordinate < lower or coordinate > upper:
+                    return None
+                continue
+            t_enter = (lower - coordinate) / delta
+            t_exit = (upper - coordinate) / delta
+            if t_enter > t_exit:
+                t_enter, t_exit = t_exit, t_enter
+            t_min = max(t_min, t_enter)
+            t_max = min(t_max, t_exit)
+            if t_min > t_max:
+                return None
+        return SimpleInterval.from_data(t_min, t_max, Bound.CLOSED, Bound.CLOSED)
+
 
 @dataclass(eq=False)
 class BoundingBox:
@@ -1212,12 +1245,9 @@ class BoundingBox:
 
         :return: The corners, in the same frame as ``origin``.
         """
-        lower = np.array(
-            [self.x_interval.lower, self.y_interval.lower, self.z_interval.lower]
-        )
-        upper = np.array(
-            [self.x_interval.upper, self.y_interval.upper, self.z_interval.upper]
-        )
+        x, y, z = self.x_interval, self.y_interval, self.z_interval
+        lower = np.array([x.lower, y.lower, z.lower])
+        upper = np.array([x.upper, y.upper, z.upper])
         return Bounds(lower, upper)
 
     def to_point3_bounds(self) -> Bounds[Point3]:
@@ -1226,16 +1256,17 @@ class BoundingBox:
 
         :return: The corners, in the same frame as ``origin``.
         """
+        x, y, z = self.x_interval, self.y_interval, self.z_interval
         lower = Point3(
-            self.x_interval.lower,
-            self.y_interval.lower,
-            self.z_interval.lower,
+            x.lower,
+            y.lower,
+            z.lower,
             reference_frame=self.origin.reference_frame,
         )
         upper = Point3(
-            self.x_interval.upper,
-            self.y_interval.upper,
-            self.z_interval.upper,
+            x.upper,
+            y.upper,
+            z.upper,
             reference_frame=self.origin.reference_frame,
         )
         return Bounds(lower, upper)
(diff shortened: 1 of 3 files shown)


semantic_digital_twin/src/semantic_digital_twin/world_description/graph_of_convex_sets/boxes.py
```diff
@@ -40,6 +40,93 @@
 
 logger = logging.getLogger(__name__)
 
+Coordinates3D = Tuple[float, float, float]
```
tomsch420: Don't do that, use point3

semantic_digital_twin/src/semantic_digital_twin/world_description/graph_of_convex_sets/boxes.py
```diff
@@ -40,6 +40,93 @@
...
+    Edge payload connecting two adjacent bounding boxes in a
+    :class:`GraphOfBoundingBoxes`.
+    """
+
+    intersection: BoundingBox
+    """
+    The region where the two adjacent boxes overlap or touch.
+    """
+
+    weight: float
```
tomsch420: Weight seems like a weird name here if it's just the euclidian discstance

semantic_digital_twin/src/semantic_digital_twin/world_description/graph_of_convex_sets/boxes.py
```diff
@@ -111,13 +200,19 @@ def _intersection_box(a_min, a_max, b_min, b_max):
                 HomogeneousTransformationMatrix(reference_frame=self.world.root),
             )
 
+        def _center(mn, mx):
```
tomsch420: Is this a method that takes the bounds of a bounding box ? If so, it should belong there

semantic_digital_twin/src/semantic_digital_twin/world_description/graph_of_convex_sets/boxes.py
```diff
@@ -111,13 +200,19 @@ def _intersection_box(a_min, a_max, b_min, b_max):
...
+            return ((mn[0] + mx[0]) / 2, (mn[1] + mx[1]) / 2, (mn[2] + mx[2]) / 2)
+
+        def _distance(a, b):
```
tomsch420: If you use point3 this method isn't needed

tomsch420 (commented):
Also make an experiment that you do not commit but measure the following from: load the apartment lab and randomly sample points from the free space using probabilistic_model uniform_measure_of_event or so. Compare for random pairs of the points the time it takes to find the paths, the length in hops and the length in meters between the old and new implementation

semantic_digital_twin/src/semantic_digital_twin/world_description/graph_of_convex_sets/boxes.py
```diff
@@ -41,6 +45,82 @@
 logger = logging.getLogger(__name__)
 
 
+def _segment_box_overlap(
```
tomsch420: Use existing classes. Also it looks like this could belong somewhere

semantic_digital_twin/src/semantic_digital_twin/world_description/graph_of_convex_sets/boxes.py
```diff
@@ -41,6 +45,82 @@
...
+        if t_enter > t_exit:
+            t_enter, t_exit = t_exit, t_enter
+        t_min = max(t_min, t_enter)
+        t_max = min(t_max, t_exit)
+        if t_min > t_max:
+            return None
+    return t_min, t_max
+
+
+def _covers_unit_interval(
```
tomsch420: Also looks like it belongs somewhere. Also use existing classes

semantic_digital_twin/src/semantic_digital_twin/world_description/graph_of_convex_sets/boxes.py
```diff
@@ -241,19 +336,95 @@ def path_from_to(self, start: Point3, goal: Point3) -> Optional[List[Point3]]:
...
+                result.append(waypoints[index - 1])
+                anchor_index = index - 1
+        result.append(waypoints[-1])
+        return result
+
+    def _segment_is_collision_free(
+        self,
+        start: Point3,
+        end: Point3,
+        node_bounds: List[
```
tomsch420: Use existing classes, not tuples

sunava (approved):

semantic_digital_twin/src/semantic_digital_twin/world_description/graph_of_convex_sets/boxes.py
```diff
@@ -51,7 +76,7 @@ class GraphOfBoundingBoxes(GraphOfConvexSets):
     node is a box; every edge represents the adjacency between two boxes.
     """
 
-    graph: rx.PyGraph[BoundingBox] = field(
+    graph: rx.PyGraph[BoundingBox, BoundingBoxAdjacency] = field(
```
LucaKro, line 79: does this mean the graph is a lot bigger now, because we have more nodes?

semantic_digital_twin/src/semantic_digital_twin/world_description/geometry.py
```diff
@@ -1126,6 +1126,39 @@ class Bounds(Generic[T], SubClassSafeGeneric):
     The corner with the largest coordinate on every axis.
     """
 
+    def clip_segment(
+        self, start: np.ndarray, direction: np.ndarray
```
LucaKro: typing? can this be spatial types?

LucaKro (changes requested):
very nice

semantic_digital_twin/src/semantic_digital_twin/world_description/graph_of_convex_sets/boxes.py
tomsch420: No, the second thing is the edge type

semantic_digital_twin/src/semantic_digital_twin/world_description/geometry.py
tomsch420: Unfortunately not, converting to spatial types here is very slow and this runs very often

LucaKro (approved):

PR #522, opened 2026-08-07
Fix mesh temp file leak

ichumuh:
I had 113k meshes in my /tmp. How many do you have?

### DIFF
--- a/semantic_digital_twin/src/semantic_digital_twin/world_description/geometry.py
+++ b/semantic_digital_twin/src/semantic_digital_twin/world_description/geometry.py
@@ -5,11 +5,11 @@ import logging
 import math
 import os
 import shutil
-import tempfile
 from abc import ABC, abstractmethod
 from copy import deepcopy
 from dataclasses import dataclass, field, fields
 from functools import cached_property
+from pathlib import Path
 
 import numpy as np
 import trimesh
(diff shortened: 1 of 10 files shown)


LucaKro (approved):

semantic_digital_twin/src/semantic_digital_twin/world_description/geometry.py
```diff
@@ -722,9 +723,28 @@ def from_trimesh(
         scale: Optional[Scale] = None,
         uv: Optional[np.ndarray] = None,
         texture_file_path: Optional[str] = None,
-        dirname: str = "/tmp",
+        directory: Optional[Path] = None,
```
tomsch420, line 726: i think rebooting is normal behavior, therefore i think having a default to /tmp is the correct call here

tomsch420 (changes requested):

semantic_digital_twin/src/semantic_digital_twin/world_description/geometry.py
ichumuh: its a different parameter now. they old one decided the abs path where to place the files, the new one defines only the name of the directory inside /tmp.

ichumuh: updated the doc str to make that clear

tomsch420 (approved):

PR #470, opened 2026-07-23
[SemDT][Experiments] Re experiments

tomsch420:
This pull request introduces several enhancements and additions across the codebase, focusing on experiment reproducibility, feature completeness for Mujoco geometry parsing, and improved modularity in RoboCasa dataset loading. The most important changes are summarized below.

**Experiment reproducibility and documentation:**

* Added a new module `complement_worst_case_experiment.py` that demonstrates and measures the exponential worst-case complexity of the `Event.complement` operation, including both adversarial and control constructions, and outputs results as captioned Typst figures.
* Added a `render_figure` method to `TypstRenderer` in `experiment_definitions.py` to wrap tables with captions for better documentation and presentation.

**Mujoco geometry parsing improvements:**

* Added support for parsing ellipsoid geometries in MJCF by approximating them as scaled icosphere meshes, filling a previous gap where ellipsoids were not handled.
* Fixed the parsing of light directionality in `parse_light` to use the correct field (`mujoco_light.directional`) for improved compatibility with Mujoco.

**RoboCasa dataset loader modularity:**

* Refactored the RoboCasa loader to expose `build_task_environment` and `load_task_from_environment` as separate methods, allowing users to access and reuse the raw environment before parsing it into a `RoboCasaTask`. This supports workflows that require access to ground-truth environment state. [[1]](diffhunk://#diff-5246d610677533bae385158674cbd1e06a8e38998ad20d33dd6e16a6f3755447R321-R348) [[2]](diffhunk://#diff-5246d610677533bae385158674cbd1e06a8e38998ad20d33dd6e16a6f3755447R364-R374)

**Minor improvements and fixes:**

* Improved docstrings and formatting for clarity in various methods, including MJCF adapter and RoboCasa loader. [[1]](diffhunk://#diff-b54a6768c334bb072675452db7819d4ff0b939f310ad90ea8eb2825b9f55cfe0L260-R267) [[2]](diffhunk://#diff-5246d610677533bae385158674cbd1e06a8e38998ad20d33dd6e16a6f3755447L133-R134)
* Added the `trimesh` import required for mesh creation in MJCF parsing.

PR #523, opened 2026-08-07
[RE][SemDT][Experiments] Random events dataset surveys, region approximation, and RoboCasa export

tomsch420:
Completes the migration of the random-events experiments off the `re-experiments`
branch. #470 already landed the scalability and complement-worst-case
experiments, and #476 the propositional-where translation; this is the rest.

### Reading and surveying the conditions task suites state

Under `random_events_experiments/task_condition_survey/`:

- `task_conditions` — the vocabulary the surveys share: what a condition is made
  of, which kind of predicate each part is, and what it relates.
- `condition_extraction` — recovers the EQL condition behind a task's success
  criterion from the ordinary Python it is written in.
- `robocasa_condition_survey`, `mani_skill_condition_survey` — for suites that
  compute and return an outcome.
- `rlbench_condition_survey` — for a suite that instead registers the
  requirements the outcome is composed of.
- `condition_event_translation` — decides whether a stated condition is
  propositional enough to become a random event at all, counting a quantified one
  as first order rather than grounding it.
- `set_kind_survey` — the experiment entry point.

A suite's task definitions are located rather than imported, so reading them does
not require the simulator the suite drives to be installed.

### Supporting library changes

`random_events`: every set in the sigma algebra has a **size** — the length of an
interval, the number of elements of a set of elements, and the product of those
over the variables of a set of tuples. A composite set is made disjoint before
its parts are summed, so a region covered by several simple sets is counted once.

`semantic_digital_twin`: a bounding box has a **volume**.

`experiment_definitions`: a table can record its rows as a JSON manifest
alongside the table presenting them, a result reported within another contributes
its own values and not just its own columns, and `TypstRenderer` renders a cell —
a named quality as its label, a quantity at the reported precision, and a
measurement the experiment did not establish as absent rather than as zero. The
row-type assertion is replaced by `RowsOfDifferingTypes` and
`RowIsNotAnExperimentResult`.

### Two further experiments

`curved_region_approximation` — a predicate thresholding a Euclidean distance
describes a ball, and no box is a ball, but an `Event` is a *union* of boxes, so
the fraction of the ball its representation captures is a consequence of how many
boxes it spends rather than a limit of the representation. The covering is always
a covering one, so a condition represented this way is never satisfied by fewer
states than it should be.

`robocasa_set_export` — exports RoboCasa's own ground truth for what a scene or
task admits. RoboCasa states this in whatever form its implementation happens to
use: a placement region as sampler parameters, a gripper exclusion zone as a body
position and threshold, a fixture's state as a normalized MuJoCo joint position,
a device's state as a dictionary entry. The two Cartesian tracks are attached to
a `World` as `Region`s in the same shape model the rest of that world uses,
rather than read off RoboCasa's internal storage.

## Results

All three experiments run end to end locally.

**`set_kind_survey`**, against the installed RoboCasa, ManiSkill and RLBench:

| Dataset | Set Kind | Conditions | % | First Order | Avg Simple Sets | Avg IoU |
|---|---|---:|---:|---:|---:|---:|
| RoboCasa | Hybrid | 241 | 76.03 | 55 | 1.26 | 0.83 |
| RoboCasa | Discrete | 49 | 15.46 | 3 | 1.26 | 1.00 |
| RoboCasa | Continuous | 26 | 8.20 | 3 | 1.00 | 0.63 |
| RoboCasa | Undetermined | 1 | 0.32 | 0 | 1.00 | -- |
| ManiSkill | Continuous | 18 | 56.25 | 0 | 1.11 | 0.84 |
| ManiSkill | Undetermined | 9 | 28.12 | 0 | 1.00 | -- |
| ManiSkill | Hybrid | 5 | 15.62 | 0 | 1.00 | 0.88 |
| RLBench | Discrete | 56 | 70.89 | 0 | 1.00 | 1.00 |
| RLBench | Continuous | 20 | 25.32 | 0 | 1.00 | 1.00 |
| RLBench | Hybrid | 3 | 3.80 | 0 | 1.00 | 1.00 |

**`curved_region_approximation`**:

| Region | Subdivisions | Simple Sets | IoU |
|---|---:|---:|---:|
| Ball | 1 | 1 | 0.52 |
| Ball | 2 | 1 | 0.52 |
| Ball | 4 | 7 | 0.64 |
| Ball | 8 | 50 | 0.77 |
| Ball | 16 | 245 | 0.86 |
| Disc | 1 | 1 | 0.79 |
| Disc | 2 | 1 | 0.79 |
| Disc | 4 | 3 | 0.84 |
| Disc | 8 | 4 | 0.90 |
| Disc | 16 | 8 | 0.94 |

The single-box rows are the analytically known values, which checks the rest: a
unit ball against its bounding box is `(4π/3) / 8 = 0.5236`, and a unit-radius
disc of height 1 is `π / 4 = 0.7854`. Reading the ball's column, 1 → 245 simple
sets buys 0.52 → 0.86 — worth weighing against the survey above, where real
conditions need 1.0 to 1.26 simple sets on average, so the expensive end of this
table is not where they sit.

**`robocasa_set_export`**, against the installed RoboCasa: 14 placement regions
attached, 15 placement region fidelities measured, 4 forbidden zones attached, 3
joint openness intervals and 11 discrete device state entries written.

## Open questions for review

**The survey's percentages are over the conditions it could read, not over each
suite.** `stated_conditions()` drops any condition not read in full, which is the
right mechanism — a partially read condition would poison the kind counts — but
the denominators differ from the suite sizes:

| Suite | Tasks found | Fully read | Surveyed |
|---|---:|---:|---:|
| RoboCasa | 352 | 318 | 317 |
| ManiSkill | 46 | 32 | 32 |
| RLBench | 106 | 79 | 79 |

The caption currently says "of that dataset's conditions", which reads as all of
them. Should the read rate become a column, or should the caption name its
denominator?

**ManiSkill's 28% Undetermined** may be a genuine property of how ManiSkill
phrases its conditions, or a gap in `ManiSkillPredicateGeometry`. RoboCasa has 1.

**`robocasa_set_export` has no tests.** It is migrated as it stands on
`re-experiments`. This is against AGENTS.md; it is also the one module needing a
real RoboCasa install to run, so whatever tests it gets will need to mock the
loader or skip in CI.

**It also duplicates the manifest logic this PR adds elsewhere** — it carries its
own module-level `write_manifest` and `as_json_value` rather than using
`ExperimentsTable.write_manifest`. The one real difference is that its version
handles numpy scalars, which the table's does not. Deduplicating means teaching
`ExperimentsTable.as_json_value` about `np.generic` and dropping both functions.


ichumuh (changes requested):
labels when?

experiments/src/experiments/random_events_experiments/robocasa_set_export.py
```diff
@@ -0,0 +1,878 @@
...
+    if isinstance(value, enum.Enum):
+        return value.name
+    if isinstance(value, np.generic):
+        return value.item()
+    if isinstance(value, dict):
+        return {key: as_json_value(entry) for key, entry in value.items()}
+    return value
+
+
+def main():
```
Narenvasant: doc string

experiments/src/experiments/random_events_experiments/task_condition_survey/condition_extraction.py
```diff
@@ -0,0 +1,1073 @@
...
+            the body.
+        """
+        return exists(self.objects, self.body)
+
+
+# %% reading a success condition
+
+
+@dataclass
+class ConditionExtractor:
```
Narenvasant, line 167: such a huuuggggeeee class. can this be split or is it ok to be this big?

LucaKro (approved):

random_events/src/random_events/product_algebra.py
```diff
@@ -125,6 +126,23 @@ def contains(self, item: Tuple) -> bool:
...
+        A variable left out of the event is left out of the product, so the size is the
+        one taken over the variables the event constrains.
+
+        :return: The size of this event, which is zero as soon as one variable is
+            assigned nothing of size and infinite when one is unbounded.
+        """
+        factors = [assignment.size for assignment in self.assignments]
+        if any(factor == 0.0 for factor in factors):
+            return 0.0
+        return math.prod(factors)
```
Narenvasant, line 148: i think empty list for math.prod will return 1 not  0. could you add doc and/or even a test for this case?

Narenvasant (changes requested):

experiments/src/experiments/random_events_experiments/task_condition_survey/condition_extraction.py
tomsch420: nah, its fine for the experiment

random_events/src/random_events/product_algebra.py
tomsch420: you are right, its fixed

ichumuh (approved):

Narenvasant (approved):
Gucci

PR #524, opened 2026-08-07
[RE][SemDT][Experiments] Set size, bounding box volume, curved region approximation, and RoboCasa export

tomsch420:
Stacked on #523 — review that one first; this PR's diff against `main` will
include it until it lands.

Completes the migration of the random-events experiments off `re-experiments`.

## What this adds

**`random_events`: every set in the sigma algebra has a size.** The length of an
interval, the number of elements of a set of elements, and the product of those
over the variables of a set of tuples. A composite set is made disjoint before
its parts are summed, so a region covered by several simple sets is counted once,
and a variable left out of a simple event is left out of the product.

**`semantic_digital_twin`: a bounding box has a volume.**

**`curved_region_approximation`**, which uses both. A predicate thresholding a
Euclidean distance describes a ball, and no box is a ball — but an `Event` is a
*union* of boxes, so the fraction of the ball its representation captures is a
consequence of how many boxes it spends, not a limit of the representation. The
covering is always a covering one, so a condition represented this way is never
satisfied by fewer states than it should be.

**`robocasa_set_export`**, which exports RoboCasa's own ground truth for what a
scene or task admits. RoboCasa states this in whatever form its implementation
happens to use — a placement region as sampler parameters, a gripper exclusion
zone as a body position and threshold, a fixture's state as a normalized MuJoCo
joint position, a device's state as a dictionary entry. The two Cartesian tracks
are attached to a `World` as `Region`s in the same shape model the rest of that
world uses, rather than read off RoboCasa's internal storage.

## Results

Both experiments run end to end locally.

`curved_region_approximation`:

| Region | Subdivisions | Simple Sets | IoU |
|---|---:|---:|---:|
| Ball | 1 | 1 | 0.52 |
| Ball | 2 | 1 | 0.52 |
| Ball | 4 | 7 | 0.64 |
| Ball | 8 | 50 | 0.77 |
| Ball | 16 | 245 | 0.86 |
| Disc | 1 | 1 | 0.79 |
| Disc | 2 | 1 | 0.79 |
| Disc | 4 | 3 | 0.84 |
| Disc | 8 | 4 | 0.90 |
| Disc | 16 | 8 | 0.94 |

The single-box rows are the analytically known values, which checks the rest: a
unit ball against its bounding box is `(4π/3) / 8 = 0.5236`, and a unit-radius
disc of height 1 is `π / 4 = 0.7854`.

Reading the ball's column: 1 → 245 simple sets buys 0.52 → 0.86. Worth weighing
against what #523 reports — the conditions actually found in the three suites
need 1.0 to 1.26 simple sets on average, so the expensive end of this table is
not where real conditions sit.

`robocasa_set_export` against the installed RoboCasa: 14 placement regions
attached, 15 placement region fidelities measured, 4 forbidden zones attached,
3 joint openness intervals and 11 discrete device state entries written.

## Two things for review

**`robocasa_set_export` has no tests.** It is migrated as it stands on
`re-experiments`. This is against AGENTS.md and I have not written them here; the
export is also the one module that needs a real RoboCasa install to run, so
whatever tests it gets will need to mock the loader or skip in CI. Flagging it
rather than burying it.

**It also duplicates the manifest logic #523 adds.** It carries its own
module-level `write_manifest` and `as_json_value` rather than using
`ExperimentsTable.write_manifest`. The one real difference is that its version
handles numpy scalars, which the table's does not. Deduplicating means teaching
`ExperimentsTable.as_json_value` about `np.generic` and dropping both functions —
straightforward, but it changes a file already under review in #523 and I would
not want to do it without a test for the numpy case. Happy to do it either here
or as a follow-up, whichever you prefer.

## Notes

- Tests: 99 passing across `random_events_test`, the bounding box tests, and
  `test_curved_region_approximation`. Nothing new for the export, per above.
- `size` is a new abstract property on `AbstractSimpleSet`. Its only three
  subclasses in the workspace — `SimpleInterval`, `SetElement`, `SimpleEvent` —
  all implement it.
- The two ORM commits are regenerator output: six data access objects for the
  region approximation, ten for the export.


tomsch420:
Consolidated into #523, which now carries all of this branch's commits unchanged. Closing to keep the migration in a single pull request.

PR #525, opened 2026-08-07
World specification fresh worlds

LucaKro:
currently WorldSpecifications hold a live world. I did not consider that this means, if you spawn multiple of the same world and then merge then with each other (for example when describing a room layout of a building, and then spawning them on top of each other for different floors) , that this wont work since they contain the same set of IDs.
This PR fixes it


semantic_digital_twin/src/semantic_digital_twin/adapters/mjcf.py
```diff
@@ -89,10 +90,27 @@ def __post_init__(self):
...
-        self.world = World()
 
     @classmethod
-    def from_xml_string(cls, xml_string: str) -> "MJCFParser":
+    def from_file(
+        cls,
+        file_path: str,
+        prefix: Optional[str] = None,
+        mimic_joints: Optional[Dict[str, str]] = None,
+    ) -> "MJCFParser":
```
ichumuh: type

semantic_digital_twin/src/semantic_digital_twin/api.py
```diff
@@ -1167,17 +1168,17 @@ class WorldSpecification:
...
-    (:meth:`to_domain_object`) merges every robot into the environment, then spawns all
-    starting objects, and returns the augmented environment world.
+    The environment is described by the parser that builds it (obtained from a model file
+    with :meth:`from_urdf`, :meth:`from_mjcf` or :meth:`from_gazebo`). Applying it
+    (:meth:`to_domain_object`) parses the environment anew, merges every robot into it,
+    then spawns all starting objects, and returns the augmented environment world.
     """
 
-    world: World
+    world_parser: WorldModelParser | None
```
ichumuh: default value

ichumuh (approved):

tomsch420 (approved):

semantic_digital_twin/src/semantic_digital_twin/adapters/world_model_parser.py
```diff
@@ -0,0 +1,46 @@
...
+    Declares no fields, so each format keeps its own payload field (the description text
+    or the path it is read from) as its first constructor parameter.
+
+    Every parse produces freshly created world entities, so a parser is the way to obtain
+    a world that shares no identifiers with any previously parsed one.
+    """
+
+    @classmethod
+    @abstractmethod
+    def from_file(cls, file_path: str, prefix: Optional[str] = None) -> Self:
```
Tigul, line 28: Finally 🙏

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
```diff
@@ -163,21 +180,25 @@ def parent_connection_specification(cls) -> ConnectionSpecification:
 
     @classmethod
     @abstractmethod
-    def get_default_root_specification(
+    def get_default_root_entity_specification(
```
Tigul: How about get_default_body_specification as name? I think that is more clear

LucaKro: hmm but there exist the same method for regions, so then we would have two methods that do essentially the same thing being named differently. right now we can rely on the contract that any HasRootKinematicStructureEntity has a method get_default_root_entity_specification.

with the above change we would lose that

Tigul: True in that case the name doesn't make sense, I just feel that  ...root_entity_specification is still a bit ambiguous and can be easily misinterpreted

LucaKro: do you think get_default_root_kinematic_structure_entity_specification would be better?

I personally think root entity is unambiguous, especially since i also added another disclaimer to the docstring of the method :sweat_smile:

Tigul: I guess that would be worse :sweat_smile: But in my opinion the two methods sounds still rather similar while they to different things

Tigul: Maybe @tomsch420 or @ichumuh have another idea on this?

tomsch420: What was wrong with the old method name?

ichumuh: i think the old name is better than the new one and if you that wasn't specific enough for you, i'd go all the way with get_default_root_kinematic_structure_entity_specification

LucaKro: renamed to get_default_root_kinematic_structure_entity_specification

PR #527, opened 2026-08-08
Keep a world's modification history replayable, and rebuild it when it is not

tomsch420:
## The problem

A stored world is rebuilt by replaying its recorded modification history onto an empty world. That only works if the history starts from nothing — and for **541 of 8,857 worlds** in the shared database on `neem-4`, it doesn't. Their history opens with a *removal* of a connection nothing had added, so the replay raises immediately and the world cannot be read back at all:

```
watering_can_1   modifications applied: 1
  1  RemoveConnectionModification  parent_id=0450bcc1…  WorldEntityNotFoundError

stored order:  5907 RemoveConnection            <- first
               5908 AddKinematicStructureEntity
               5909 AddKinematicStructureEntity
               5910 AddConnection
```

## Where the removal comes from

`move_branch_to_new_world` cleaned each connection up in the **new** world rather than the old one:

```python
self.remove_kinematic_structure_entity(connection.parent)
self.remove_kinematic_structure_entity(connection.child)
new_world.remove_connection(connection)   # <- new_world is still empty
new_world.add_connection(connection)
```

`remove_connection` deliberately doesn't check membership first — its docstring explains why — so this silently succeeded on an empty world and left a removal as the first thing its history knows about.

Only the multi-body branch does this. The `if not child_bodies:` branch only ever adds, which is why single-body worlds always round-tripped fine and every affected world has connections.

## Two changes, because there are two problems

**`world.py`** — remove the connection from `self`, the world that actually holds it. Stops new worlds being written this way.

**`orm/model.py`** — that does nothing for worlds already stored. They aren't missing anything: `from_domain_object` persists the entities, connections, DOFs and annotations alongside the history, and those describe the finished world directly. So `to_domain_object` now falls back to the stored contents when the history can't be replayed. Replay stays the primary path.

The fallback first releases the contents from the world an aborted replay was building — a replay that fails partway has already claimed some entities, and one may belong to only one world.

## Verification

Against the live database on `neem-4`, with entity/connection counts checked against what the database records, so "it loaded" isn't mistaken for "it loaded correctly":

| | Before | After |
|---|---|---|
| 60 worlds whose history starts with a removal | 0 / 60 | **60 / 60**, all correct shape |
| 40 healthy control worlds | 40 / 40 | **40 / 40**, unchanged |
| Fallback engaged on control worlds | — | **0** |

And reproduced from source — parsing `beds_grp.fbx`:

| | Unpatched | Patched |
|---|---|---|
| Multi-body worlds produced | 32 | 32 |
| Histories starting with a removal | **32** | **0** |
| Round-trip failures | **32** | **0** |

## What a reviewer should know

- **The repo's test suite has not been run.** The two regression tests here are written and syntax-checked but never executed; CI on this PR is the first real run.
- The affected worlds are identifiable without loading them — whether a world's first modification is a `Remove*` predicted the failure exactly across every world tested, and the 6.1% it predicts matched the 5.3–6.4% measured by sampling actual loads.
- `_detach_contents_from_any_world` reads `item._world` via `getattr`. It's guarded, but a public "is this attached to a world" predicate would be cleaner if one exists.
- A separate, unrelated defect (`WorldEntityWithIDNotInKwargs`) affects some worlds and is **not** addressed here — it behaves identically with and without this change.

https://claude.ai/code/session_016LqvKXfPQS1xU3vkD5a5ey


LucaKro:
why was this done from a cram2 branch, and not from your fork?

semantic_digital_twin/src/semantic_digital_twin/world.py
```diff
@@ -1767,7 +1767,14 @@ def move_branch_to_new_world(self, new_root: KinematicStructureEntity) -> World:
...
                 self.remove_kinematic_structure_entity(connection.parent)
                 self.remove_kinematic_structure_entity(connection.child)
-                new_world.remove_connection(connection)
+                # Clean the connection up in the world that actually holds it.
+                # Removing its parent and child above already dropped the edge
+                # from this world's kinematic structure, which is exactly the
+                # case `remove_connection` documents as still needing an explicit
+                # removal on our side. Doing this to `new_world` instead records
+                # a removal as the first entry of a history that has not added
+                # anything yet, and that history can no longer be replayed.
```
LucaKro: if thats the case, write a test for it that please

semantic_digital_twin/src/semantic_digital_twin/orm/model.py
```diff
@@ -60,16 +67,101 @@ def from_domain_object(cls, obj: World):
...
+
+        Such a world is not missing anything: `from_domain_object` stores the
+        entities, connections, degrees of freedom and annotations next to the
+        history, and those describe the finished world directly. So a history
+        that cannot rebuild the world is not fatal - the contents are used
+        instead, and the world still comes back.
+        """
+        try:
+            result = self._replay_modification_history()
+        except (WorldEntityNotFoundError, BrokenWorldModificationHistoryError):
```
LucaKro: i dont like this. why are we supporting both ways, when, in my opinion, _replay_modification_history should *always* suffice, and if it doesnt it means that we have a bug somewhere...

LucaKro (changes requested):
hmm not convinced

semantic_digital_twin/src/semantic_digital_twin/world.py
```diff
@@ -1013,8 +1017,11 @@ def remove_connection(self, connection: Connection) -> None:
             if we want to remove the parent or child from the world, before removing the connection from the world.
             In that case, rustworkx automatically removes the edge representing the connection, which results in
             self.is_connection_in_world returning False, even though we have not cleaned up the connection properly on
-            our side.
+            our side. The ownership the connection itself records survives that,
+            which is what makes it usable as the check here.
         """
+        if connection._world is not self:
```
LucaKro, line 1120: hmm actually thats not a bad idea...is there any way to add this to the atomic world modification decorator itself, so that its open closed for all current and future modifications? and if not can you also add this to the other cases, because it will be the same bug for all of them

LucaKro (changes requested):
noice

semantic_digital_twin/src/semantic_digital_twin/world.py
tomsch420: So i dont think it is generally possible unless i assume that the first argument of every such method is the WorldEntity, but i will add it to the respective methods

LucaKro (approved):

davidprueser (approved):

PR #529, opened 2026-08-09
[DevTools] Read upstream review threads from a fork via a dispatched Action

AbdelrhmanBassiouny:
This gives you the benefit of letting claude read reviews off cram2 pull requests using a GitHub Action so no unintended behaviour and no write access to claude needed. Safe, Useful, and Deterministic behaviour. 

Works cloud sessions as well 😉

In addition, now the plan-item-resolve skill uses this to get not just your fork own pr reviews but also includes cram2 pr reviews before making a plan to resolve the pr.


.claude/upstream_reviews/upstream_reviews.py
```diff
@@ -0,0 +1,1091 @@
...
+    only when something happened to parse that field.
+
+    ..note:: A reader needing more than the object itself cannot state this
+        contract, so :class:`PullRequestReviewSnapshot`, which is assembled from
+        threads gathered across several responses, does not implement it.
+    """
+
+    @classmethod
+    @abstractmethod
+    def from_json(cls, data: dict[str, Any]) -> JSONModel:
```
LucaKro, line 62: I know you dont want to depend on SubclassJsonSerializer here, but didnt you do sth like this class last time you did not want to depend on it?

.claude/upstream_reviews/upstream_reviews.py
```diff
@@ -0,0 +1,1091 @@
...
+class ThreadMarker(StrEnum):
+    """
+    The annotations a thread can carry in the report.
+    """
+
+    RESOLVED = "resolved"
+    OUTDATED = "outdated"
+
+
+UNKNOWN_LOGIN = "(unknown)"
```
LucaKro: I dont like these things being defined randomly in the middle of files.

.claude/upstream_reviews/upstream_reviews.py
LucaKro: also are you sure you dont want to depend on it? i feel like its just accumulating duplicated code

LucaKro: also now that i look at it., cant you just use from_json from subclassjsonserializer?
it should be able to deal with dataclasses like this out of the box without you needing to define your from_json methods for everything

LucaKro (changes requested):
depend on krrood? 👉 👈

.claude/upstream_reviews/upstream_reviews.py
AbdelrhmanBassiouny: Yes, we have discussed this. The upcoming PRs will move everything and test the krrood dependency.

.claude/upstream_reviews/upstream_reviews.py
AbdelrhmanBassiouny: would you say ThreadMarker is also in the middle of the file?

.claude/upstream_reviews/upstream_reviews.py
LucaKro: just for the record, we discussed this after i wrote the comment :D

.claude/upstream_reviews/upstream_reviews.py
AbdelrhmanBassiouny: moved to be a class var, and moved another variable to be at the top of the module.

LucaKro (approved):

tomsch420 (approved):

tomsch420:
Software Engineering with Git Domain Model Gaming

PR #533, opened 2026-08-10
fix incorrect assumption for pose calculation in coraplex

LucaKro:
currently the coraplex locations assume that the parent of the robot is the world root, while worldspecification add another entitiy for odom in between. this caused the  navigation poses to be wrong if the odom pose wasnt identical to the world pose.

### DIFF
--- a/coraplex/src/coraplex/locations/base.py
+++ b/coraplex/src/coraplex/locations/base.py
@@ -4,6 +4,7 @@ import logging
 from abc import abstractmethod
 from copy import deepcopy
 from dataclasses import dataclass, field
+from typing import cast
 
 from typing_extensions import (
     Callable,
@@ -90,7 +91,9 @@ class Location(Iterable[Pose]):
 
     def __iter__(self) -> Iterator[Pose]:
         test_world = deepcopy(self.world)
-        test_robot = test_world.get_semantic_annotations_by_type(type(self.robot))[0]
+        test_robot = cast(
+            AbstractRobot, test_world.get_semantic_annotation_by_id(self.robot.id)
+        )
         for validator in self.validators:
             validator.context = Context(
                 world=test_world,
@@ -105,7 +108,7 @@ class Location(Iterable[Pose]):
 
         for pose_candidate in self.generator:
 
-            test_robot.root.parent_connection.origin = pose_candidate
+            test_robot.set_root_pose(pose_candidate)
 
             test_world.collision_manager.clear_temporary_rules()
             test_world.collision_manager.add_temporary_rule(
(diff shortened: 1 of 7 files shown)


coraplex/src/coraplex/locations/base.py
```diff
@@ -90,7 +91,9 @@ def ground(self) -> Pose:
 
     def __iter__(self) -> Iterator[Pose]:
         test_world = deepcopy(self.world)
-        test_robot = test_world.get_semantic_annotations_by_type(type(self.robot))[0]
+        test_robot = cast(
```
tomsch420, line 94: why do you need to do that?

tomsch420 (changes requested):

coraplex/src/coraplex/locations/base.py
ichumuh: so pycharm doesn't say yellow

semantic_digital_twin/src/semantic_digital_twin/robots/robot_parts.py
```diff
@@ -811,6 +812,26 @@ def drive(self) -> Optional[WheeledDrive]:
...
+        ``odom``, for example) sit between that frame and the pose's own.
+
+        ..note:: A drive that cannot represent every degree of freedom applies only what
+            it can, so the root reaches ``pose`` only within the drive's own limits.
+
+        :param pose: The pose the robot's root should end up at.
+        """
+        connection = self.root.parent_connection
+        parent_kinematic_structure_entity = connection.parent
+        if pose.reference_frame is not parent_kinematic_structure_entity:
```
ichumuh, line 830: hm this becomes unnecessary after your other pr merges

ichumuh (approved):

semantic_digital_twin/src/semantic_digital_twin/robots/robot_parts.py
LucaKro: no it doesnt, because for now we only changed it for connection6dof, because you said drives are a little bit different no? or maybe i misunderstood

coraplex/src/coraplex/locations/base.py
LucaKro: yes it literally does nothing at runtime, it is just the canonical way of doing 
```python
test_robot: AbstractRobot = test_world.get_semantic_annotation_by_id(self.robot.id)
```

semantic_digital_twin/src/semantic_digital_twin/robots/robot_parts.py
ichumuh: hm yeah i said that, but apparently its needed anyway, maybe we should add this to those joints too

coraplex/src/coraplex/locations/base.py
LucaKro: in this version, ` test_world.get_semantic_annotation_by_id(self.robot.id)` would still be marked yellow, and only everything after the type annotation would be correct. with cast, nothing is yellow

tomsch420: lol, til

tomsch420 (approved):

PR #534, opened 2026-08-10
velocity limits

sunava:
Furniture joints (a cupboard door, a drawer e.g) often do not have a valocity limit so math would be None x 0.01 -> crashing 
fix: if joint ahas not top speed, skip it for the check "done moving"

### DIFF
--- a/giskardpy/src/giskardpy/motion_statechart/graph_node.py
+++ b/giskardpy/src/giskardpy/motion_statechart/graph_node.py
@@ -914,7 +914,8 @@ def velocity_convergence_expression(
     :param maximum_threshold: Upper bound for the per-degree-of-freedom velocity
         threshold.
     :param degrees_of_freedom: Degrees of freedom to check for convergence. Defaults to
-        every active degree of freedom in the world when ``None``.
+        every active degree of freedom in the world when ``None``. Those without an
+        upper velocity limit are skipped, since no threshold can be derived for them.
     :param minimum_time: Minimum elapsed control time before the expression can become
         true.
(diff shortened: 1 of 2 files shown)


giskardpy/src/giskardpy/motion_statechart/graph_node.py
```diff
@@ -933,6 +934,11 @@ def velocity_convergence_expression(
     ref = []
     symbols = []
     for dof in degrees_of_freedom:
+        if dof.limits.upper.velocity is None:
```
tomsch420, line 937: sounds like it should be a default value in the parsing

tomsch420 (changes requested):

giskardpy/src/giskardpy/motion_statechart/graph_node.py
ichumuh: no i think this is fine actually. None means they have no velocity limits so they also shouldn't be considered when checking for convergence.

ichumuh: there is no meaningful default velocity for furniture joints

ichumuh (approved):

giskardpy/src/giskardpy/motion_statechart/graph_node.py
sunava: @tomsch420 :dagger:

tomsch420: i see

tomsch420 (approved):

PR #536, opened 2026-08-10
[DevTools] Fix duplication of artifcats.

AbdelrhmanBassiouny:
New sessions can miss an already existing URL for the artifact (plan dashboard site) and create a new one, thus having duplicate artifacts. This fixes this but making them always check using a script the existing ones.


LucaKro (commented):
krrood dependency? 🥹

LucaKro (approved):

tomsch420 (approved):

tomsch420:
I think we need a software engineering domain model in KRROOD where all this stuff can go :D

PR #539, opened 2026-08-10
[Agents] New skill for creating a new plan item.

AbdelrhmanBassiouny:
Instead of first discussing a new idea, then telling Claude to add a new plan item and having to figure out where exactly and based on which branch or task, this automates that.

So just call the skill with your idea description, then it will discuss it with you and check what the best place for this task in the available plans is and if it even needs a new plan of its own.


.claude/skills/add-plan-item/check_scope_overlap.py
```diff
@@ -0,0 +1,348 @@
...
+    :return: The subset of *paths* that exist there.
+    """
+    listing = run_git(
+        repository_path, "ls-tree", "-r", "--name-only", branch, "--", *paths
+    )
+    found = {line for line in listing.splitlines() if line}
+    return {path for path in paths if path.rstrip("/") in found or path in found}
+
+
+def changed_paths_since_base(
```
LucaKro, line 270: are there possible superclasses to bundle all of those module level methods? 
I feel like a lot of these skills always have super many module level methods instead of classes

LucaKro (changes requested):

.claude/skills/add-plan-item/check_scope_overlap.py
AbdelrhmanBassiouny: I checked them , and these methods can maybe be in a Repository class, but than actually needs to package move because Repository is one of the things that will be duplicate if I add i here, because another consequence of non package structure is you cannor import from folders on same level as your folder and Repository class is defined in anther place.

LucaKro: okay :+1: so many todos for after the package move :laughing:

LucaKro (approved):

.claude/skills/add-plan-item/check_scope_overlap.py
AbdelrhmanBassiouny: Yes 😶‍🌫️

tomsch420 (approved):

PR #543, opened 2026-08-12
No ORMatic interfaces are commitable anymore

tomsch420:
## Summary

`ormatic_interface.py` files are generated by ORMatic and never reviewed, so they no longer belong in diffs. Every tracked one (`coraplex`, `experiments`, `giskardpy`, `segmind`, `semantic_digital_twin`, krrood test dataset) is now committed as an **empty placeholder**, with several layers keeping it that way:

- **`empty-ormatic-interface` pre-commit hook** — truncates any of these files back to empty whenever they get staged, so real generated content can never land in a commit.
- **CI check** (`check_generated_orm_interfaces_are_empty`) — fails the build if one is ever committed non-empty, backstopping commits made without pre-commit installed.
- **`scripts/regenerate_all_orm.py`** now marks each file it regenerates with git's skip-worktree bit, so your locally regenerated interface (needed for DB work) stays on disk and is *never* proposed for staging — not even by `git add -A`/`git commit -a`. Run `scripts/protect_generated_orm_interfaces.py` on its own to get that protection without regenerating.
- CI test runs already regenerate these files per-package (`generate_orm.py`) and krrood's `conftest.py` already regenerates its copy per test session — both ephemeral, unchanged.

**Net effect:** these files vanish from review, git never bugs you about local changes to them, and everyone still gets a real, working copy locally by running the regen script once.


Tigul (approved):

README.md
```diff
@@ -169,6 +169,15 @@ sudo apt install pre-commit
 pre-commit install
 ```
 
+`ormatic_interface.py` files are generated and never reviewed, so the repository tracks
```
AbdelrhmanBassiouny: Remove (DRY with contributing.rst)

AbdelrhmanBassiouny (changes requested):

AbdelrhmanBassiouny (approved):

AGENTS.md
```diff
@@ -3,6 +3,7 @@
 ## Avoid Behaviour
 - Avoid using global variables
 - Avoid accessing any ormatic_interface.py files. if there are issues regarding the ormatic interface run the script `scripts/regenerate_all_orm.py`. If it does not fix the issue, consider consulting the developer.
```
LucaKro, line 5: Do we even need this anymore?

LucaKro (changes requested):

AGENTS.md
tomsch420: yes, locally they will have content and claude shouldnt read that unless you want a token burning speedrun

ichumuh (approved):

LucaKro (approved):

PR #544, opened 2026-08-12
Fix/cartesian velocity constraints

AbdElRahmanFarhan:
## Problem

`CartesianPositionVelocityLimit` and `CartesianRotationVelocityLimit` measured the tip's motion against the **root**, not against where it is going:

- `add_translational_velocity_limit` bounded `d‖x_tip‖/dt` — the rate the tip recedes from the root *origin* — i.e. only the radial component. Motion **across** the root direction was unconstrained (overshoot `1/cos θ`, unbounded). On a path tangential to the base the constraint sees nothing.
- `add_rotational_velocity_limit` had the identical defect on the current-to-root angle, and its observation expression measured the same wrong quantity, so nothing detected it.

The existing `test_cartesian_position_velocity_limit` never caught this: it only checks that halving the limit increases the cycle count, and drives `base_footprint` radially outward from `odom_combined` — the one geometry where the old math happens to be correct.

## Fix

Measure the distance/angle to the **goal** the tip is driven toward instead of to the root. When the tip travels straight at its goal, that rate *is* its true (angular) speed, so the bound caps the Euclidean norm of the tip velocity exactly.

- **`constraint_builders.py`** — `add_translational_velocity_limit` / `add_rotational_velocity_limit` take an optional `frame_P_goal` / `frame_R_goal` and bound the rate toward it; added the missing zero-guard to the rotational limit for the arrival degeneracy.
- **`cartesian_tasks.py`** — new shared `CartesianVelocityLimitTask` base that **freezes the goal into the root frame** at bind time via a `ForwardKinematicsBinding` (so a goal given relative to a moving frame, e.g. the tip, does not travel with it). `CartesianPositionVelocityLimit` takes `goal_point`, `CartesianRotationVelocityLimit` takes `goal_orientation`, and `CartesianVelocityLimit` forwards both. The goal is **opt-in**; without it the old root-referenced behaviour is preserved, so existing tests are unaffected. Field docstrings now state the real guarantee.

## Verification (measured, not assumed)

On geometries where the tip moves across the root direction:

| | before | after |
|---|---|---|
| linear (tangential) | **0.399 m/s** vs 0.05 limit (8×) | **0.0500** |
| angular (offset) | **0.705 rad/s** vs 0.2 limit (3.5×) | **0.2000** |

**Arrival**: the run completes at the goal and the max speed across all ticks — including the final ones where the distance-to-goal zero guard drops the constraint — stays at the limit, so the trailing-reference fallback was not needed.

New tests in `TestVelocityLimitBoundsAbsoluteSpeed` assert absolute speed on across-the-root geometries for both limits and through arrival. TDD: the failing test proving the bug was written first.

### DIFF
--- a/giskardpy/src/giskardpy/motion_statechart/constraint_builders.py
+++ b/giskardpy/src/giskardpy/motion_statechart/constraint_builders.py
@@ -167,18 +167,30 @@ class GeometricConstraintBuilder:
         frame_P_current: Point3,
         max_velocity: sm.ScalarData,
         quadratic_weight: sm.ScalarData,
+        frame_P_goal: Optional[Point3] = None,
         max_violation: sm.ScalarData = np.inf,
         name: Optional[str] = None,
     ) -> None:
         """
-        Adds constraints to limit the translational velocity of frame_P_current.
+        Adds constraints to bound the tip's speed toward frame_P_goal.
+
+        The bounded quantity is the time derivative of the distance between frame_P_current
+        and frame_P_goal. When the tip travels straight at the goal, that derivative equals
+        the tip's actual speed, so the bound caps the Euclidean norm of the tip velocity.
+        Measuring against the root origin instead (frame_P_goal unset) only bounds the
+        radial component and leaves motion across the root direction unconstrained.
 
         Be aware that the velocity is relative to frame.
         :param frame_P_current: a vector describing a 3D point. It should depend on
             active dofs.
+        :param frame_P_goal: the point the tip is driven toward. Defaults to the root
+            origin, which only bounds the radial speed.
         :param max_violation: m/s
         """
-        translation_error = frame_P_current.norm()
+        if frame_P_goal is None:
+            translation_error = frame_P_current.norm()
+        else:
+            translation_error = frame_P_current.euclidean_distance(frame_P_goal)
         translation_error = sm.if_eq_zero(
             translation_error, sm.Scalar(0.01), translation_error
         )
@@ -198,19 +210,31 @@ class GeometricConstraintBuilder:
         frame_R_current: RotationMatrix,
         max_velocity: sm.ScalarData,
         quadratic_weight: sm.ScalarData,
+        frame_R_goal: Optional[RotationMatrix] = None,
         max_violation: sm.ScalarData = LargeNumber,
         name: Optional[str] = None,
     ) -> None:
         """
-        Add velocity constraints to limit the velocity of frame_R_current.
+        Add velocity constraints to bound the tip's angular speed toward frame_R_goal.
+
+        The bounded quantity is the time derivative of the angle between frame_R_current
+        and frame_R_goal. When the tip turns straight at the goal, that derivative equals
+        the tip's actual angular speed. Measuring against the root orientation instead
+        (frame_R_goal unset) leaves turning that keeps the current-to-root angle constant
+        unconstrained.
 
         Be aware that the velocity is relative to frame.
         :param frame_R_current: Rotation matrix describing the current rotation. It
             should depend on active dofs.
+        :param frame_R_goal: the orientation the tip is driven toward. Defaults to the
+            root orientation, which only bounds the current-to-root angle rate.
         :param max_velocity: rad/s
         """
-        root_Q_tipCurrent = frame_R_current.to_quaternion()
-        angle_error = root_Q_tipCurrent.to_axis_angle()[1]
+        if frame_R_goal is None:
+            angle_error = frame_R_current.to_quaternion().to_axis_angle()[1]
+        else:
+            angle_error = frame_R_current.rotational_error(frame_R_goal)
+        angle_error = sm.if_eq_zero(angle_error, sm.Scalar(0.01), angle_error)
         self.collection.add_velocity_constraint(
             upper_velocity_limit=max_velocity,
             lower_velocity_limit=-max_velocity,
--- a/giskardpy/src/giskardpy/motion_statechart/tasks/cartesian_tasks.py
+++ b/giskardpy/src/giskardpy/motion_statechart/tasks/cartesian_tasks.py
@@ -226,9 +226,6 @@ class CartesianPositionTrajectory(CartesianTask):
     )
     """Reference velocity for normalization in m/s."""
 
-    goal_reference_frame_P_current_target_point: Point3 = field(init=False, repr=False)
-    """Symbolic expression representing the current target point in the goal reference frame."""
-
     remaining_distance: FloatVariable = field(init=False, repr=False)
     """Distance left to travel along the trajectory, rewritten every control cycle."""
 
@@ -252,6 +249,21 @@ class CartesianPositionTrajectory(CartesianTask):
                 )
         return reference_frame
 
+    @cached_property
+    def goal_reference_frame_P_current_target_point(self) -> Point3:
+        """
+        Per-tick target point on the trajectory, riding ``look_ahead_distance`` ahead of the
+        tip along the path, expressed in the goal reference frame.
+
+        It is created lazily so a velocity limit can reference it before the trajectory is
+        built. Its value is registered and updated during execution.
+        """
+        target_point = Point3.create_with_variables(
+            "goal_reference_frame_P_current_target_point"
+        )
+        target_point.reference_frame = self.goal_reference_frame
+        return target_point
+
     def _goal_points_to_np(self):
         self._goal_points_np = np.array(
             [point.to_np()[:-1] for point in self.goal_points]
@@ -271,7 +283,7 @@ class CartesianPositionTrajectory(CartesianTask):
             trajectory the tip already is.
         """
         artifacts = NodeArtifacts()
-        self._init_goal_reference_frame_P_current_target_point(
+        self._register_goal_reference_frame_P_current_target_point(
             context.float_variable_data
         )
         self._init_remaining_distance(context.float_variable_data)
@@ -358,12 +370,25 @@ class CartesianPositionTrajectory(CartesianTask):
             ),
             sparse=False,
         )
+        self._bind_compiled_point_to_memory(context)
+
+    def _bind_compiled_point_to_memory(self, context: MotionStatechartContext):
+        """
+        Point the compiled tip expression at the live state buffers.
+
+        Registering float variables reallocates the data array, so this is repeated in
+        :meth:`on_start` once every node has been built and the array is final.
+        """
         self._compiled_goal_reference_frame_P_tip.bind_args_to_memory_view(
             0, context.world.state.positions
         )
         self._compiled_goal_reference_frame_P_tip.bind_args_to_memory_view(
             1, context.float_variable_data.data
         )
+
+    def on_start(self, context: MotionStatechartContext):
+        super().on_start(context)
+        self._bind_compiled_point_to_memory(context)
 
     def _update_trajectory_index(self, goal_reference_frame_P_tip_np: np.ndarray):
         """
@@ -440,19 +465,13 @@ class CartesianPositionTrajectory(CartesianTask):
         )
         return None
 
-    def _init_goal_reference_frame_P_current_target_point(
+    def _register_goal_reference_frame_P_current_target_point(
         self, float_variable_data: FloatVariableData
     ):
         """
-        Initialize the symbolic expression representing the current target point in the goal reference frame.
+        Register the current target point with the data store and seed its value.
         :param float_variable_data: The FloatVariableData instance to register the expression with.
         """
-        self.goal_reference_frame_P_current_target_point = Point3.create_with_variables(
-            "goal_reference_frame_P_current_target_point"
-        )
-        self.goal_reference_frame_P_current_target_point.reference_frame = (
-            self.goal_reference_frame
-        )
         float_variable_data.register_expression(
             self.goal_reference_frame_P_current_target_point
         )
--- a/giskardpy/src/giskardpy/motion_statechart/tasks/cartesian_tasks.py
+++ b/giskardpy/src/giskardpy/motion_statechart/tasks/cartesian_tasks.py
@@ -5,7 +5,7 @@ from dataclasses import field, dataclass
 from functools import cached_property
 
 import numpy as np
-from typing_extensions import ClassVar, List
+from typing_extensions import ClassVar, List, Optional
 
 import krrood.symbolic_math.symbolic_math as sm
 from giskardpy.motion_statechart.binding_policy import (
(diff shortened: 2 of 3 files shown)


giskardpy/src/giskardpy/motion_statechart/tasks/cartesian_tasks.py
```diff
@@ -358,13 +370,26 @@ def compile_current_point_on_tick(self, context: MotionStatechartContext):
             ),
             sparse=False,
         )
+        self._bind_compiled_point_to_memory(context)
+
+    def _bind_compiled_point_to_memory(self, context: MotionStatechartContext):
```
ichumuh, line 375: why is this not called in build?

giskardpy/src/giskardpy/motion_statechart/tasks/cartesian_tasks.py
```diff
@@ -717,14 +736,68 @@ def expand(self, context: MotionStatechartContext) -> None:
 
 
 @dataclass(eq=False, repr=False)
-class CartesianPositionVelocityLimit(Task):
+class CartesianVelocityLimitTask(Task, ABC):
```
ichumuh, line 739: ABC should always come first, otherwise python doesn't properly enforce it

giskardpy/src/giskardpy/motion_statechart/tasks/cartesian_tasks.py
```diff
@@ -717,14 +736,68 @@ def expand(self, context: MotionStatechartContext) -> None:
...
+    Higher weights give this constraint soft priority over lower weighted constraints when conflicts occur."""
+
+    _goal_binding: Optional[ForwardKinematicsBinding] = field(
+        default=None, init=False, repr=False
+    )
+    """Freezes the transform from the root to the goal reference frame, or ``None`` when no goal is set."""
+
+    def _root_frame_goal(
+        self, context: MotionStatechartContext, goal: Optional[SpatialType]
+    ) -> Optional[SpatialType]:
```
ichumuh, line 766: why is goal optional?

giskardpy/src/giskardpy/motion_statechart/tasks/cartesian_tasks.py
```diff
@@ -717,14 +736,68 @@ def expand(self, context: MotionStatechartContext) -> None:
...
+        """
+        Express goal in the root frame, frozen at bind time, or ``None`` when unset.
+
+        The goal's reference frame is captured through a ForwardKinematicsBinding so that a
```
ichumuh, line 770: that is not the purpose of ForwardKinematicsBinding. the purpose is to decide whether the thing should be calculated on build or on start. that makes a difference if you are executing a sequence of goals

giskardpy/src/giskardpy/motion_statechart/tasks/cartesian_tasks.py
```diff
@@ -717,14 +736,68 @@ def expand(self, context: MotionStatechartContext) -> None:
...
+            root=self.root_link,
+            tip=goal.reference_frame,
+            float_variable_data=context.float_variable_data,
+        )
+        self._goal_binding.bind(context.world)
+        return self._goal_binding.root_T_tip @ goal
+
+    def on_start(self, context: MotionStatechartContext):
+        if self._goal_binding is not None:
+            self._goal_binding.bind(context.world)
```
ichumuh, line 788: that looks sus, tell claude to properly figure out how the forward kinematics bindings work

giskardpy/src/giskardpy/motion_statechart/tasks/cartesian_tasks.py
```diff
@@ -717,14 +736,68 @@ def expand(self, context: MotionStatechartContext) -> None:
...
+@dataclass(eq=False, repr=False)
+class CartesianPositionVelocityLimit(CartesianVelocityLimitTask):
     """
     Limit the Cartesian (translational) velocity of a tip link relative to a root link.
 
-    This goal enforces a strict cap on the linear speed of the frame defined by
-    the kinematic transform from `root_link` to `tip_link`. Enforcement is performed
-    by adding constraints to the optimizer and by providing an observation expression
-    that evaluates whether the current translational speed is within the limit.
+    With a `goal_point` the limit is a strict cap on the Euclidean norm of the tip
```
ichumuh, line 796: i'm confused, i thought you split it up into 3 constraints?

giskardpy/src/giskardpy/motion_statechart/tasks/cartesian_tasks.py
```diff
@@ -733,29 +806,19 @@ class CartesianPositionVelocityLimit(Task):
...
-    """
-    tip_link: KinematicStructureEntity = field(kw_only=True)
+    goal_point: Optional[Point3] = field(default=None, kw_only=True)
     """
-    Tip link of the kinematic chain. 
-    The translational velocity of this link (expressed in the root link frame) is constrained.
+    Point the tip is driven toward by its goal task. When given, the limit bounds the tip's
+    actual speed toward this point. When ``None`` the limit falls back to bounding only the
+    radial speed away from the root origin, which is correct solely when the tip moves along
+    the root direction.
```
ichumuh, line 814: hmm this is also sus, i gotta talk to claude about this

ichumuh (changes requested):
My Claude response 😆 : 
 The direction is right — measuring against the root origin was plainly wrong — but the central claim of the PR is not true, and I verified that by running it. The plumbing, the trajectory work, and the tests are good; the premise is off 
  by exactly one directional projection.                                                                                                                                                                                                       
                                                                                                                                                                                                                                               
  I ran the branch in a throwaway worktree (all 7 new tests pass) and then wrote a probe against its own claim. Nothing was posted to the PR.                                                                                                  
                                                                                                                                                                                                                                               
  The core claim doesn't hold                                                                                                                                                                                                                  
                                                                                                                                                                                                                                               
  Docstring: "When the tip travels straight at the goal, that derivative equals the tip's actual speed, so the bound caps the Euclidean norm of the tip velocity."

  d/dt ||p − g|| = −(unit vector toward g) · v — still a single component of the velocity, just re-centered from the root origin to the goal. Everything perpendicular to the goal direction stays unconstrained. The condition in the first   
  half of that sentence is never enforced, and the second half doesn't follow from it.                                                                                                                                                         
                                                                                                                                                                                                                                               
  The failure needs nothing exotic — a robot that can't move in a straight line is enough. One revolute joint, tip 1 m out, goal on the same circle, the limit handed exactly the goal the drive task uses, nothing else acting:               
                                                                                                                                                                                                                                               
  ┌──────────┬───────┬────────────────────┬───────┐                                                                                                                                                                                            
  │ goal arc │ limit │ max achieved speed │ ratio │                                                                                                                                                                                            
  ├──────────┼───────┼────────────────────┼───────┤                                                                                                                                                                                            
  │ 90°      │ 0.05  │ 0.0707             │ 1.41× │                                                                                                                                                                                            
  ├──────────┼───────┼────────────────────┼───────┤                                                                                                                                                                                            
  │ 2.0 rad  │ 0.05  │ 0.0925             │ 1.85× │                                                                                                                                                                                            
  └──────────┴───────┴────────────────────┴───────┘                                                                                                                                                                                            
                                                                                                                                                                                                                                               
  That is limit / cos(θ_remaining/2) to three digits — chord versus tangent. It grows without bound as the remaining angle approaches π, where the closing rate is zero while the tip moves at full joint speed. Every arm swinging its wrist  
  around to a pose hits this.                                                                                                                                                                                                                  
                                                                                                                                                                                                                                               
  What the honest translational fix looks like                                                                                                                                                                                                 
                                                                                                                                                                                                                                               
  ||J q̇||₂ ≤ v is a second-order cone constraint and genuinely not expressible in the QP — that's the real reason for the original hack. But a box on the components is, and it needs no goal at all: three velocity constraints on             root_P_tip[0..2], each bounded by ±v/√3 to strictly guarantee ||v||₂ ≤ v (or ±v for the looser ||v||₂ ≤ √3·v). Structurally that's what add_point_goal_constraints already does. It removes the new goal_point parameter, the binding, and   
  root_P_tip[0..2], each bounded by ±v/√3 to strictly guarantee ||v||₂ ≤ v (or ±v for the                                                                                                                                                      
                                                                                                                                                                                                                                               
  Rotation is genuinely harder — ω is not the time derivative of any scalar function of q, which is why the original used the angle — so a goal-relative angle is a defensible approximation there. It just has to be documented as one.       
                                                                                                                                                                                                                                               
  Nothing in production actually opts in                                                                                                                                                                                                       
                                                                                                                                                                                                                                               
  coraplex/src/coraplex/robot_plans/motions/gripper.py:202-228 builds both limits with no goal, immediately next to the CartesianPosition/CartesianPose it is limiting — self.target is right there at line 240. So every real caller keeps the
  old root-origin behavior, and the PR fixes nothing at any live call site.                                                                                                                                                                    
                                                                                                                                                                                                                                               
  The API invites the bug back                                                                                                                                                                                                                 
                                                                                                                                                                                                                                               
  The goal is copied into a second node with nothing checking the two agree. A stale copy silently degrades to the exact failure mode the PR set out to fix — silently, because the fallback is the None default. Against AGENTS.md's "make    
  interfaces hard to misuse." If the goal-referenced form survives, the limit should take the goal node, or the goal task should own its limit, so there's one source. 

  Observation and constraint still disagree — and now inconsistently                                                                                                                                                                           
                                                                                                                                                                                                                                               
  CartesianPositionVelocityLimit's observation is unchanged (root_P_tip_dot.norm(), the true speed) while its constraint bounds a projection — in my arc run the observation reads False through most of the motion while the constraint is    
  satisfied. Meanwhile CartesianRotationVelocityLimit's observation was changed to track its constraint. Either convention is defensible; the two classes now following different ones is a trap.                                              
                                                                                                                                                                                                                                               
  Numerics at arrival                                                                                                                                                                                                                          
                                                                                                                                                                                                                                               
  Both if_eq_zero guards fire only on exact 0.0. That was tolerable when the reference was the root origin (tip coincident with root — rare). With the goal as reference, the distance goes to zero at convergence on every run, so the        
  constraint's gradient direction becomes arbitrary right at arrival. The rotational one is the sharper case: rotational_error → to_angle → 2·acos(w) carries a 1/sin(θ/2) factor in its derivative — precisely why                            
  add_rotation_goal_constraints uses quaternion components instead of the angle. The two *_holds_through_arrival tests pass, so it survives in practice, but their docstrings claim they exercise the zero guard and they don't — the guard    
  needs exact 0.0.                                                                                                                                                                                                                             
                                                                                                                                                                                                                                               
  The memory-view rebinding is a band-aid on a silent-corruption bug                                                                                                                                                                           
                                                                                                                                                                                                                                               
  FloatVariableData.register_expression does self.data = np.concatenate(...) (float_variable_data.py:64), replacing the array object. Every function previously bound with bind_args_to_memory_view keeps pointing at the old buffer and       
  silently reads stale values — no error, just wrong numbers. The PR's own new ForwardKinematicsBinding is what exposed this.                                                                                                                  
                                                                                                                                                                                                                                               
  Re-binding in on_start works only because all registration currently happens at build time. Anything that registers a float variable later re-breaks it, silently. The fix belongs in FloatVariableData — grow a stable buffer, or track     
  bound consumers and re-bind on growth — and it deserves its own issue regardless of what happens to this PR.                                                                                                                                 
                                                                                                                                                                                                                                               
  Smaller                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                               
  - control_frequency: float = 20.0 in the test class is a retyped copy of QPControllerConfig.create_with_simulation_defaults' target_frequency=20, which _run never sets. Should read executor.qp_controller.config.control_dt. If the default
  moves, every speed is computed against the wrong dt and the assertions go quietly meaningless. (AGENTS.md: assert against the definition, not a copy.)                                                                                       
  - _root_frame_goal returns a value and mutates self._goal_binding; its Optional[SpatialType] return then feeds parameters typed Optional[Point3]/Optional[RotationMatrix], so the typing is lost at the boundary.                            
  - No test covers the case the docstring's strongest claim implies — speed perpendicular to the goal direction. Adding one would have caught this.                                                                                            
                                                                                                                                                                                                                                               
  What's genuinely good                                                                                                                                                                                                                        
                                                                                                                                                                                                                                               
  The trajectory work. Making goal_reference_frame_P_current_target_point a cached_property so a limit can reference it before the trajectory is built, and feeding the look-ahead point in as the limit's goal, does bound the tangential     
  speed correctly — that's the one configuration where the constrained direction really tracks the motion. Worth noting it works for a different reason than the docstring gives: the reference rides ahead with the tip along the path, so the
  chord and the tangent stay aligned. That the constraint treats the target as constant (joint_position_and_velocity_variables skips non-joint variables) is load-bearing here and undocumented.                                               
                                                                                                                                                                                                                                               
  Happy to draft the per-axis translational version and a test for the arc case if you want to take that route.

giskardpy/src/giskardpy/motion_statechart/tasks/cartesian_tasks.py
```diff
@@ -856,14 +919,20 @@ class CartesianVelocityLimit(Parallel):
...
     """Maximum allowed linear speed of the tip in meters per second (m/s).
-    Default: 0.1 m/s. The enforcement ensures the Euclidean norm of the
-    tip-frame translational velocity does not exceed this value."""
+    Default: 0.1 m/s. With ``goal_point`` set the enforcement ensures the Euclidean norm of
+    the tip velocity does not exceed this value; without it only the radial speed is bounded."""
     max_angular_velocity: float = field(default=0.4, kw_only=True)
     """Maximum allowed angular speed. Interpreted in radians per second (rad/s).
-    Default: 0.5 rad/s. The enforcement ensures the magnitude of the instantaneous
-    rotation rate does not exceed this threshold."""
+    Default: 0.4 rad/s. With ``goal_orientation`` set the enforcement ensures the tip's
```
AbdelrhmanBassiouny, line 934: no need to mention the number in the doc string, it is already apparent in the field definition

giskardpy/src/giskardpy/motion_statechart/tasks/cartesian_tasks.py
```diff
@@ -856,14 +919,20 @@ class CartesianVelocityLimit(Parallel):
...
+    """Point the tip is driven toward. Forwarded to the linear limit so it bounds the tip's
+    actual speed. When ``None`` the linear limit falls back to bounding only the radial speed."""
+    goal_orientation: Optional[RotationMatrix] = field(default=None, kw_only=True)
+    """Orientation the tip is driven toward. Forwarded to the angular limit so it bounds the
+    tip's actual angular speed. When ``None`` the angular limit falls back to the current-to-root angle rate."""
     max_linear_velocity: float = field(default=0.1, kw_only=True)
     """Maximum allowed linear speed of the tip in meters per second (m/s).
-    Default: 0.1 m/s. The enforcement ensures the Euclidean norm of the
-    tip-frame translational velocity does not exceed this value."""
+    Default: 0.1 m/s. With ``goal_point`` set the enforcement ensures the Euclidean norm of
```
AbdelrhmanBassiouny, line 930: no need to mention the number in the doc string, it is already apparent in the field definition

giskardpy/src/giskardpy/motion_statechart/constraint_builders.py
```diff
@@ -198,19 +210,31 @@ def add_rotational_velocity_limit(
...
+            root orientation, which only bounds the current-to-root angle rate.
         :param max_velocity: rad/s
         """
-        root_Q_tipCurrent = frame_R_current.to_quaternion()
-        angle_error = root_Q_tipCurrent.to_axis_angle()[1]
+        if frame_R_goal is None:
+            angle_error = frame_R_current.to_quaternion().to_axis_angle()[1]
+        else:
+            angle_error = frame_R_current.rotational_error(frame_R_goal)
+        angle_error = sm.if_eq_zero(angle_error, sm.Scalar(0.01), angle_error)
```
AbdelrhmanBassiouny, line 237: hard coded number emoji, define this number at a place where it is easier to see if we wanted to modify it, like a default argument to this function for example or a class variable or if multiple classes then a moudle level variable

AbdelrhmanBassiouny (changes requested):

PR #548, opened 2026-08-12
Make motion statechart nodes JSON serializable

### LINKED ISSUE #510: Make MotionStateChartNodes a SubclassJSONSerializer
Currently it uses the generic dataclass json serializer, meaning bodies get fully serialized. Make it use the id for bodies instead

MarvinDufner:
Fixes #510 

Makes MotionStateChartNodes a SubclassJSONSerializer so that it doesnt use the generic json serializer anymore. 
It now correctly serializes world entities (when bodies are used as root or tip link for example) by id, making the json sizes of motion state charts much smaller.

Mirrors `WorldEntityWithID.to_json`


Tigul (approved):

giskardpy/src/giskardpy/motion_statechart/graph_node.py
```diff
@@ -607,6 +609,50 @@ def _create_state_variables(self):
...
 
+    def to_json(self) -> Dict[str, Any]:
+        """
+        Serialize this node, writing referenced world entities as their id rather than
+        by value.
+        """
+        result = super().to_json()
+        for attribute in DataclassOnlyIntrospector().discover(self.__class__):
+            value = getattr(self, attribute.public_name)
+            if isinstance(value, (list, set)):
```
ichumuh, line 620: hm feels like we have an open closed situation here. this will probably break on a parameter that is a dict. isn't there a recursive way to doing it?

giskardpy/src/giskardpy/motion_statechart/graph_node.py
```diff
@@ -607,6 +609,50 @@ def _create_state_variables(self):
...
+        Rebuild a node, resolving serialized ids back to the world's own entities.
+        """
+        values = {}
+        for attribute in DataclassOnlyIntrospector().discover(cls):
+            if attribute.public_name not in data:
+                continue
+            value = data[attribute.public_name]
+            if isinstance(value, list):
+                items = [
+                    WorldEntityWithID._item_from_json(item, **kwargs) for item in value
```
ichumuh, line 640: same as above, i don't like the spatial case for list

ichumuh (changes requested):

giskardpy/src/giskardpy/motion_statechart/graph_node.py
MarvinDufner: true, that probabably needs to be changed in world entity too then

ichumuh: why? i don't see this kind of if else in WorldEntity

MarvinDufner: Im pretty sure WorldEntitryWithID also cant handle dicts:
[in from_json](https://github.com/AICOR-Solutions/cognitive_robot_abstract_machine/blob/11bc063e44710411a940cf1355e3d8909397eae5/semantic_digital_twin/src/semantic_digital_twin/world_description/world_entity.py#L223-L237)
[in to_json](https://github.com/AICOR-Solutions/cognitive_robot_abstract_machine/blob/11bc063e44710411a940cf1355e3d8909397eae5/semantic_digital_twin/src/semantic_digital_twin/world_description/world_entity.py#L160-L161)

ichumuh: oh, i guess i looked in the wrong place. ok i think this is out of scope for this PR then. i'll approve when the tests are fixed

semantic_digital_twin/src/semantic_digital_twin/adapters/ros/ros_msg_serializer.py
```diff
@@ -36,6 +39,13 @@ def to_json(cls, obj: Any) -> Dict[str, Any]:
 
     @classmethod
     def from_json(cls, data: Dict[str, Any], clazz: Type, **kwargs) -> Any:
+        # convert_dictionary_to_ros_message does not accept the tracker that from_json threads
+        # through for resolving world entity ids, since ROS messages never reference them
+        kwargs = {
+            key: value
+            for key, value in kwargs.items()
+            if not isinstance(value, WorldEntityWithIDKwargsTracker)
```
LucaKro, line 47: but why do we need to filter them out then? shouldnt it just be ignored if we pass a tracker we dont need?

giskardpy/src/giskardpy/motion_statechart/graph_node.py
LucaKro: hmm yea there we needed for semantic annotations, since we have so many of them, and there are so many references to world entities in them.

but i acknowledge that doing it this way is not in the spirit of SubclassJsonSerializer.

doing it recursively would be essentially be doing it like SubclassJsonSerializer usually does it no?

we need to judge if with motion state chart nodes we have a similar situation to semantic annotations or if it should be done the intended way

LucaKro: okay i talked to simon and he says its the same situation as with semantic annotations, where we want only want to reference the ID when converting to json

LucaKro (changes requested):

giskardpy/src/giskardpy/motion_statechart/graph_node.py
Tigul: I think I implemented serialization of dicts in the SubClassJSONSerializr a while back for the JointStates

LucaKro: hmm really? can you point me to where that was?

tomsch420: why do you even want JSON serialization for this?

ichumuh: > why do you even want JSON serialization for this?

for what? motion statechart nodes?

Tigul: https://github.com/cram2/cognitive_robot_abstract_machine/blob/main/krrood/src/krrood/adapters/json_serializer.py#L231

tomsch420: yes

ichumuh: > yes

motion statecharts need to be sent over ros, so the need to be serializable

ichumuh: > https://github.com/cram2/cognitive_robot_abstract_machine/blob/main/krrood/src/krrood/adapters/json_serializer.py#L231

i don't think that triggers in the lines that @MarvinDufner posted

semantic_digital_twin/src/semantic_digital_twin/adapters/ros/ros_msg_serializer.py
MarvinDufner: what do you mean ignore the tracker? if we pass kwargs with the tracker into convert_dictionary_to_ros_message then the whole thing crashes with
`TypeError: convert_dictionary_to_ros_message() got an unexpected keyword argument '<name-mangled tracker key>'`

giskardpy/src/giskardpy/motion_statechart/graph_node.py
MarvinDufner: so whats the result of this discussion now? :sweat_smile: 
Anything that needs to be added in this pr or is everything else out of scope?

ichumuh:
fixed in https://github.com/cram2/cognitive_robot_abstract_machine/pull/618

PR #550, opened 2026-08-14
Give doors a hinge when their URDF wires them straight to their cabinet

tomsch420:
The world reasoner classified a Door wherever a body sits between a revolute connection (from its cabinet) and a fixed connection (to its handle), which is exactly how bodies like iai_fridge_door come out of URDF: cabinet -(revolute)-> door, with no intervening hinge body. Such doors never got a Hinge, leaving mechanical_joint unset even though the annotation calls for one.

WorldReasoner now calls a new Door.ensure_hinge() after registering each inferred door: if its root is still directly revolute-connected to its parent, it spawns a collision-less Hinge at the door's own pose, copying the axis/multiplier/offset/limits from that connection, and mounts it with door.add(hinge).

Mounting exposed a gap in MechanicalJoint._mount_strategy: it assumed the whole already had a plain fixed connection to its parent and the joint supplied the active one. For a door already wired active-to-parent, that produced two stacked revolute connections instead of the intended cabinet -> revolute -> hinge -> fixed -> door. The strategy now detects when the whole's existing connection has the same type the joint provides, discards it (and the degree of freedom it used, so it doesn't end up orphaned) in favor of a fixed connection between joint and whole.

Made with the help of Claude.


semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/semantic_annotations.py
```diff
@@ -281,14 +291,40 @@ def _mount_strategy(
...
+            main_has_root_body_annotation._world.move_branch_with_fixed_connection(
+                main_has_root_body_annotation.root, self.root
+            )
+            self._discard_orphaned_dofs(whole_connection)
+        else:
+            main_has_root_body_annotation._world.move_branch(
+                main_has_root_body_annotation.root, self.root
+            )
+
+    def _discard_orphaned_dofs(self, discarded_connection: Connection) -> None:
```
LucaKro: we dont need this, i think coming out of a modify world block should already do this

LucaKro (changes requested):
do we try to merge before ijcai or do we do local merge?

overall this looks fine, thank you

LucaKro (approved):
the test failing is not flaky

LucaKro (approved):

semantic_digital_twin/src/semantic_digital_twin/reasoning/world_reasoner.py
```diff
@@ -81,6 +87,10 @@ def _update_world_attributes(self):
             else:
                 for semantic_annotation in attr_value:
                     self.world.add_semantic_annotation_recursively(semantic_annotation)
+                    if isinstance(semantic_annotation, Door):
```
AbdelrhmanBassiouny: that violates open-closed, can't the semantic annotation know what default mechanical joint it needs? also there could be a common class for classes that have a default mechanical joint.

test/semantic_digital_twin_test/test_semantic_annotations/test_factories.py
```diff
@@ -1450,6 +1452,183 @@ def test_mechanical_joint_mount_splices_under_whole_parent():
...
+    ``create_default_mechanical_joint`` does nothing for a door mounted with a fixed
+    connection (the normal case built through the factories), leaving it without a
+    mechanical joint.
+    """
+    world = _world_with_root()
+    with world.modify_world():
+        door = Door.create_with_new_body_in_world(
+            name="door", scale=Scale(0.03, 1, 2), world=world
+        )
+        door.create_default_mechanical_joint(Hinge)
```
AbdelrhmanBassiouny: this is unintuitive, creating a mechanical joint Hinge leaves the mechanical joint None?

AbdelrhmanBassiouny (changes requested):

LucaKro (approved):

AbdelrhmanBassiouny (approved):

PR #553, opened 2026-08-19
[mujoco] Fixed free-jointed object spawn 

sorinar329:
- Fix free-jointed bodies spawning at the origin in MuJoCo keyframes instead of their authored pose.
- Add regression test with four free-jointed cubes to catch it.

### DIFF
--- a/semantic_digital_twin/src/semantic_digital_twin/adapters/multi_sim.py
+++ b/semantic_digital_twin/src/semantic_digital_twin/adapters/multi_sim.py
@@ -1802,18 +1802,30 @@ class MujocoBuilder(MultiSimBuilder):
         key_element.set("name", "home")
         key_element.set("time", "0")
         qpos = []
-        for body in self.world.bodies:
+        for body in self.world.bodies_topologically_sorted:
             parent_connection = body.parent_connection
             if (
                 isinstance(parent_connection, self._ignore_connection_types)
                 or parent_connection is None
             ):
                 continue
-            qpos += [
-                self.world.state[dof.id].position
-                for dof in parent_connection.active_dofs
-                + parent_connection.passive_dofs
-            ]
+            if isinstance(parent_connection, Connection6DoF):
+                # A free joint's DOF state is relative to the connection frame
+                # and stays at the identity by default; the body's actual
+                # placement lives in parent_T_connection_expression instead.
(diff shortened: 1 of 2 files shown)


tomsch420 (approved):

semantic_digital_twin/src/semantic_digital_twin/adapters/multi_sim.py
```diff
@@ -1802,18 +1802,30 @@ def _end_build(self, file_path: str):
...
                 isinstance(parent_connection, self._ignore_connection_types)
                 or parent_connection is None
             ):
                 continue
-            qpos += [
-                self.world.state[dof.id].position
-                for dof in parent_connection.active_dofs
-                + parent_connection.passive_dofs
-            ]
+            if isinstance(parent_connection, Connection6DoF):
```
LucaKro, line 1812: hmm this seems very sus. please just use the intended method for this (set Connection6DOF.origin) and everything should be calculated correctly in there. setting the world state like this is not the intended access point

LucaKro (changes requested):

semantic_digital_twin/src/semantic_digital_twin/adapters/multi_sim.py
sorinar329: i dont really understand your comment, here we basically read the position+orientation and build the qpos for mujoco.

ichumuh: exactly. this looks right to me. its kind of inefficient to compile and evaluate the function every time though.

ichumuh: my claude did a similar fix btw, but also inefficient, we can worry about that when runtime is an issue

ichumuh (approved):

semantic_digital_twin/src/semantic_digital_twin/adapters/multi_sim.py
LucaKro: oh actually my bad, i read it as "you are updating the world state", but you are actually reading. nvm

LucaKro (approved):

LucaKro:
lets make sure the flaky test actually passes once, since this is affecting multisim

PR #554, opened 2026-08-20
Franka panda integration

sorinar329:
## Discussion Points

1. Where to place the robot description and assets? Currently under `coraplex/resources/robots`.
2. The meshes are currently downloaded upon the first execution of the demo — is this something we want to change?
3. Where should the demo itself be placed? Currently in `coraplex/demos`; an alternative might be the physical simulation package.
4. There is actually not test yet for it. For testing it would require to use some placeholder meshes, and i'm not sure if that should be the right way.  One solution would be to look for a official franka ROS package

## Changes

1. Added the Panda description along with one small demo file that spawns the Panda and executes `ParkArmsAction()` in MuJoCo. This demo will be expanded further in future PRs.
2. Added a real-time factor so execution can be limited to real time if required. This is needed because the simulation would otherwise run as fast as the hardware allows, which can produce different outcomes on different hardware.
3. `prediction_horizon` is now an attribute with a default value of 4, so other demos/tests still work as before. This is needed because the Panda requires a longer horizon (set to 20 in the demo).
5. Changed the jerk weight from `None` to `0.01`, since leaving it as `None` caused the Panda to crash.
6. Keeping `ctrl` in sync with `qpos`. Previously the synchronizer only pushed the kinematic state into `qpos` without updating `ctrl`, so the actuator tried to pull the joint back to its previous `ctrl` value.


coraplex/demos/coraplex_panda_demo/panda_mesh_assets.py
```diff
@@ -0,0 +1,110 @@
...
+    """
+
+    revision: str = "main"
+    """
+    Git revision of ``mujoco_menagerie`` to download from.
+
+    Pin this to a commit to keep a scene reproducible against upstream changes.
+    """
+
+    timeout: float = 60.0
```
AbdelrhmanBassiouny: timedelta

semantic_digital_twin/src/semantic_digital_twin/adapters/multi_sim.py
```diff
@@ -2859,6 +2859,19 @@ def _resolve_qpos_adr(self, connection: Connection) -> Optional[int]:
             return None
         return mj_model.jnt_qposadr[joint_id]
 
+    def _resolve_ctrl_adr(self, connection: Connection) -> Optional[int]:
```
AbdelrhmanBassiouny: no abbreviations

semantic_digital_twin/src/semantic_digital_twin/adapters/multi_sim.py
```diff
@@ -3018,18 +3031,27 @@ def _write_1dof_to_qpos(
         self,
         connection: ActiveConnection1DOF,
         qpos_adr: int,
+        ctrl_adr: Optional[int],
```
AbdelrhmanBassiouny: no abbreviations

semantic_digital_twin/src/semantic_digital_twin/adapters/multi_sim.py
```diff
@@ -3018,18 +3031,27 @@ def _write_1dof_to_qpos(
         self,
         connection: ActiveConnection1DOF,
         qpos_adr: int,
```
AbdelrhmanBassiouny: no abbreviations

semantic_digital_twin/src/semantic_digital_twin/robots/panda.py
```diff
@@ -0,0 +1,220 @@
...
+            root=robot_root._world.get_body_in_branch_by_name(robot_root, "link0"),
+            tip=robot_root._world.get_body_in_branch_by_name(robot_root, "link7"),
+        )
+
+
+@dataclass(eq=False)
+class Panda(AbstractRobot, HasOneArm[PandaArm]):
+    """The Franka Emika Panda arm, as used in the stacking demo.
+
+    https://franka.de/
```
AbdelrhmanBassiouny: is this the proper way to add a link in the rst/sphinx docs?

AbdelrhmanBassiouny (changes requested):

sorinar329:
This has to be merged first 
https://github.com/code-iai/iai_robots/pull/51

coraplex/demos/coraplex_panda_demo/panda_mesh_assets.py
sorinar329: removed

semantic_digital_twin/src/semantic_digital_twin/robots/panda.py
sorinar329: changed to be inline hyperlink.

coraplex/demos/coraplex_panda_demo/demo.py
```diff
@@ -0,0 +1,223 @@
...
+    """
+
+    ros_node_name: ClassVar[str] = "panda_demo_node"
+
+    multi_sim: MujocoSim | None = field(init=False, default=None)
+    """
+    The MuJoCo simulation backing this run, started once the scene is fully populated.
+    """
+
+    def build_simulated_world(self) -> World:
```
LucaKro, line 60: please check out experiments.real_stretch_apartment_demo.demo.StretchApartmentDemonstration.build_simulated_world

1. dont hardcode panda here, pass it as used_robot to RobotDemonstration. that way its easily swapable with other robots if we want to (we claim that with cram we write robot agnostic plans where we can swap out robots)
2. please use WorldSpecification to spawn the world together with the robot

coraplex/demos/coraplex_panda_demo/demo.py
```diff
@@ -0,0 +1,223 @@
...
+
+    def populate_scene(self, world: World) -> None:
+        """
+        Add a ground plane and two cubes, then start the MuJoCo simulation.
+
+        The simulation is built from the world's state at construction time, so it can
+        only start once every body the scene needs is already in the world.
+        """
+        with world.modify_world():
+            ground_plane = Body(name=PrefixedName("ground_plane"))
```
LucaKro: sounds like a floor. please use The floor semantic annotation and the floor default root kinematic structure entitiy specification, it does everything you do by hand here.

check out experiments.real_stretch_apartment_demo.demo.StretchApartmentDemonstration.populate_scene

coraplex/demos/coraplex_panda_demo/demo.py
```diff
@@ -0,0 +1,223 @@
...
+
+        panda = world.get_semantic_annotations_by_type(self.used_robot)[0]
+        arm = panda.get_arms()[0]
+        for connection in arm.active_connections:
+            connection.child.simulator_additional_properties.append(
+                MujocoBody(gravitation_compensation_factor=1.0)
+            )
+
+        headless = os.environ.get("CI", "false").lower() == "true"
+        self.multi_sim = MujocoSim(world=world, headless=headless, step_size=1e-3)
```
LucaKro: hmm maybe we could add this to the general RobotDemonstration class? this would make it a lot easier for people who use mujoco rarely (like myself) to try out the demonstrations in mujoco. for example would it also be an opportunity to try out the stretch demo in mujoco, if you add an easy way to do that. 

this is exactly why i created this robotdemonstration interface 😎

coraplex/demos/coraplex_panda_demo/demo.py
```diff
@@ -0,0 +1,223 @@
...
+
+    def build_plan(self, context: Context) -> PlanNode:
+        return sequential([ParkArmsAction(Arms.BOTH)], context=context)
+
+    def tear_down(self) -> None:
+        """
+        Let the simulation settle for a moment before stopping it, so a non-headless
+        viewer sees the final pose rather than the window closing mid-motion.
+        """
+        if self.multi_sim is not None:
```
LucaKro: this can then just go into the Superclass for example

semantic_digital_twin/src/semantic_digital_twin/adapters/multi_sim.py
```diff
@@ -2859,6 +2859,19 @@ def _resolve_qpos_adr(self, connection: Connection) -> Optional[int]:
             return None
         return mj_model.jnt_qposadr[joint_id]
 
+    def _resolve_ctrl_address(self, connection: Connection) -> Optional[int]:
```
LucaKro: no abbreviations. also param end return doc

semantic_digital_twin/src/semantic_digital_twin/adapters/multi_sim.py
```diff
@@ -3017,19 +3030,28 @@ def _write_6dof_to_qpos(
     def _write_1dof_to_qpos(
         self,
         connection: ActiveConnection1DOF,
-        qpos_adr: int,
+        qpos_address: int,
```
LucaKro, line 3037: same here

semantic_digital_twin/src/semantic_digital_twin/robots/panda.py
LucaKro, line 1: what was your experience like writing this (or if claude did this, did it oneshot it? :D)

semantic_digital_twin/src/semantic_digital_twin/robots/robot_parts.py
```diff
@@ -800,6 +802,24 @@ def _setup_velocity_limits(self):
...
         self.tighten_dof_velocity_limits_proportionally(maximum_velocity=1)
 
+    def _setup_acceleration_limits(self):
+        """
+        Sets up acceleration limits for 1-DOF connections in the robot.
+
+        No-op by default; robots with known acceleration limits (e.g. from the
+        manufacturer's datasheet) should override this.
+        """
+        pass
```
LucaKro, line 812: are there sensible defaults you could provide here and in the jerk setup?
maybe @ichumuh has an idea?

Tigul:
Regarding the discussion points: 
1. We wanted to keep robot assets out of the Monorepo, so just put the ros package path in the Panda description and use that
2. Similar to 1, do we not have them in iai_robots? If not might add them 
3. Can also be moved to expreriments and use the RobotDemonstration base class 
4. There is no oficial Franke ROS package???

semantic_digital_twin/src/semantic_digital_twin/robots/robot_parts.py
```diff
@@ -872,6 +892,50 @@ def tighten_dof_velocity_limits_of_1dof_connections(
                 ),
             )
 
+    def tighten_dof_acceleration_limits_of_1dof_connections(
```
LucaKro, line 895: can you call this and tighten_dof_velocity_limits_of_1dof_connections sequentially without them overriding each other?

LucaKro (changes requested):
Good job! Please also write tests for the other methods you added

coraplex/src/coraplex/plans/executables.py
```diff
@@ -118,6 +118,27 @@ class GiskardExecutable(Executable):
     :py:class:`pycram.motion_executor.ExecutionEnvironment`.
     """
 
+    real_time_factor: ClassVar[Optional[float]] = None
```
Tigul: Having those here as class variables is not very accessible, might want to put them in the context

coraplex/src/coraplex/execution_environment.py
```diff
@@ -45,34 +61,63 @@ class ExecutionEnvironment:
...
+        prediction_horizon: Optional[int] = None,
+    ):
         """
         Configure the environment for use as a context manager, allowing ``with
-        simulated_robot(collision_avoidance=True):``.
+        simulated_robot(collision_avoidance=True, real_time_factor=1.0,
+        prediction_horizon=20):``.
         """
         self.collision_avoidance = collision_avoidance
+        self.real_time_factor = real_time_factor
```
Tigul: You don't need these if they are in the context

Tigul (changes requested):

semantic_digital_twin/src/semantic_digital_twin/robots/panda.py
sorinar329: It was long ago, but i think it was pretty much one-shot.

semantic_digital_twin/src/semantic_digital_twin/adapters/multi_sim.py
sorinar329: are you sure we want to write out qpos? its also used 66 times in this file

semantic_digital_twin/src/semantic_digital_twin/robots/robot_parts.py
sorinar329: claude tells me that the jerk limits are computed from the vel limits and QP prediction horizon

semantic_digital_twin/src/semantic_digital_twin/robots/robot_parts.py
sorinar329: yes, i think they should only touch their respective slot in DerivativeMap

coraplex/src/coraplex/plans/executables.py
sorinar329: adressed

coraplex/src/coraplex/execution_environment.py
sorinar329: addressed

sorinar329:
Addressed all comments, but i would actually not want to merge this PR yet. Currently we are working on transferring the demos we made to the real robot, and since we want to close the gap to the real world as much as possible in simulation through mujoco, the whole PR might change.

semantic_digital_twin/src/semantic_digital_twin/adapters/multi_sim.py
LucaKro: yes i am, but also nvm this sounds out of scope for your PR, so dw about it. but as a general rule, yes. No abbreviations :gun:

semantic_digital_twin/src/semantic_digital_twin/robots/panda.py
LucaKro: cool cool

semantic_digital_twin/src/semantic_digital_twin/robots/robot_parts.py
LucaKro: can you write a test that confirms that?

LucaKro:
> Addressed all comments, but i would actually not want to merge this PR yet. Currently we are working on transferring the demos we made to the real robot, and since we want to close the gap to the real world as much as possible in simulation through mujoco, the whole PR might change.

alright, then ill wait with the next review until you give the go

semantic_digital_twin/src/semantic_digital_twin/robots/robot_parts.py
ichumuh: giskard doesn't support acceleration limits, jerk limits only kinda, but it's not intended to be set by the user

PR #555, opened 2026-08-20
[Ormatic] Fix branch switches with ormatic interfaces + a progress bar

AbdelrhmanBassiouny:
See https://github.com/AbdelrhmanBassiouny/cognitive_robot_abstract_machine/pull/183


scripts/ensure_orm_interfaces.py
LucaKro, line 1: i dont like it, just because it exists doesnt mean its correct. i think we should stick to one generation script, and i think regenerate_all_orm.py covers all needed cases

LucaKro (changes requested):

scripts/ensure_orm_interfaces.py
AbdelrhmanBassiouny: me neither, but left it for others to comment on it as I am not sure haha.

LucaKro: remove it :gun:

AbdelrhmanBassiouny: removed

LucaKro (approved):

tomsch420 (approved):

PR #556, opened 2026-08-20
Ijcai planning tutorials

LucaKro:
Adds the tutorial i wrote for IJCAI 2026 to the main repository, as well as fixes i had to implement to get everything to work properly, specifically targetting the giskard location and collision avoidance


coraplex/demos/coraplex_kitchen_fridge_demo/demo.py
```diff
@@ -0,0 +1,515 @@
...
+from semantic_digital_twin.world_description.world_entity import (
+    Body,
+    KinematicStructureEntity,
+)
+
+# %% what holds what
+
+
+@symbolic_function
+def contains(
```
Tigul, line 80: Do we not have this as a predicate?

coraplex/demos/coraplex_kitchen_fridge_demo/demo.py
```diff
@@ -0,0 +1,515 @@
...
+            the body the kitchen names it after.
+        """
+        counter_top = variable(CounterTop, domain=world.semantic_annotations)
+        return the(
+            entity(counter_top).where(
+                counter_top.root.name.name == "kitchen_island_surface"
+            )
+        ).first()
+
+    def place_spot_on_island(self, world: World, milk: Milk) -> Point3:
```
Tigul, line 240: Do you really need all these small methods? I think that bloats the code and makes it harder to read

coraplex/demos/coraplex_kitchen_fridge_demo/demo.py
```diff
@@ -0,0 +1,515 @@
...
+# %% the demonstration
+
+
+@dataclass
+class KitchenFridgeDemonstration(RobotDemonstration):
+    """
+    A robot fetches a milk out of the kitchen fridge and puts it on the kitchen island.
+    """
+
+    ros_node_name: ClassVar[str] = "kitchen_fridge_demo_node"
```
tomsch420: doc and classvar grrrrr

coraplex/demos/coraplex_kitchen_fridge_demo/demo.py
```diff
@@ -0,0 +1,515 @@
...
+        :return: The steps with the container steps around them, and ``actions`` itself
+            when the body stands in the open.
+        """
+        handle = self.handle_of_enclosing_container(body, context.world)
+        if handle is None:
+            return actions
+
+        door_arm = Arms.LEFT
+
+        return [
```
tomsch420, line 350: very gucci

coraplex/src/coraplex/locations/backends.py
```diff
@@ -15,45 +20,58 @@
...
-from semantic_digital_twin.spatial_types.spatial_types import Pose
+from semantic_digital_twin.spatial_types.spatial_types import Point3, Pose
 from semantic_digital_twin.world import World
 from semantic_digital_twin.world_description.world_entity import Body
 
+logger = logging.getLogger("coraplex")
+
 
 @dataclass
 class GiskardLocationBackend(PoseGeneratorBackend):
```
tomsch420, line 41: this is out of scope but i truly believe this should be an eql backend. That would make things way easier

coraplex/src/coraplex/locations/base.py
```diff
@@ -185,6 +205,20 @@ class PoseGeneratorBackend:
     def __iter__(self) -> Iterator[Pose]:
         pass
 
+    def copy_for_world(self, world: World) -> PoseGeneratorBackend:
```
tomsch420, line 208: this doesnt return a copy so the name is confusing to me

coraplex/src/coraplex/locations/costmaps.py
```diff
@@ -30,6 +30,12 @@
 
 logger = logging.getLogger("coraplex")
 
+BASE_CLEARANCE = 0.1
```
tomsch420: no module variables

coraplex/src/coraplex/locations/costmaps.py
```diff
@@ -375,19 +385,30 @@ def __iter__(self) -> Iterator[Pose]:
         ):
             self.number_of_samples = self.map.flatten().shape[0]
 
-        segmented_maps = self.segment_map()
-        samples_per_map = self.number_of_samples // len(segmented_maps)
-        for seg_map in segmented_maps:
+        for seg_map in self.segment_map():
```
tomsch420: no abbrivations

coraplex/src/coraplex/locations/costmaps.py
```diff
@@ -375,19 +385,30 @@ def __iter__(self) -> Iterator[Pose]:
...
+            weights = seg_map.flatten()
+            sample_count = min(self.number_of_samples, np.count_nonzero(weights))
+            if sample_count == 0:
+                continue
 
             if self.sample_randomly:
-                indices = np.random.choice(seg_map.size, samples_per_map, replace=False)
+                # A costmap is a distribution, so a random draw follows its values: an
+                # even draw over the whole segment would spend most samples on the cells
```
tomsch420, line 376: this all could be so much cooler with eql 😭

coraplex/src/coraplex/plans/executables.py
```diff
@@ -321,7 +274,7 @@ def _execute_simulation(self) -> None:
         executor.compile(motion_state_chart)
 
         counter = 0
-        while counter < len(self.motion_mappings) * 2000:
+        while counter < len(self.motion_nodes) * 2000:
```
tomsch420: uhhh magic number?

coraplex/src/coraplex/plans/failures.py
```diff
@@ -16,6 +18,16 @@
     from semantic_digital_twin.datastructures.definitions import StaticJointState
 
 
+MOTION_DID_NOT_WORK_OUT = (TimeoutError, InfeasibleException, CollisionViolatedError)
```
tomsch420: superclass time?

coraplex/src/coraplex/robot_plans/actions/core/container.py
```diff
@@ -52,6 +55,18 @@ class OpenAction(ActionDescription):
     Arm that should be used for opening the container.
     """
 
+    approach_direction: ApproachDirection = ApproachDirection.FRONT
```
tomsch420, line 58: these are bundles of fields i also have to regularily introduce because coraplex association is not really working well. Perhaps we need to figure out some better architecture for stuff like this

coraplex/src/coraplex/robot_plans/actions/core/container.py
```diff
@@ -149,42 +164,89 @@ class CloseAction(ActionDescription):
...
 
+    @staticmethod
+    def pre_condition(
+        variables: Dict[str, Variable], context: Context, kwargs: Dict[str, Any]
+    ) -> ConditionType:
+        """
+        The gripper with which to close the container has to be free and the handle has
+        to be reachable.
+        """
+        end_effector = ViewManager.get_end_effector_view(
```
tomsch420, line 221: frick the arms enum

coraplex/src/coraplex/robot_plans/actions/base.py
```diff
@@ -50,18 +50,11 @@ def world(self) -> Optional[World]:
...
 
-        if self.plan.context.evaluate_conditions:
-            self.evaluate_pre_condition()
-
-        result = None
-
-        result = self.execute()
-
-        return result
+        return self.execute()
```
tomsch420, line 57: fixing buggy conditions gaming :D

coraplex/src/coraplex/robot_plans/motions/base.py
```diff
@@ -34,6 +35,16 @@ class BaseMotion(Designator):
     motions or actions. Motions create exactly one goal.
     """
 
+    holds_its_goal_until_the_motion_ends: ClassVar[bool] = False
```
tomsch420: if its a classvar its for all motions which is weird

tomsch420 (changes requested):

coraplex/demos/coraplex_kitchen_fridge_demo/demo.py
LucaKro: both come from the superclass

coraplex/src/coraplex/locations/base.py
LucaKro: the subclasses override it with an actual copy. i guess we could copy this here as well, but its not actually necessary here, so it would only be unnecessary computation. 
i think it can stay like this, but if you think i should still copy it ill do it

coraplex/src/coraplex/locations/costmaps.py
LucaKro: maybe at some point :pray: but out of scope for here

coraplex/src/coraplex/plans/executables.py
LucaKro: yes, but out of scope for this PR imo. although if you want i could expose it i guess

coraplex/src/coraplex/robot_plans/actions/core/container.py
LucaKro: i agree, maybe we can talk about this in next cramgineers meeting or sth 
but i think its also out of scope for this PR

coraplex/src/coraplex/robot_plans/actions/base.py
LucaKro: yea it was dead code, so jonas agreed that i can remove it

coraplex/src/coraplex/locations/backends.py
LucaKro: i agree!

coraplex/src/coraplex/locations/costmaps.py
LucaKro: no abbreviations go brr in this file

coraplex/doc/source/_config.yml
```diff
@@ -14,6 +14,14 @@ logo                        : "_static/images/cora-plex-dark.png"  # A path to t
 # See https://jupyterbook.org/content/execute.html
 execute:
   execute_notebooks: auto
+  # The exercises drive a simulated robot: they need ROS and the robot descriptions, which
```
Tigul, line 17: Shouldn't the runner have the ROS environment since it is the defualt docker, otherwise this should be easily changable in the workflow config

coraplex/scripts/convert_exercises_for_self_assessment.sh
Tigul, line 1: Why is this its own script and not integrated in the other convert script?

coraplex/scripts/test_exercises.sh
Tigul, line 1: Same question, why the new sctipt?

coraplex/self_assessment/exercises/0_writing_a_robot_plan.md
```diff
@@ -0,0 +1,1119 @@
...
+    CounterTop,
+    Fridge,
+    Milk,
+    ShelfLayer,
+)
+from semantic_digital_twin.spatial_types import HomogeneousTransformationMatrix
+from semantic_digital_twin.spatial_types.spatial_types import Point3, Pose
+from semantic_digital_twin.world import World
+from semantic_digital_twin.world_description.geometry import Color, Mesh, Scale
+from semantic_digital_twin.world_description.world_entity import (
```
Tigul, line 108: HAben wir noch imports? 

Image

coraplex/self_assessment/exercises/0_writing_a_robot_plan.md
```diff
@@ -0,0 +1,1119 @@
...
+    ],
+    context,
+)
+
+with simulated_robot(collision_avoidance=True):
+    plan.perform()
+```
+
+```{code-cell} ipython3
+:tags: [verify-solution, remove-input]
```
Tigul, line 582: What is with the formatting here

coraplex/self_assessment/exercises/0_writing_a_robot_plan.md
```diff
@@ -0,0 +1,1119 @@
...
+            _debug=True,
+        )
+
+    # TODO: build the whole plan
+    def build_plan(self, context: Context) -> PlanNode:
+        return ...
+
+
+demonstration = KitchenFridgeDemonstration(used_robot=PR2)
+# demonstration.run()
```
Tigul, line 999: commented code

coraplex/src/coraplex/locations/backends.py
```diff
@@ -63,28 +79,76 @@ class GiskardLocationBackend(PoseGeneratorBackend):
...
+
+    distance_to_obstacle: Optional[float] = None
+    """
+    Distance by which the obstacles should be inflated, in meters.
+
+    Defaults to the clearance the robot's base needs, see
+    :meth:`~coraplex.locations.costmaps.OccupancyCostmap.default_distance_to_obstacle`.
+    """
+
+    candidate_seeds: int = 50
```
Tigul, line 93: Isn't 50 a bit high? that would take very long to terminate

coraplex/src/coraplex/locations/backends.py
```diff
@@ -63,28 +79,76 @@ class GiskardLocationBackend(PoseGeneratorBackend):
...
 
     def __post_init__(self):
-        base_bb = self.robot.mobile_base.bounding_box
-        self.distance_to_obstacle = (base_bb.width / 2 + base_bb.depth / 2) / 2 + 0.5
+        if self.distance_to_obstacle is None:
+            self.distance_to_obstacle = OccupancyCostmap.default_distance_to_obstacle(
+                self.robot
+            )
+
+    def copy_for_world(self, world: World) -> GiskardLocationBackend:
```
Tigul, line 112: I dont think this method does what the name implies

coraplex/src/coraplex/locations/costmaps.py
Tigul: PRs are welcome

coraplex/src/coraplex/plans/executables.py
```diff
@@ -83,26 +89,15 @@ class GiskardExecutable(Executable):
     the motions, pre -and postconditions and the pause and interrupt calls.
     """
 
-    motion_mappings: Dict[MotionNode, Task] = field(kw_only=True)
```
Tigul, line 86: That should stay here

coraplex/src/coraplex/plans/executables.py
```diff
@@ -457,13 +398,18 @@ class UnderspecifiedExecutable(Executable):
...
-        from coraplex.plans.failures import PlanFailure, EmptyUnderspecified
+        from coraplex.plans.failures import (
+            PlanFailure,
+            EmptyUnderspecified,
+        )
 
         while self.node.advance():
-            try:
-                self.node.current_candidate.parse().execute()
+            with self.context.world.reset_state_context() as attempt:
```
Tigul, line 407: Why reset the state?

coraplex/src/coraplex/plans/factories.py
```diff
@@ -140,6 +142,8 @@ def make_node(action_like: ActionLike) -> PlanNode:
         return underspecified_action
     elif isinstance(action_like, ActionDescription):
         return ActionNode(designator=action_like)
+    elif isinstance(action_like, StandaloneMotion):
+        return StandaloneMotionNode(designator=action_like)
```
Tigul, line 146: What is a StadaloneMotionNode and why do we need this?

coraplex/src/coraplex/robot_plans/actions/core/container.py
Tigul: I have a new for handling parameter in the failure handling PR

Tigul (changes requested):

coraplex/demos/coraplex_kitchen_fridge_demo/demo.py
LucaKro: the reason i made those small methods is because this is the basis for the tutorial, so i extracted methods i didnt want the student to worry about. that way i could provide the implementation for parts of the code, while leaving the other half open

LucaKro: but i can see if some things can be moved deeper into the core framework, for the cases where thats not possible id like to keep it this way

coraplex/doc/source/_config.yml
LucaKro: I dont quite understand what you mean, can you elaborate? do you want me to change something, or are you asking a question?

coraplex/scripts/convert_exercises_for_self_assessment.sh
LucaKro: because the other script is in semdt, i guess i can abstract it into a upper level cram script

coraplex/self_assessment/exercises/0_writing_a_robot_plan.md
LucaKro: image

coraplex/scripts/test_exercises.sh
LucaKro: same answer

coraplex/self_assessment/exercises/0_writing_a_robot_plan.md
LucaKro: what formatting? do you mean the tags? or why the if statements are the way they are?
the tags are automatically treated a specific way when executing the different scripts, and the formatting is this way because at some point you told me i should remove assert statements from the exercises, so i changed it to this. They are in one line, because i didnt want to bloat the cells unnecessarily and its always the same pattern.

coraplex/self_assessment/exercises/0_writing_a_robot_plan.md
LucaKro: intended, as this is telling the student how to run it on their own. if its not commented out this is called when the cell is executed in the CI

coraplex/src/coraplex/locations/backends.py
LucaKro: problem is that if we are unlucky we need all the tries. In this tutorial like 75% or sth of the poses are sampled in bad spots when trying to get the milk from the fridge, because we have no way to condition these costmaps for smarter sampling. But this branches version is actually faster than on main branch, because i quickly skip through samples that cannot see the target object, so we have more samples we try, but less compute time

coraplex/src/coraplex/locations/backends.py
LucaKro: why not? it creates another Giskard Location Backend with the same parameters, but with the target updated for the new world instance.

coraplex/src/coraplex/plans/executables.py
LucaKro: why? it got replaced by motion_nodes, because i had the problem that motion mappings had bindings at the wrong time, so now we just have the motion nodes and the mappings are built lazily when they are needed.

coraplex/src/coraplex/plans/executables.py
LucaKro: because otherwise, if we do one attempt and it manipulates the state and then fails, we continue on a poisoned state no?

coraplex/src/coraplex/plans/factories.py
LucaKro: ill check again if we still need it or if i managed to achieve the same behavior using underspecified, but i remember there was a problem where we needed open and close to be standalone motions, otherwise the plan would not work when everything was in one motion statechart. Ill check again

coraplex/src/coraplex/robot_plans/actions/core/container.py
LucaKro: hmm interesting

coraplex/demos/coraplex_kitchen_fridge_demo/demo.py
LucaKro: i dont think so, not that i can find. for now id prefer to keep it here, as this is a function that is written by the student in chapter 3, knowledge representation and reasoning. So this is a bit of a red thread the student can orient on and feel like the tutorials are actually connected

coraplex/demos/coraplex_kitchen_fridge_demo/demo.py
LucaKro: hmm i looked at it just now, and i think i would like it to stay the way it is. as i said, this is for a tutorial, and i think having these kinds of methods here also allows the user to see the capabilities of the system in small, fairly selfcontained ways.

for example island_counter_top
by keeping it here and not, for example, abstracting it away, i can keep some things specific to this scenario (like the name, the method name etc) that i otherwise would need to generalize. which makes it harder for the student to understand, even if i point them to the right method.

coraplex/src/coraplex/plans/factories.py
LucaKro: okay i was able to recall the problem from my chat with simon. the problem was that when opening the door in a big motion statechart, the door would automatically open a few degrees even if the robot did not grasp it yet, because if there is a joint with no constraint on it, (which is the case here), sometimes small velocity values can be produced by the QP.

Our solution was that opening must be in its own statechart so that the door joint is never in a QP without a constraint. there may be a better solution, but that sounded like a small giskard side adventure, and i think having the option to say that a certain action should be in a standalone statechart makes sense, since we already needed that concept twice (once in underspecified, and once in model change )

coraplex/src/coraplex/plans/executables.py
LucaKro: hmm actually now that i think about it, im not sure if this is 100% correct. but i also wouldnt say its 100% wrong. 
I think with underspecified we should be able to specify a pick action with out, for example, naming the grasp it should use right? 

now if this were to be executed, how would this play out on a real robot? 1. would it actually run the plan and move? or 2. would it simulate it beforehand to get a valid parameterization?

I think currently its 1, but i think we kind of want 2 no?

because if its just 1, then if the robot tries out one obviously bad state that could easily discarded in the simulation, then we may end up in a bad state in real life that does not allow us to find any solution. if we did it in simulation first with reset state in between, like i do right now, then we can actually find a solution.


let me ground this in the concrete example where i had the problem:

when the robot tries to close the door, there is exactly 1 approach direction that works to grasp the handle without any unwanted collision etc: the BACK direction

Now when leaving this underspecified, the domain, as defined here:
```python
class ApproachDirection(Grasp):
    """
    Enum for the approach direction of a gripper.

    The AxisIdentifier is used to identify the axis of the gripper, and the int is used
    to identify the direction along  that axis.
    """

    FRONT = (AxisIdentifier.X, -1)
    BACK = (AxisIdentifier.X, 1)
    RIGHT = (AxisIdentifier.Y, -1)
    LEFT = (AxisIdentifier.Y, 1)
```

would always try grasps in the same order: FRONT -> BACK -> RIGHT -> LEFT

now it so happened, tha the front grasp, was actually approaching from the wrong side of the door, so the pr2 pressed against the door and failed. all good. but now we try the next grasp (which we know is the *only correct* grasp) WITHOUT the state reset. Now we dont have any near convex path from the current position behind the door, to the final handle grasp pose. this means giskard rightly fails, because it cannot find any path around. Now we try RIGHT and LEFT, and both fail. Now we sample a new position to try and close the door again. but we run into the same problem. FRONT is always tried first, then back, but back will fail because there is no path giskard can take.

So of course we could add some "recovery behavior" between runs, and in some cases this may be the correct solution. But i think we should *also* utilize underspecified to weed out the obviously bad parameterizations before we even try to move the real robot. In my opinion its fair to assume that if a parameterization does not work in simulation, it will not work in real life. so discarding all parameterizations that dont work in simulation, and having recovery behavior for parameterizations that pass in simulations but fail in real life sounds like the best compromise imo.


But also this sounds like a big side adventure, thats out of the scope of this PR. But this PR cant work without that feature, because the robot gets stuck and can never close the door. But also, thinking about it, this current state *should not be merged*, as resetting the state here actually sounds dangerous if this is used on the real robot

coraplex/demos/coraplex_kitchen_fridge_demo/demo.py
Tigul: But this is also something that would be useful in the main repo, so I'd say move or duplicate

LucaKro:
closed at least until we figure out the underspecified stuff

PR #557, opened 2026-08-21
[RDR] Splice 2: ConclusionAid (expert-labelling assistance)

AbdelrhmanBassiouny:
## What

Full detail: https://github.com/AbdelrhmanBassiouny/cognitive_robot_abstract_machine/pull/63


krrood/src/krrood/entity_query_language/rdr/aid.py
```diff
@@ -0,0 +1,49 @@
...
+
+from __future__ import annotations
+
+from typing_extensions import TYPE_CHECKING, Any, Optional
+
+if TYPE_CHECKING:
+    from krrood.entity_query_language.rdr.interface import CaseContext
+
+
+class ConclusionAid:
```
LucaKro: idk i dont like the name "Aid"

LucaKro: but i also dont know a better term to use rn :c

LucaKro (changes requested):
this code is not used anywhere but the tests, so to me this is dead code

krrood/src/krrood/entity_query_language/rdr/aid.py
tomsch420: perhaps suggestion?

test/krrood_test/test_eql_rdr/test_aid.py
tomsch420, line 1: these test look pretty useless

tomsch420 (changes requested):

krrood/src/krrood/entity_query_language/rdr/aid.py
AbdelrhmanBassiouny: so accepted until a better name is found?
Suggestion would only cover the suggest method, not the present method which actually can provide aids like visual aids for example if you are adding rules for semantic annotations it would single out the bodies in question or colour them

test/krrood_test/test_eql_rdr/test_aid.py
AbdelrhmanBassiouny: ok fair, there actual tests need the full RDR that's why these dummy ones where added instead of no tests. They can be removed.

AbdelrhmanBassiouny: removed

krrood/src/krrood/entity_query_language/rdr/aid.py
AbdelrhmanBassiouny: ok so I will implement a ConclusionHelper abstract base, and a mixin for each method, so a helper can implement either of the mixins or both. Each mixin is a ConclusionHelper. The mixins will be ConclusionSuggestor, and another ConclusionSupportPresenstor.

AbdelrhmanBassiouny: Done

tomsch420 (approved):

LucaKro (approved):

PR #559, opened 2026-08-22
[MultiSim] Fix lost world→sim pose writes that made test_world_sim_state_sync flaky

tomsch420:
## Problem

`test_world_sim_state_sync` fails intermittently in CI:

```
FAILED test_multi_sim.py::test_world_sim_state_sync - AssertionError: Box did not settle at target:
final_pos=[-2.83720686e-18  7.66259388e-15  1.49892245e-01], expected≈[0.3  0.2  0.15]
```

This is not a CI hardware-resource problem. The `z` coordinate in the failure is **correct** — the box did have time to fall and settle. It settled at the *origin* because the teleport to `(0.3, 0.2, 5.0)` was silently lost.

Reproduced by pinning to 2 cores with competing CPU load: **7 failures in 12 runs**, same signature. Tracing every `qpos` access in a failing run shows the write landing and then vanishing, with no write in between:

```
write6 MainThread before=[0.0, 0.0, 0.34] after=[0.3, 0.2, 5.0]   <- teleport written to qpos
read6  Thread-1   before=[-0.0, -0.0, 0.1392]                      <- gone, nothing wrote in between
```

## Cause

Two unsynchronised paths in `MujocoSynchronizer`, both of which drop a `world → sim` pose write without raising anything.

**1. `qpos` writes race `mj_step` (dominant).** `_on_state_change` writes `_mj_data.qpos` without taking the simulator's `_model_lock` — the lock `step_callback` holds across `mj_step`. The scene integrates with RK4, which saves the state at the top of the step and writes the integrated `qpos` back at the end, so a pose written while a step is in flight is overwritten and disappears; the body simply carries on from its old pose.

On an idle machine a step is short and the write almost always lands between steps. On a loaded runner the physics thread is descheduled mid-step and spends most of its wall time there, so the write lands inside a step often enough to fail the test. **The CPU shortage exposes the race rather than causing the failure** — which is why this only shows up in CI.

**2. Pose writes are not atomic against the `sim → world` pull.** `WorldState` takes `_world_lock` per accessor, so the seven degree-of-freedom writes in `Connection6DoF.origin` and the notification that pushes them are not one unit. A pull landing between them overwrites the new pose with the pre-write `qpos` values and — because it pauses the sibling state-change callback while it does so — swallows the notification that would have pushed the pose into MuJoCo. The closing `update_previous_world_state` then rebases the diff baseline, so no later notification sees the change either.

## Fix

- Hold `_world_lock` across the whole pull and the whole push, and across the writes plus notification in the three `origin` setters that have this shape (`Connection6DoF`, `OmniDrive`, `DifferentialDrive`).
- Take `_model_lock` around every `qpos` access, so pushes and pulls are serialised against `mj_step`.
- `_world_lock` stays the outer lock, matching the order `modify_world` already establishes (`_world_lock` → `_model_lock`), so no new lock ordering is introduced.
- The pull now resumes its callback in a `finally` block. An exception previously left it paused, silently disabling the `world → sim` direction for the rest of the run.

The test is **not** skipped and **not** marked flaky. The existing `@pytest.mark.flaky` is removed — it was inert, since no flaky plugin is installed and pytest only warned about an unknown mark.

## Tests

Two regression tests that drive each interleaving directly rather than waiting for it to happen under load, so they are deterministic:

- `test_pose_written_during_sim_to_world_pull_reaches_the_simulator` — holds the pull open inside `_read_6dof_from_qpos` while another thread writes a pose.
- `test_pose_write_waits_for_the_running_physics_step` — holds `_model_lock` from another thread, standing in for a step in flight, and asserts the push blocks until it is released and then lands.

Both **fail on the unfixed code** and pass with the fix.

## Verification

Run in a container built from `cram:latest` with the workspace re-synced (`uv sync --extra dev --active`).

| Check | Result |
|---|---|
| Starvation harness (2 cores + load), before | 7/12 failed |
| Starvation harness, after | **0/25 failed** |
| `test_multi_sim.py` | 18 passed |
| Full `semantic_digital_twin` suite, `-n 4` on 4 cores | **1048 passed, 0 failed**, 284.5s |
| Same suite on unpatched `main` (baseline) | 1046 passed, 282.6s |

No measurable cost from the added locking (284.5s vs 282.6s).

Note: 4 `ERROR collecting test_worlds/__init__.py - No module named 'pydrake'` appear in both the patched and baseline local runs — drake is installed by a separate CI step and is absent from the local container. Unrelated to this change.

### DIFF
--- a/semantic_digital_twin/src/semantic_digital_twin/adapters/multi_sim.py
+++ b/semantic_digital_twin/src/semantic_digital_twin/adapters/multi_sim.py
@@ -2929,6 +2929,18 @@ class MujocoSynchronizer(MultiSimSynchronizer):
         ``sync_rate_hz`` wall-clock Hz. The sibling state-change callback is
         paused across the write so our own ``notify_state_change`` does not
         echo back into :meth:`_on_state_change`.
+
+        The whole pull runs under ``World._world_lock``. ``WorldState`` takes
+        that lock per accessor, so holding it only for the individual writes
+        would let a writer thread land in the middle of this pull: the pose it
+        just wrote gets overwritten here with the pre-write qpos values, and
+        the notification that would have pushed that pose into MuJoCo is
+        swallowed by the ``pause()`` below. The write is then lost without a
+        trace, because the closing ``update_previous_world_state`` rebases the
+        diff baseline onto what this pull just wrote. ``_world_lock`` is the
+        outermost lock in the codebase -- ``modify_world`` takes it before the
+        simulator's ``_model_lock`` -- so acquiring it here preserves that
+        order.
         """
         if self.sync_rate_hz <= 0:
             return
@@ -2937,34 +2949,43 @@ class MujocoSynchronizer(MultiSimSynchronizer):
             return
         self._last_sync_time = now
 
-        changed = False
-        self._state_callback.pause()
+        with self._world._world_lock:
+            changed = False
+            self._state_callback.pause()
+            try:
+                # Read qpos under the model lock so the whole pull sees one
+                # coherent post-step state rather than a mixture of poses from
+                # either side of an ``mj_step`` running on another thread.
+                with self.simulator._model_lock:
+                    for connection in self._world.connections:
+                        if isinstance(connection, FixedConnection):
+                            continue
+                        qpos_adr = self._resolve_qpos_adr(connection)
+                        if qpos_adr is None:
+                            continue
 
-        for connection in self._world.connections:
-            if isinstance(connection, FixedConnection):
-                continue
-            qpos_adr = self._resolve_qpos_adr(connection)
-            if qpos_adr is None:
-                continue
+                        if isinstance(connection, Connection6DoF):
+                            self._read_6dof_from_qpos(connection, qpos_adr)
+                            changed = True
+                        elif isinstance(connection, ActiveConnection1DOF):
+                            self._read_1dof_from_qpos(connection, qpos_adr)
+                            changed = True
+                        else:
+                            logger.warning(
+                                "sim→world sync: unsupported connection type %s "
+                                "for joint %s; skipping",
+                                type(connection).__name__,
+                                connection.name.name,
+                            )
 
-            if isinstance(connection, Connection6DoF):
-                self._read_6dof_from_qpos(connection, qpos_adr)
-                changed = True
-            elif isinstance(connection, ActiveConnection1DOF):
-                self._read_1dof_from_qpos(connection, qpos_adr)
-                changed = True
-            else:
-                logger.warning(
-                    "sim→world sync: unsupported connection type %s for "
-                    "joint %s; skipping",
-                    type(connection).__name__,
-                    connection.name.name,
-                )
-
-        if changed:
-            self._world.notify_state_change()
-            self._state_callback.update_previous_world_state()
-        self._state_callback.resume()
+                if changed:
+                    self._world.notify_state_change()
+                    self._state_callback.update_previous_world_state()
+            finally:
+                # Always resume: a callback left paused by an exception would
+                # silently disable the world -> sim direction for the rest of
+                # the run.
+                self._state_callback.resume()
 
     def _write_6dof_to_qpos(
         self,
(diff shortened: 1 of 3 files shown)


semantic_digital_twin/src/semantic_digital_twin/adapters/multi_sim.py
```diff
@@ -2937,34 +2949,43 @@ def _sim_to_world(self) -> None:
...
-
-        if changed:
-            self._world.notify_state_change()
-            self._state_callback.update_previous_world_state()
-        self._state_callback.resume()
+        with self._world._world_lock:
+            changed = False
+            self._state_callback.pause()
+            try:
+                # Read qpos under the model lock so the whole pull sees one
```
tomsch420: too much nesting, extract this into a method

semantic_digital_twin/src/semantic_digital_twin/adapters/multi_sim.py
```diff
@@ -3036,51 +3057,67 @@ def _on_state_change(self) -> None:
...
+            # the physics thread. Writing qpos without it is not just a torn
+            # read: the model integrates with RK4, and ``mj_step`` writes the
+            # integrated qpos back from state it saved at the top of the step,
+            # so a write that lands mid-step is overwritten and vanishes. The
+            # body then simply carries on from its old pose, with nothing
+            # raised anywhere. Acquired inside ``_world_lock`` to match the
+            # order ``modify_world`` already establishes.
+            with self.simulator._model_lock:
+                for connection in self._world.connections:
+                    if isinstance(connection, FixedConnection):
```
tomsch420: also too much nesting

semantic_digital_twin/src/semantic_digital_twin/adapters/multi_sim.py
tomsch420: Done in 58c4861f3.

Both directions walked `self._world.connections` with the same skip-fixed / resolve-qpos-address / dispatch-on-type logic, so I pulled that out into `_joint_backed_connections`, a generator yielding `(connection, qpos_adr)`. Each direction now gets one method over it — `_read_connections_from_qpos` here and `_write_connections_to_qpos` for the push — and each owns its own `_model_lock` acquisition, so the lock sits next to the code it protects and the reason for holding it lives in that method's docstring rather than in a comment buried six levels deep.

`_sim_to_world` is now just the throttle, the lock, and pause/resume around a single call. The duplicated unsupported-connection warning became one `_warn_unsupported_connection` helper.

No behaviour change — same locks, same order, same work. Re-verified after the refactor: `test_multi_sim.py` 18 passed, and the CPU-starvation harness that reproduced the original CI failure 7/12 is 0/25.

semantic_digital_twin/src/semantic_digital_twin/adapters/multi_sim.py
tomsch420: Done in 58c4861f3, same extraction as the pull side.

The body of this loop is now `_write_connections_to_qpos(positions, previous_positions)`, which iterates the shared `_joint_backed_connections` generator and takes `_model_lock` itself. The RK4/`mj_step` rationale for that lock moved from an inline comment into that method's docstring, next to the code it explains.

`_on_state_change` is left with just the two things that are actually its own concern: the `_world_lock`, and the shape check on `positions` vs `previous_positions` that decides whether a column-wise diff is meaningful.

semantic_digital_twin/src/semantic_digital_twin/adapters/multi_sim.py
```diff
@@ -2883,6 +2892,101 @@ def _decompose_pose_matrix(
...
+        with the qpos address of that joint.
+
+        Fixed connections carry no DoFs, and a connection that does not resolve
+        to a joint is not in the compiled model, so neither has anything to
+        sync. Both sync directions walk the same set, so they share this.
+        """
+        for connection in self._world.connections:
+            if isinstance(connection, FixedConnection):
+                continue
+            qpos_adr = self._resolve_qpos_adr(connection)
```
tomsch420: no abbrivations

semantic_digital_twin/src/semantic_digital_twin/adapters/multi_sim.py
```diff
@@ -2883,6 +2892,101 @@ def _decompose_pose_matrix(
...
+
+        Held under ``_model_lock`` so the whole pull sees one coherent
+        post-step state rather than a mixture of poses from either side of an
+        ``mj_step`` running on the physics thread.
+
+        :return: Whether any connection was read.
+        """
+        changed = False
+        with self.simulator._model_lock:
+            for connection, qpos_adr in self._joint_backed_connections():
```
tomsch420: no abbrivations

semantic_digital_twin/src/semantic_digital_twin/adapters/multi_sim.py
tomsch420: Done in 35209b556 — `qpos_adr` → `qpos_address`, and `_resolve_qpos_adr` → `_resolve_qpos_address`, throughout the MuJoCo adapter and its tests (35 occurrences). Mechanical rename, `test_multi_sim.py` 18 passed.

I checked the rest of my diff for other abbreviations and found none — the remaining short names are MuJoCo API terms (`qpos`, `mj_data`, `mj_step`) or the codebase's existing DoF vocabulary.

Note the same abbreviation still appears 6 times in `physics_simulators/mujoco_simulator.py`, which this PR doesn't otherwise touch. I left it so the diff stays scoped to the change under review — happy to rename it here or in a follow-up, whichever you prefer.

semantic_digital_twin/src/semantic_digital_twin/adapters/multi_sim.py
tomsch420: Done in 35209b556 — `qpos_adr` → `qpos_address`, and `_resolve_qpos_adr` → `_resolve_qpos_address`, throughout the MuJoCo adapter and its tests (35 occurrences). Mechanical rename, `test_multi_sim.py` 18 passed.

I checked the rest of my diff for other abbreviations and found none — the remaining short names are MuJoCo API terms (`qpos`, `mj_data`, `mj_step`) or the codebase's existing DoF vocabulary.

Note the same abbreviation still appears 6 times in `physics_simulators/mujoco_simulator.py`, which this PR doesn't otherwise touch. I left it so the diff stays scoped to the change under review — happy to rename it here or in a follow-up, whichever you prefer.

semantic_digital_twin/src/semantic_digital_twin/adapters/multi_sim.py
```diff
@@ -2883,16 +2892,111 @@ def _decompose_pose_matrix(
         quat_xyzw = Rotation.from_matrix(pose[:3, :3]).as_quat()
         return xyz, quat_xyzw
 
-    def _read_6dof_from_qpos(self, connection: Connection6DoF, qpos_adr: int) -> None:
+    def _joint_backed_connections(self) -> Iterator[Tuple[Connection, int]]:
```
AbdelrhmanBassiouny: dataclass instead of Tuple

semantic_digital_twin/src/semantic_digital_twin/adapters/multi_sim.py
```diff
@@ -2883,16 +2892,111 @@ def _decompose_pose_matrix(
...
+        for connection in self._world.connections:
+            if isinstance(connection, FixedConnection):
+                continue
+            qpos_address = self._resolve_qpos_address(connection)
+            if qpos_address is None:
+                continue
+            yield connection, qpos_address
+
+    @staticmethod
+    def _warn_unsupported_connection(direction: str, connection: Connection) -> None:
```
AbdelrhmanBassiouny, line 2936: param docs

semantic_digital_twin/src/semantic_digital_twin/adapters/multi_sim.py
```diff
@@ -2883,16 +2892,111 @@ def _decompose_pose_matrix(
...
+        Held under ``_model_lock`` so the whole pull sees one coherent
+        post-step state rather than a mixture of poses from either side of an
+        ``mj_step`` running on the physics thread.
+
+        :return: Whether any connection was read.
+        """
+        changed = False
+        with self.simulator._model_lock:
+            for connection, qpos_address in self._joint_backed_connections():
+                if isinstance(connection, Connection6DoF):
```
AbdelrhmanBassiouny: match .. case ...?

semantic_digital_twin/src/semantic_digital_twin/adapters/multi_sim.py
```diff
@@ -2883,16 +2892,111 @@ def _decompose_pose_matrix(
...
+        establishes.
+
+        :param positions: The current ``world.state`` positions.
+        :param previous_positions: The positions as of the last notification,
+            used to find what changed. Must be the same length as ``positions``.
+        """
+        state_index = self._world.state._index
+        with self.simulator._model_lock:
+            for connection, qpos_address in self._joint_backed_connections():
+                if isinstance(connection, Connection6DoF):
```
AbdelrhmanBassiouny: same?

semantic_digital_twin/src/semantic_digital_twin/adapters/multi_sim.py
```diff
@@ -2883,16 +2892,111 @@ def _decompose_pose_matrix(
...
+                        connection,
+                        qpos_address,
+                        positions,
+                        previous_positions,
+                        state_index,
+                    )
+                else:
+                    self._warn_unsupported_connection("world→sim", connection)
+
+    def _read_6dof_from_qpos(self, connection: Connection6DoF, qpos_address: int) -> None:
```
AbdelrhmanBassiouny, line 3024: param docs

semantic_digital_twin/src/semantic_digital_twin/adapters/multi_sim.py
```diff
@@ -2937,46 +3053,29 @@ def _sim_to_world(self) -> None:
...
+                if self._read_connections_from_qpos():
+                    self._world.notify_state_change()
+                    self._state_callback.update_previous_world_state()
+            finally:
+                # Always resume: a callback left paused by an exception would
+                # silently disable the world -> sim direction for the rest of
+                # the run.
+                self._state_callback.resume()
 
     def _write_6dof_to_qpos(
```
AbdelrhmanBassiouny, line 3108: param docs

semantic_digital_twin/src/semantic_digital_twin/adapters/multi_sim.py
```diff
@@ -3006,81 +3105,56 @@ def _write_6dof_to_qpos(
...
-        mj_data.qpos[qpos_adr + 6] = mj_quat_xyzw[2]
+        mj_data.qpos[qpos_address + 0] = mj_xyz[0]
+        mj_data.qpos[qpos_address + 1] = mj_xyz[1]
+        mj_data.qpos[qpos_address + 2] = mj_xyz[2]
+        mj_data.qpos[qpos_address + 3] = mj_quat_xyzw[3]
+        mj_data.qpos[qpos_address + 4] = mj_quat_xyzw[0]
+        mj_data.qpos[qpos_address + 5] = mj_quat_xyzw[1]
+        mj_data.qpos[qpos_address + 6] = mj_quat_xyzw[2]
 
     def _write_1dof_to_qpos(
```
AbdelrhmanBassiouny, line 3163: param docs

test/semantic_digital_twin_test/test_adapters/test_multi_sim.py
```diff
@@ -851,6 +852,218 @@ def spawn_state_sync_scene(
         stop_multisim_if_running(multi_sim)
 
 
+def _build_box_on_plane_world() -> tuple[World, Body, Connection6DoF]:
```
AbdelrhmanBassiouny: dataclass for return?

AbdelrhmanBassiouny (changes requested):

AbdelrhmanBassiouny (approved):

LucaKro (approved):

PR #563, opened 2026-08-24
[EQL] Correlate query-rooted attributes used in a query's own conditions

AbdelrhmanBassiouny:
## The bug

Full detail: https://github.com/AbdelrhmanBassiouny/cognitive_robot_abstract_machine/pull/182


LucaKro (approved):

krrood/src/krrood/entity_query_language/query/query.py
```diff
@@ -288,6 +294,98 @@ def _evaluate_natively_(self) -> Iterator:
             return self._expression_.evaluate()
         return MultiArityExpressionThatPerformsACartesianProduct.evaluate(self)
 
+    def _correlate_conditions_(
```
tomsch420: is correlate the correct wording? To me correlate is statistics, but what you do seems to me as moving related conditions where they belong.

tomsch420 (changes requested):

krrood/src/krrood/entity_query_language/query/query.py
AbdelrhmanBassiouny: yeah true, the name is misleading

tomsch420 (approved):

PR #567, opened 2026-08-24
RoboKudo ROS lifecycle improvement after stretch PR

Sanic:
**Motivation**

Recent Stretch demo changes made the analysis-engine composition tests construct CollectionReader instances, which in turn create ROS-backed camera interfaces. Previously, those interfaces created and spun independent ROS nodes, causing lifecycle exceptions in tests that build analysis engines without the full RoboKudo runtime. Additional problems in Analysis Engines discovered by the tests mentioned above has been addressed as well.

### Summary

- Centralized RoboKudo ROS node ownership in a singleton-backed node registry.
  - Runtime code can initialize and register the central node.
  - ROS camera interfaces reuse that registered node rather than creating their own.
  - Tests register and clear their node explicitly.
  - Missing ROS node initialization now raises `RoboKudoROSNodeMissing`.

- Stabilized ROS and query test execution.
  - Removed camera-interface background spinning. All Camera interfaces now properly subscribe to the main node.
  - Reworked tree ticking to avoid timer cancellation/lifecycle issues.
  - Made the query test client own its ROS context and executor, and corrected multiprocessing result handling.

- Protected persisted RoboKudo storage during tests.
  - `StorageWriter` defers database cleanup until `setup()` and executes it only once. This allows us now to initialize AEs with `StorageWriter` without dropping development data. 
  - RoboKudo’s test suite now checks that an existing default Mongo database and its CAS document count remain unchanged. This is scoped to `test/robokudo_test`, not the repository-wide test configuration.

- Updated affected analysis engines for the current `py_trees` API.

- Added missing RoboKudo runtime dependencies, `transformers` and `faiss-cpu`, required when the composition sweep imports the CLIP annotator.


robokudo/src/robokudo/io/ros.py
```diff
@@ -1,17 +1,65 @@
...
-"""
-Central RoboKudo ROS node.
-"""
+from robokudo.exceptions import RoboKudoROSNodeMissing
+
+# %% central node registry
+
+
+@dataclass
+class RoboKudoNodeRegistry(metaclass=SingletonMeta):
```
LucaKro: i think @ichumuh has sth similar in giskard with rospy? maybe this is duplicated code?

robokudo/requirements.txt
```diff
@@ -9,4 +9,6 @@ transforms3d
 opencv-python>=4.11.0.86
 matplotlib>=3.0
 setuptools
-pooch
\ No newline at end of file
+pooch
+transformers
```
LucaKro: what is this for? 👀 pooch and transformers?

test/robokudo_test/conftest.py
```diff
@@ -1,9 +1,105 @@
...
+import robokudo.io.ros
+from robokudo.descriptors.camera_configs.config_mongodb_playback import (
+    MongoCameraConfig,
+)
+
+# %% Mongo storage preservation
+
+
+@dataclass(frozen=True)
+class MongoStorageSnapshot:
```
LucaKro, line 24: hmm i feel like you are starting to spend time supporting MongoDB instead of committing to the switch to our SQL stuff 🥲

LucaKro (changes requested):
good job, thank you 🙏

robokudo/src/robokudo/annotators/storage.py
```diff
@@ -68,11 +69,24 @@ def __init__(
         super().__init__(name, descriptor)
         self.rk_logger.debug("%s.__init__()" % self.__class__.__name__)
         self.storage = Storage(self.descriptor.parameters.db_name)
+        self.storage_preparation_completed: bool = False
+        """Whether storage setup has already run for this writer."""
+
+    def setup(self, **kwargs: Any) -> bool:
```
tomsch420, line 77: use correct docstring syntax

robokudo/src/robokudo/descriptors/analysis_engines/nested_pipeline.py
```diff
@@ -81,20 +82,27 @@ def implementation(self) -> Pipeline:
         """
         kinect_config = CollectionReaderDescriptorFactory.create_descriptor("kinect")
 
-        # create second 'pipeline
-        second_seq = Sequence(name="BS Pipeline")
+        second_seq = Sequence(name="BS Pipeline", memory=True)
```
tomsch420: bullshit pipeline? no abbrivations

robokudo/src/robokudo/descriptors/analysis_engines/nested_pipeline.py
```diff
@@ -81,20 +82,27 @@ def implementation(self) -> Pipeline:
...
         second_seq.add_children(
             [
-                Count(
-                    name="Annotator A", fail_until=-1, running_until=9, success_until=10
+                TickCounter(
+                    name="Annotator A",
```
tomsch420: Annotator A is not a very descriptive name

robokudo/src/robokudo/io/ros.py
tomsch420: yeah i also get duplication vibes here

robokudo/src/robokudo/scripts/query_test_client.py
```diff
@@ -39,14 +51,19 @@ class RoboKudoActionClient(Node):
     Allows sending dynamic goals and handles feedback, result, and cancellation.
     """
 
-    def __init__(self, preempt_timer: Optional[float] = None) -> None:
+    def __init__(
```
tomsch420: no custom inits

robokudo/src/robokudo/scripts/query_test_client.py
tomsch420, line 1: why is the scripts folder in source?

robokudo/src/robokudo/exceptions.py
```diff
@@ -12,12 +14,32 @@
...
+    """
+    Raised when a component requires the central RoboKudo ROS node.
+    """
+
+    def error_message(self) -> str:
+        return "RoboKudo ROS node is not initialized."
+
+    def suggest_correction(self) -> str:
+        return (
+            "initialize RoboKudo's central ROS node before constructing ROS-backed "
```
tomsch420: can you also provide here how to do it code wise?

robokudo/src/robokudo/exceptions.py
```diff
@@ -12,12 +14,32 @@
...
+            "interfaces."
+        )
 
 
 @dataclass
 class ColorToDepthRatioMissing(RoboKudoError, RuntimeError):
-    """Raised when COLOR2DEPTH_RATIO is required but missing."""
+    """
+    Raised when COLOR2DEPTH_RATIO is required but missing.
+    """
```
tomsch420, line 26: these are missing the methods error_message and suggest_correction, arent they?

test/robokudo_test/conftest.py
tomsch420, line 24: ormatic integration where? Also if you use ormatic, you get access to a handy in-memory sqlite db which makes this obsolete

tomsch420 (changes requested):

robokudo/src/robokudo/scripts/query_test_client.py
ichumuh: to avoid name conflicts if you want to link them in ros packages

robokudo/src/robokudo/io/ros.py
ichumuh: yes that's basically rospy

robokudo/requirements.txt
Sanic: pooch is used to download required binary data on-the-fly and cache it.
transformers is a library to implement transformer-based NN approaches, like CLIP for example.

test/robokudo_test/conftest.py
Sanic: This was super quick to implement and should just keep an eye on any new tests that might delete locally stored data on our dev machines. Something similar will be in place when the switch to SQL/Ormatic is done.

robokudo/src/robokudo/descriptors/analysis_engines/nested_pipeline.py
Sanic: Thanks for the hint. I've addressed that in the most recent commit.

robokudo/src/robokudo/descriptors/analysis_engines/nested_pipeline.py
Sanic: Thanks for the hint. I've addressed this in the most recent commit as well.

robokudo/src/robokudo/annotators/storage.py
Sanic: Fixed!

robokudo/src/robokudo/io/ros.py
ichumuh: talked to @Sanic its not the same as my rospy, he needs more. we decided to merge the two and move them to semantic_digital_twin/adapters/ros

robokudo/src/robokudo/scripts/query_test_client.py
Sanic: This class is a subclass of a rclpy.Node. rclpy.Node needs the Context during __init__.
Therefore i'd like to keep it as-is.

robokudo/src/robokudo/exceptions.py
Sanic: Good point! I've added a hint how to handle it code-wise in the new, more general ROSNodeNotRegisteredError Exception.

robokudo/src/robokudo/exceptions.py
Sanic: I just checked and they were already in place

Screenshot from 2026-08-31 11-59-10

test/robokudo_test/conftest.py
Sanic: Still on the agenda, but not a focus of this PR.

semantic_digital_twin/src/semantic_digital_twin/adapters/ros/node_registry.py
```diff
@@ -0,0 +1,78 @@
...
+    entities through that node while its registration is stable. They must not destroy
+    the node, change its executor ownership, or shut down its ROS context.
+
+    The lifecycle owner serializes calls to :meth:`register` and :meth:`clear`. It must
+    stop users of the node before clearing the registration, and clear it before
+    destroying the node. This registry does not initialize ROS, create or spin an
+    executor, destroy the node, or shut down its context.
+    """
+
+    _node: Node | None = field(default=None, init=False, repr=False)
```
ichumuh: i'm confused, does this only allow a single node? i thought you told me that its different from my rospy, because you need multiple nodes per process?

ichumuh (approved):

Sanic:
Hi @tomsch420 and @LucaKro 
I've addressed your comments. Regarding your suggestions to look into the ROS Stuff that Simon did: Simon and i discussed a common ground for the basic ROS Node registration and placed it into semantic_digital_twin.
Please re-review.

LucaKro (approved):

giskardpy/src/giskardpy/middleware/ros2/rospy.py
```diff
@@ -30,11 +35,27 @@ def _patched_action_client_init(self, *args, **kwargs):
 ActionClient.__init__ = _patched_action_client_init
 
 
-def spinner_thread_target():
+def get_node() -> Node:
```
tomsch420, line 38: im getting singleton vibes here. Use krrood singleton for unified handling of such things if you want singleton behavior here

tomsch420: ok i just saw its singeton already, what do you need this for then?

robokudo/src/robokudo/scripts/query_test_client.py
tomsch420: i can only strongly discourage inheriting from external classes like this. Especially in ROS, it is known that their class design and interfaces often clashes with how we do it, and this just inherits that clashing.. Just make an assoication with a ros node isntead here, this fixes the clashing

tomsch420 (changes requested):

giskardpy/src/giskardpy/middleware/ros2/rospy.py
ichumuh: this is to preserve the rospy api from giskard. using `ROSNodeRegistry().get()` instead of `rospy.get_node()` would make the interface inconsistent. i guess you could give the node registry the logic for spinning, but patrick actually doesn't need that, he always needs a different executor or spinning logic. so i say merging this further is yagni for now

tomsch420: ok, i trust the casadi friend

robokudo/src/robokudo/scripts/query_test_client.py
```diff
@@ -44,42 +45,76 @@ def pretty_print_result(self, result: Query_Result) -> str:
...
-class RoboKudoActionClient(Node):
+@dataclass(eq=False)
+class RoboKudoActionClient:
+    """
+    Send dynamic goals to the RoboKudo query action server.
     """
-    Debug Action Client for the Query Action Server.
 
-    Allows sending dynamic goals and handles feedback, result, and cancellation.
+    node: Node
```
tomsch420, line 67: im super confused now. It looks like you repeat everything from node that is in node already

tomsch420 (changes requested):

robokudo/src/robokudo/scripts/query_test_client.py
Sanic: Agreed. I usually wouldn't do that for other stuff. For ROS2 action servers its the teached way of defining Action Clients though: https://docs.ros.org/en/jazzy/Tutorials/Intermediate/Writing-an-Action-Server-Client/Py.html
But in my latest commit i've switched the class design to simply have the Node as a member and make everything a dataclass.

robokudo/src/robokudo/scripts/query_test_client.py
Sanic: I basically just moved the member definitions from the old `__init__` up to the definition of the dataclass and defined a Node reference instead of inheriting from it. Is there something i need to change?

tomsch420: in the old case, you inherited from node. Now you have a node, so why do you also have the parameters of node? I could just do Client(node=Node(node_params)), now it looks like i need Client(node_params, node=Node(node_params)). Is this not the case?

Sanic: You can do the first case: Client(node=Node(node_params)). You don't need to do the latter. 
The only additional parameter for initialization is preempt_timer, which belongs to the action client and is unrelated to the node configuration.  All other fields are used to manage the action handling. They neither belong to the node configuration nor duplicate variables from rclpy.Node.

tomsch420 (approved):

PR #569, opened 2026-08-25
Declare workspace dependencies statically in pyproject.toml

AbdelrhmanBassiouny:
## What

Dependencies are now maintained in exactly one place: `[project] dependencies` of each workspace member's `pyproject.toml`. All `requirements*.txt` files are removed.

- Every member's list is a 1:1 copy of its former `requirements.txt` (specifiers kept, comments kept where they carried information, exact duplicates dropped), **plus the sibling workspace packages its `src/` tree imports** - several of those were never declared before (e.g. `coraplex` → `giskardpy`, `krrood`; `semantic_digital_twin` → `giskardpy`, `krrood`, `physics_simulators`) and only worked because the root meta-package installs everything.
- `requirements-dev*.txt` became `[project.optional-dependencies] dev` (krrood, random_events, probabilistic_model); `semantic_digital_twin/requirements-self-assessment.txt` became the `self-assessment` extra.
- `random_events/setup.py` no longer passes `install_requires` (it would conflict with the static list).
- Consumers switched to `pip install -e .` / `pip install -e ".[dev]"`: the two coraplex Dockerfiles, `coraplex/scripts/install.sh`, the segmind devcontainer setup, the READMEs, and the per-package workflows under `.github/workflows/<package>/` (note: those live in subdirectories, so GitHub does not run them; updated for consistency only). `random_events-publish-to-pypi.yml` drops `CIBW_BEFORE_BUILD` - the build-system requirements already cover the isolated build.
- New workspace test `test/version_test/test_dependency_declarations.py` (runs in the `version` CI job): no member may list `dependencies` as dynamic, and every imported sibling must be declared. Against `main` it fails for 9 of 10 members; on this branch it passes.

## Why

`dynamic = ["dependencies"]` read from `requirements.txt` is invisible to every tool that reads `pyproject.toml` without building the package. uv's workspace graph only had the edges setuptools happened to produce, and PyCharm 2026.2's pyproject-based project model generated no dependencies between the workspace modules at all, so cross-package imports (`coraplex`, `semantic_digital_twin`, ...) were flagged unresolved in every member. With static lists PyCharm regenerates the module dependencies by itself on every `pyproject.toml` change (verified).

## Verified

- `uv lock` resolves the workspace (390 packages); `uv build --sdist --package random_events` produces correct `Requires-Dist` metadata.
- `pytest test/version_test` → 21 passed.

## Worth deciding

- `giskardpy`, `physics_simulators` and `robokudo` are not on PyPI. Declaring them makes `pip install coraplex` / `semantic_digital_twin` from PyPI fail at resolution instead of at import time as it does today; publishing those packages (or marking the imports optional) is a follow-up.
- The lists were migrated faithfully, not cleaned: krrood still ships `black`, `ruff`, `pytest`, `pdbpp`, `ucimlrepo` as runtime dependencies, semantic_digital_twin lists the obsolete `pathlib` backport, giskardpy lists `pip`. Easy follow-ups now that there is one place to edit.

Session: https://claude.ai/code/session_01RW8d6gP6dZqRpH48sY3F9a

https://claude.ai/code/session_01RW8d6gP6dZqRpH48sY3F9a


coraplex/pyproject.toml
```diff
@@ -20,7 +20,31 @@ classifiers = [
...
+    "sqlacodegen==3.0.0",
+    "tqdm==4.66.3",
+    "psutil~=5.9.0",
+    "typing_extensions>=4.10.0",
+    "probabilistic_model>=7.1.0",
+    "random_events>=4.1.18",
+    "numpy",
+    "objgraph",
+    "pint>=0.21.1",
+    "transforms3d",
```
LucaKro, line 37: i think we may need to clean up dependencies at some point, i dont think we need this anymore

LucaKro (approved):

coraplex/requirements.txt
Tigul, line 1: Add them to the coraplex pyproject.toml

Tigul (changes requested):
Did you check if this works with the PyCharm project structure?

AbdelrhmanBassiouny:
> Did you check if this works with the PyCharm project structure?

So the motivation behind this was that pycharm was showing me everything thing as red , and this fixed that.

coraplex/pyproject.toml
AbdelrhmanBassiouny: Yeah, follow up prs can do this

coraplex/requirements.txt
AbdelrhmanBassiouny: They are added, you didn't see them?

Tigul: Ups, my bad

Tigul (approved):

tomsch420 (changes requested):
Please check if the requirement groups from Jonas PR are also applicable here

tomsch420 (approved):

PR #570, opened 2026-08-26
Keep NygaInduction extreme-point widening from overlapping a sibling leaf

tomsch420:
## Summary

`NygaInduction.adjust_extreme_points` widens only the lowest and highest leaf of
*its own* circuit by `tolerance_at_extremes`, with no knowledge of any other
`NygaInduction` instance. `JointProbabilityTree.create_leaf_node` fits one such
instance independently per (decision-tree leaf, continuous variable) and mounts
them all into a shared circuit. When two sibling leaves split on the same
variable with data denser than `2 * tolerance_at_extremes` near the split
point, each leaf's independent widening can cross past the other leaf's
boundary, making the mounted circuit non-deterministic - even though a Nyga
distribution is supposed to be deterministic by construction.

- `NygaInduction` now carries a `support: SimpleInterval` field. Every uniform
  piece `InductionStep.create_uniform_distribution_from_indices` builds is
  intersected with it; the piece touching the globally first or last datapoint
  is constructed unbounded on that side beforehand, so the intersection is
  what pins its final boundary - whether that comes from the default
  tolerance widening or from a caller-supplied support. If no support is
  given, `fit()` defaults it to the data's own range widened by
  `tolerance_at_extremes` on both sides, matching the previous default
  behavior.
- `JointProbabilityTree.create_leaf_node` computes a tighter support per leaf
  via `leaf_support`, using `random_events` interval/nearest-neighbor logic
  over the raw data of every row *outside* that leaf for the variable being
  fit. This is independent of leaf processing order, since it does not rely
  on what has already been mounted into the shared circuit.
- Along the way, fixing this exposed a separate, pre-existing bug in
  `random_events`: `SimpleInterval.left`/`.right` setters passed a raw int to a
  pybind11 field typed as the native `BorderType` enum, which raised a
  `TypeError` on any use (nothing in the codebase called these setters before
  this fix needed to). That's fixed in its own commit.

## Test plan
- [x] `NygaInductionSupportTestCase`/`SupportedUniformDistributionTestCase` in
      `test_nyga_distribution.py` cover the new support-intersection mechanism
      directly (default support, a given support being respected, edge vs.
      internal pieces).
- [x] `DenseBoundaryDeterminismTestCase` in `test_jpt.py` reproduces the
      dense-boundary scenario through `create_leaf_node` and asserts
      `is_deterministic()`.
- [x] Verified new/changed tests fail against the pre-fix code, then pass
      after the fix.
- [x] Full `random_events` + `probabilistic_model` suites: 468 passed, 1
      skipped (unrelated), no regressions.

Made with the help of Claude in session https://claude.ai/code/session_012Q5tR41rBWzTBNmE4sLZmR

### DIFF
--- a/probabilistic_model/src/probabilistic_model/learning/nyga_induction.py
+++ b/probabilistic_model/src/probabilistic_model/learning/nyga_induction.py
@@ -7,7 +7,7 @@ from typing import Optional, List, Deque, Tuple, Dict, Any
 import numpy as np
 import numpy.typing as npt
 import random_events
-from random_events.interval import closed, closed_open, SimpleInterval, Bound
+from random_events.interval import closed, closed_open, Interval, SimpleInterval, Bound
 from random_events.product_algebra import SimpleEvent, Event
 from random_events.variable import Continuous
 from typing_extensions import Self
@@ -419,13 +419,21 @@ class NygaInduction:
     """
 
     def fit(
-        self, data: np.ndarray, weights: Optional[np.ndarray] = None
+        self,
+        data: np.ndarray,
+        weights: Optional[np.ndarray] = None,
+        forbidden_region: Optional[Interval] = None,
     ) -> ProbabilisticCircuit:
         """
         Fit the distribution to the data.
 
         :param data: The data to fit the distribution to.
         :param weights: The optional log_weights of the data points.
+        :param forbidden_region: Space that the fitted support must not overlap, for
+            example the already-fitted support of a sibling distribution for the same
+            variable elsewhere in a larger circuit. Only constrains
+            :meth:`adjust_extreme_points`; the induced quantiles themselves are
+            unaffected.
 
         :return: The fitted distribution.
         """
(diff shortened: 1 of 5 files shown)


probabilistic_model/src/probabilistic_model/learning/nyga_induction.py
```diff
@@ -411,6 +425,18 @@ class NygaInduction:
     float precision.
     """
 
+    support: Optional[SimpleInterval] = None
```
tomsch420: Make this per default the interval on the real line, then the code should have less annoying checks

tomsch420: Done - `support` now defaults to `reals().simple_sets[0]` via a `default_factory`, and all three `is not None` checks (two in `create_uniform_distribution_from_indices`, one in `fit`) are gone. `fit()` now unconditionally narrows `support` by intersecting it with the tolerance window around the data, which is a no-op when support is still the default real line, and correctly narrows a caller-supplied support too. Pushed as b1fa60c1b.

LucaKro (approved):

davidprueser (approved):

PR #571, opened 2026-08-26
Human readable unique names

LucaKro:
TF name conflict resolution selfcontained within the tf publisher, and uniqueness conflict resolution using UUIDs.

I had critiques of Jonas first solutions, which was very similar to this one. here is my updated stances on it after playing around with namespaces alot.

> if we have a name "table", and a name "table_1" then the "unique" name for "table" could generate a name thats not actually unique no? also if the dict is ever cleaned up you may give the same index multiple times.

these kind of name conflicts can never happen here because we use the UUID

> also we have "WorldEntity.update_name", and "World.force_root_name" which update name in place, but this is not invalidated in the map. also then a new unique name would be generated.

the entity is remembered via its id, so its name will simply not change in the TF tree

> removing kinematic structure entities never invalidates the entries in the dict

true, but i think its fine since a name can never given twice anyways, so this is at worst a small memory leak of the uuid and a name contained within an instance of the tf publisher

> is not thread safe if im not mistaken, which has caused problems in the past.

This version is, as there arent multiple threads working on the TF publisher, with jonas version where it was inside the world it did

> since this is not synchronized, and you use this as a TF frame name, what can happen is that two different processes assign different names to the same entity, which are then published under two different TF names.

The names are only assigned inside the tf publisher, and simon and I think we will only ever have one TF publsher per world (multiple ones will create flickering tf trees during state synchronization, and why would we need multiple anyways?)

> the tf thing would also be a big problem if we persist TF in neems, because the same world executed twice can create two completely different TF trees if we are unlucky with the ordering

this is still the case, but its gonna be very rare cases (most of the tf tree stays the same, only duplicates will change), and any solution will have tradeoffs and i think this is the smallest tradeoff. For learning we shouldnt depend of tf frames as features anyways, so who cares. Also the IDs are consistent within a neem, so it will create the same TF tree in those cases

> entity of another world could be added to this dict as well

here it cannot, because we get the entities from the world


semantic_digital_twin/src/semantic_digital_twin/adapters/ros/tf_publisher.py
```diff
@@ -30,6 +31,75 @@
...
+
+    def reserve_frames_of_other_publishers(self, frame_names: Iterable[str]) -> None:
+        """
+        Record the frames another publisher is already broadcasting.
+
+        :param frame_names: The frame names observed on the tf tree.
+        """
+        self.frames_of_other_publishers.update(frame_names)
+
+    def assign(self, entity: KinematicStructureEntity) -> str:
```
tomsch420, line 61: rst syntax of return

semantic_digital_twin/src/semantic_digital_twin/adapters/ros/tf_publisher.py
```diff
@@ -30,6 +31,75 @@
...
+            return self._frame_name_per_entity[entity.id]
+
+        frame_name = self._unused_frame_name_for(entity)
+        self._frame_name_per_entity[entity.id] = frame_name
+        self._assigned_frame_names.add(frame_name)
+        return frame_name
+
+    def _unused_frame_name_for(self, entity: KinematicStructureEntity) -> str:
+        """
+        Work out the frame name a not yet published entity should get.
```
tomsch420: rst syntax of return

tomsch420 (changes requested):
Very clean, very good

tomsch420 (approved):

semantic_digital_twin/src/semantic_digital_twin/adapters/ros/tf_publisher.py
```diff
@@ -235,10 +318,13 @@ def create_with_ignore_existing_tf(cls, world: World, node: Node) -> Self:
             for kse in world.kinematic_structure_entities
             if str(kse.name) in all_frames
         )
+        frame_names = TfFrameNames()
+        frame_names.reserve_frames_of_other_publishers(all_frames)
```
ichumuh: sus

ichumuh (changes requested):

ichumuh (approved):

Tigul (commented):
Doesn't this also need to be in the VizMarkerPublisher?

semantic_digital_twin/src/semantic_digital_twin/callbacks/callback.py
```diff
@@ -32,6 +35,27 @@ class Callback(WorldEntityWithClassBasedID, SubclassJSONSerializer, ABC):
     Flag that indicates if the callback is paused.
     """
 
+    @classmethod
+    @abstractmethod
+    def _callbacks_of_world(cls, world: World) -> List[Callback]:
```
tomsch420: the name is only partially describing what it does and at first it was confusing to me. I think it should be more precise here.

semantic_digital_twin/src/semantic_digital_twin/callbacks/callback.py
```diff
@@ -32,6 +35,27 @@ class Callback(WorldEntityWithClassBasedID, SubclassJSONSerializer, ABC):
...
+        """
+        :param world: The world whose callbacks are asked for.
+        :return: every callback the world notifies of the change this kind reacts to.
+        """
+        raise NotImplementedError
+
+    @classmethod
+    def all_of_world(cls, world: World) -> List[Self]:
```
tomsch420: same here

semantic_digital_twin/src/semantic_digital_twin/callbacks/callback.py
tomsch420: also if i see correctly these methods are only used once and i think its yagni

tomsch420 (changes requested):

semantic_digital_twin/src/semantic_digital_twin/callbacks/callback.py
LucaKro: used in synchronizer as well, but renamed

tomsch420 (approved):

ichumuh (approved):

PR #575, opened 2026-08-28
Add distribution_of/probability_of EQL constructs, reuse average() for expectations

tomsch420:
## Summary

Two new EQL query constructs for querying a probabilistic model directly instead of enumerating rows, both evaluable only under `ProbabilisticBackend`:

- **`distribution_of(match, *variables)`** — the probabilistic interpretation of `a()`/`an()`/`the()`: reuses a `Match`'s existing conditioning (literal-valued kwargs) and truncation (`.where(...)`) semantics, returning the resulting `ProbabilisticModel` directly instead of sampling instances from it. Optional `*variables` marginalize the result to a subset of the match's free variables.
- **`probability_of(condition)`** — the probability of any condition a `.where(...)` clause already accepts, resolved via `ProbabilisticModel.probability`.

No new construct for the expectation: `ProbabilisticBackend` now recognizes a *bare* `average(...)` selection (the existing aggregator, unmodified) and resolves it in closed form via `ProbabilisticModel.moment` instead of sampling and averaging rows — same call, correct answer under either backend.

Both new constructs are also verbalizable via `verbalize_expression`, e.g.:

```
The distribution over a Coin given that its c is 1.5, where its a is greater than 0.2
```

## Shared plumbing

- `UnderspecifiedParameters.resolve_conditioned_and_truncated_model`/`.apply_krrood_variable_truncation` extract the conditioning/truncation sequence `ProbabilisticBackend` already applied to a sampled `Match`, refactored out so `distribution_of` reuses it too.
- `ModelRegistry`'s parameter type (`ModelResolutionParameters`) is widened to also accept the new `SelectedAttributesParameters`/`ConditionParameters` shapes `probability_of`/`average` use.
- New `grammar/probabilistic_queries/rules.py` (auto-registered — no other verbalization wiring needed) adds `DistributionRule`/`ProbabilityRule`, reusing the match assembler's `given that`/`where` grammar under a new `Directive.DISTRIBUTION_OVER` header instead of the imperative `Find`/`Generate`.

## Test plan

- [x] New test suite `test/krrood_test/test_eql/test_probabilistic_queries/` (distribution, probability, average, native-evaluation-rejection, verbalization)
- [x] Full `test_eql/` + `test_feature_extraction/` regression suite green (1319 passed, 3 skipped)
- [x] Full `test_verbalization/` suite green, including the registry-completeness check and auto-discovered doctests on the new rules
- [x] Every doc example in `krrood/doc/eql/user/probabilistic_queries.md` verified by actually running it

### DIFF
--- a/krrood/doc/eql/user/probabilistic_queries.md
+++ b/krrood/doc/eql/user/probabilistic_queries.md
@@ -13,66 +13,47 @@ kernelspec:
 
 # Probabilistic Queries
 
-Sometimes a query should not return rows at all, but something *about* the
-distribution those rows would have come from -- the distribution itself, the
-probability of a condition, or the expectation of an attribute. `set_of`/`entity`
-always resolve to enumerated results; instead:
-
-| Read as... | Construct | Resolves to |
-|---|---|---|
-| "the distribution of ..." | {py:func}`~krrood.entity_query_language.factories.distribution_of` | a `ProbabilisticModel` |
-| "the probability of ..." | {py:func}`~krrood.entity_query_language.factories.probability_of` | a `float` |
-| "the average of ..." | {py:func}`~krrood.entity_query_language.factories.average` (already existing) | a `float` |
-
-Like `cause`/`confounder` (see {doc}`causality`), `distribution_of`/`probability_of`
-only mean anything under
-{py:class}`~krrood.entity_query_language.backends.ProbabilisticBackend`: querying a
-probabilistic model directly is a probabilistic operation, not a data selection, so
-neither has a native or SQL evaluation strategy at all -- evaluating either any other
-way raises
-{py:class}`~krrood.entity_query_language.exceptions.BackendCannotEvaluateProbabilisticQuery`.
-`average(...)` is different: it already has an ordinary native meaning (the mean of
-enumerated rows), and `ProbabilisticBackend` recognizes a *bare* `average(...)`
-selection and answers it in closed form instead -- same call, correct answer under
-either backend.
-
-## `distribution_of`: the probabilistic interpretation of `a`/`an`/`the`
-
-{py:func}`~krrood.entity_query_language.factories.a` already builds a
-{py:class}`~krrood.entity_query_language.query.match.Match` that
-{py:class}`~krrood.entity_query_language.backends.ProbabilisticBackend` conditions and
-truncates before *sampling* instances from it -- literal-valued kwargs condition the
-circuit, `.where(...)` conditions truncate it, and underspecified (`...`) fields are
-the free variables it samples.
-{py:func}`~krrood.entity_query_language.factories.distribution_of` asks for exactly
-that same conditioned-and-truncated model, just returned directly instead of sampled
-from:
+Mira runs quality control at a coin mint. Every coin that comes off the press gets
+measured by three sensors -- diameter, thickness, and weight -- and the mint has
+already fitted a probabilistic model to a long history of these measurements: given
+how the press was set up on a given run, what do the resulting coins tend to look
+like? Mira doesn't build that model herself; she just has questions for it.
 
 ```python
 from dataclasses import dataclass
 
+@dataclass
+class Coin:
+    a: float  # diameter
+    b: float  # thickness
+    c: float  # weight
+```
+
+`set_of`/`entity` answer questions about *actual coins* -- "find me the coins whose
+diameter is under spec". Mira's questions today are different: not about which coins
+exist, but about the *model itself*. Three constructs answer that kind of question:
+
+| Mira's question reads as... | Construct | Answer |
+|---|---|---|
+| "On average, how wide are the coins?" | {py:func}`~krrood.entity_query_language.factories.average` | a number |
+| "What fraction of coins are dangerously thin?" | {py:func}`~krrood.entity_query_language.factories.probability_of` | a number |
+| "Given today's press settings, what does the shape of the remaining measurements look like?" | {py:func}`~krrood.entity_query_language.factories.distribution_of` | a distribution |
+
+All three need the fitted model to answer, so they only work when evaluated with
+{py:class}`~krrood.entity_query_language.backends.ProbabilisticBackend`, the backend
+that knows how to reach it:
+
+```python
 from probabilistic_model.distributions.uniform import UniformDistribution
 from probabilistic_model.probabilistic_circuit.rx.probabilistic_circuit import (
     ProbabilisticCircuit, ProductUnit, leaf,
 )
 from random_events.interval import closed
-from random_events.product_algebra import SimpleEvent
 from random_events.variable import Continuous
 
-from krrood.entity_query_language.backends import ProbabilisticBackend
-from krrood.entity_query_language.factories import a, average, distribution_of, probability_of
-from krrood.parametrization.model_registries import DictRegistry
-
-
-@dataclass
-class Coin:
-    a: float
-    b: float
-    c: float
-
-
-# Named "Coin.<field>" up front, the same convention used in the causality example, so
-# DictRegistry can look each field up by its qualified name.
+# Stand-in for the mint's already-fitted model: today's press run, in isolation,
+# produces diameter/thickness/weight uniformly across a range each. (Named
+# "Coin.<field>" so DictRegistry -- see below -- can look each field up by name.)
 var_a = Continuous("Coin.a")
 var_b = Continuous("Coin.b")
 var_c = Continuous("Coin.c")
@@ -83,8 +64,60 @@ root.add_subcircuit(leaf(UniformDistribution(variable=var_a, interval=closed(0, 
 root.add_subcircuit(leaf(UniformDistribution(variable=var_b, interval=closed(0, 2).simple_sets[0]), circuit))
 root.add_subcircuit(leaf(UniformDistribution(variable=var_c, interval=closed(0, 3).simple_sets[0]), circuit))
 
-backend = ProbabilisticBackend(model_registry=DictRegistry({Coin: circuit}))
+from krrood.entity_query_language.backends import ProbabilisticBackend
+from krrood.entity_query_language.factories import a, average, distribution_of, probability_of, variable
+from krrood.parametrization.model_registries import DictRegistry
 
+backend = ProbabilisticBackend(model_registry=DictRegistry({Coin: circuit}))
+```
+
+## "On average, how wide are the coins?"
+
+The simplest question, and it uses a construct Mira already knows -- `average`, the
+same one that averages a column of matched rows. Called *bare* (no `set_of`, no
+`.grouped_by()`) against a model-backed backend, it answers straight from the model
+instead of averaging a sample of rows:
+
+```python
+x = variable(Coin)
+print(average(x.a).first(backend=backend))
+# 0.5
+```
+
+That's it -- same function Mira would use in an ordinary query, just pointed at a
+model instead of a table of coins.
+
+## "What fraction of coins are dangerously thin?"
+
+Suppose anything under 0.5 fails inspection.
+
+```python
+x = variable(Coin)
+print(probability_of(x.a < 0.5).first(backend=backend))
+# 0.5
+```
+
+There's a simpler way to think about this that Mira already knows from ordinary
+querying: take a big batch of coins, count how many have diameter under 0.5, divide by
+how many coins there are. That *is* what a probability means, and it's a perfectly
+valid way to estimate one. `probability_of` doesn't do that, though -- there's no batch
+of coins involved at all. It reads the answer directly off the fitted model, so
+instead of an estimate that gets better with a bigger batch, Mira gets the exact
+number the model implies, every time.
+
+## "Given today's press settings, what does the rest look like?"
+
+The press was calibrated to punch out 1.5g coins today, and Mira has already screened
+out anything under the 0.2 diameter minimum. She's not asking for a number now -- she
+wants to see the *shape* of what's left: how thickness and diameter vary once weight
+is pinned down and the too-narrow coins are gone.
+
+This is exactly the kind of thing `a(...)`/`an(...)`/`the(...)` already describe --
+Mira would normally write this same shape of match to *generate* example coins that
+fit these settings. `distribution_of` asks the identical question, just answered
+differently: instead of generating coins, it hands back the distribution itself.
+
+```python
 match = a(Coin)(a=..., b=..., c=1.5)
 match.where(match.variable.a > 0.2)
 
@@ -93,22 +126,24 @@ print(sorted(v.name for v in result.variables))
 # ['Coin.a', 'Coin.b', 'Coin.c']
 ```
 
-`c=1.5` conditions the circuit (`c` stays in the model, but as a point mass at 1.5 --
-conditioning doesn't drop the variable, it collapses its distribution); `a > 0.2`
-truncates it; `b` is free and untouched, since it's independent of `a`/`c`:
+`c=1.5` is the press setting; `.where(a > 0.2)` is the screening step; `a`/`b`
+(diameter/thickness) are left open (`...`) -- what Mira wants to see the shape of.
+The result is an ordinary distribution object, queryable like the mint's original
+model:
 
 ```python
+from random_events.product_algebra import SimpleEvent
+
 a_range = SimpleEvent.from_data({var_a: closed(0.2, 1)}).as_composite_set()
 print(result.probability(a_range))
-# 1.0 -- everything left satisfies a > 0.2
+# 1.0 -- everything left satisfies a > 0.2, by construction
 
 b_range = SimpleEvent.from_data({var_b: closed(0, 1)}).as_composite_set()
 print(result.probability(b_range))
-# 0.5 -- unchanged, b is independent of a and c
+# 0.5 -- unchanged: thickness never depended on diameter or weight to begin with
 ```
 
-Pass extra `*variables` to narrow the result to a subset of the match's free variables
-(further marginalization) -- e.g. just `a`, dropping `b` and the now-degenerate `c`:
+If Mira only cares about diameter specifically, she can ask for just that:
 
 ```python
 narrowed = distribution_of(match, match.variable.a).first(backend=backend)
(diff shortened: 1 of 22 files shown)


tomsch420:
Self-reviewed and pushed a fix-up commit (\`82d568223\`) addressing 7 findings from review:

**Correctness fixes:**
- \`_single_owner_class\` keyed model-registry lookups off \`Attribute._owner_class_\` (the immediate parent's class), not the attribute chain's root — a multi-hop path like \`x.arm.battery\` resolved to \`Arm\` instead of \`Robot\`, the class actually registered. Now uses \`_chain_root_._type_\`.
- \`RelationalCircuitRegistry.get_model\` crashed with a raw \`AttributeError\` when given \`probability_of\`/\`average\`'s lighter parameter classes (no match statement to ground). Now raises a clear \`RelationalCircuitRegistryRequiresMatch\`.
- \`average(x.a, distinct=True)\` under \`ProbabilisticBackend\` silently ignored \`distinct=True\` and returned the plain closed-form expectation. Now falls through to native evaluation instead.
- \`verbalize_expression(probability_of(True))\` crashed with an unrelated \`AttributeError\` (a bare bool has nothing to scan for referring expressions). Now raises a clear \`ProbabilityConditionNotVerbalizable\`.

**Cleanup:**
- \`_resolve_average\` now calls the existing \`ProbabilisticModel.expectation()\` helper instead of hand-rebuilding the same moment call.
- Several docstrings still named pre-rename/removed constructs (\`marginal(...)\`, \`moment(...)\`, \`distribution(...)\`) instead of the actual API (\`distribution_of\`, \`probability_of\`, \`average\`).

New tests: \`test_probabilistic_queries/test_bug_fixes.py\`. Full regression (\`test_eql/\` + \`test_feature_extraction/\`) still green: 1323 passed, 3 skipped.

krrood/src/krrood/entity_query_language/exceptions.py
```diff
@@ -934,6 +938,61 @@ def suggest_correction(self) -> str:
...
+            f"{self.expression} cannot be evaluated natively: querying a probabilistic "
+            f"model directly is a probabilistic operation, not a data selection."
+        )
+
+    def suggest_correction(self) -> str:
+        return "Evaluate with a ProbabilisticBackend backed by a model registry, e.g. .evaluate(backend=ProbabilisticBackend(...))."
+
+
+@dataclass
+class ProbabilityConditionNotVerbalizable(DataclassException):
```
tomsch420: when does this happen?

krrood/src/krrood/parametrization/model_registries.py
```diff
@@ -13,16 +18,29 @@
 from probabilistic_model.probabilistic_circuit.rx.helper import fully_factorized
 from probabilistic_model.probabilistic_model import ProbabilisticModel
 
+ModelResolutionParameters = Union[
```
tomsch420: Do they not have a shared superclass that can be used?

krrood/src/krrood/parametrization/parameterizer.py
```diff
@@ -585,3 +658,111 @@ def _process_attribute_match_type(type_):
             return most_appropriate_variable_type(types)
         else:
             return type_
+
+
+def _single_owner_class(attributes: Iterable[Attribute]) -> Type:
```
tomsch420: this method should belong to some class and not be orphaned like this. For instance attribute?

krrood/src/krrood/entity_query_language/verbalization/vocabulary/english.py
```diff
@@ -209,16 +209,21 @@ class Keywords(VocabEnum):
...
+    The opening phrase of a request: *"Find"* a match in the domain, *"Generate"* an
+    underspecified one, or *"The distribution over"* one whose free variables'
+    distribution is wanted instead of a concrete match -- the two imperatives ask for
+    rows, the third asks for a description of the query, so it takes the same slot
+    without being one itself.
     """
 
     FIND = KeyWord("Find")
     GENERATE = KeyWord("Generate")
+    DISTRIBUTION_OVER = KeyWord("The distribution over")
```
tomsch420, line 226: will this be keyworded also by color?

krrood/doc/eql/user/probabilistic_queries.md
```diff
@@ -0,0 +1,235 @@
...
+|---|---|---|
+| "the distribution of ..." | {py:func}`~krrood.entity_query_language.factories.distribution_of` | a `ProbabilisticModel` |
+| "the probability of ..." | {py:func}`~krrood.entity_query_language.factories.probability_of` | a `float` |
+| "the average of ..." | {py:func}`~krrood.entity_query_language.factories.average` (already existing) | a `float` |
+
+Like `cause`/`confounder` (see {doc}`causality`), `distribution_of`/`probability_of`
+only mean anything under
+{py:class}`~krrood.entity_query_language.backends.ProbabilisticBackend`: querying a
+probabilistic model directly is a probabilistic operation, not a data selection, so
+neither has a native or SQL evaluation strategy at all -- evaluating either any other
```
tomsch420: theoretically that is evaluatable by dividing the number of rows where probability_of(...) holds by the number of total rows

krrood/doc/eql/user/probabilistic_queries.md
tomsch420, line 1: i think this file should be more about teaching what the things mean in a more relatable way with less implementation details. Cook up a story, and use the construct in the story to solve some problems

krrood/src/krrood/entity_query_language/exceptions.py
tomsch420: Good question — I dug into it and it turned out to reveal a real gap. It only happens if someone writes `probability_of(True)`/`probability_of(False)` literally, bypassing `and_`/`or_`/`not_`/comparators entirely (those already normalize a bare bool into a `Literal`, so they never produce a plain `bool` here). Rather than keep a dedicated exception for that, I normalize `Probability.condition` into a `Literal` on construction too, in `82d568223`/`d62a1fa2c` — same treatment `and_`/`or_`/`not_` already give a bare bool. That means `probability_of(True)` now verbalizes cleanly ("the probability that True") instead of crashing. It still can't *evaluate*, though — a content-free condition names no class, so there's no model to resolve it against — so that now raises the existing `JointQueryAcrossClassesNotSupported`, whose message I extended to cover the "zero classes" case explicitly instead of just "more than one".

krrood/src/krrood/parametrization/model_registries.py
tomsch420: You're right, they should. Added `ModelQueryParameters`, a shared ABC (`variables`/`queried_class` abstract properties) that `UnderspecifiedParameters`/`SelectedAttributesParameters`/`ConditionParameters` all now inherit, and `ModelRegistry.get_model` types against that directly instead of the `Union`. In `d62a1fa2c`.

krrood/src/krrood/parametrization/parameterizer.py
tomsch420: Agreed it needed a home. I put it on `ModelQueryParameters` (the new shared base from the comment below) as a staticmethod instead of `Attribute` — it operates on a whole collection of attributes to find their single common owner class, which doesn't fit as an instance method on one `Attribute`, but fits naturally as shared logic the two parameter classes that need it (`SelectedAttributesParameters`, `ConditionParameters`) both call via `self._single_owner_class(...)`. In `d62a1fa2c`.

krrood/src/krrood/entity_query_language/verbalization/vocabulary/english.py
tomsch420: Checked by actually rendering it: yes, automatically. Coloring is keyed by `SemanticRole` (`ROLE_COLORS` in `rendering/formatter.py`), not by the specific word — every `KeyWord`/`VocabEnum` member (`FIND`, `GENERATE`, `WHERE`, ...) gets tagged `SemanticRole.KEYWORD` by `as_fragment()` and colored the same yellow (`#eded18`) regardless of which one it is. `Directive.DISTRIBUTION_OVER`/`Keywords.THE_PROBABILITY_THAT` use the exact same `KeyWord` class, so they're covered for free — confirmed with `VerbalizationPipeline.html()`, e.g. `<span style="color:#eded18">The distribution over</span>` in the actual rendered output.

krrood/doc/eql/user/probabilistic_queries.md
tomsch420: Fair point, that overstated it. Rewrote that section: `probability_of` computes the exact value a "count matching rows / count all rows" estimate is aiming for, reading it directly off the model instead of estimating it from a sample. Folded into the doc rewrite from the comment below (`d62a1fa2c`).

krrood/doc/eql/user/probabilistic_queries.md
tomsch420: Rewrote the doc around a coin mint's quality engineer (Mira) working through three real questions — average diameter, fraction of thin coins, and the shape of what's left after a press-setting/screening condition — using `average`/`probability_of`/`distribution_of` to answer each, with the implementation mechanics pushed to a short recap + API reference at the end instead of being the focus. In `d62a1fa2c`; every code block re-verified by actually running it.

krrood/doc/eql/user/probabilistic_queries.md
tomsch420: Can you make probability of work for every backend using the counting strategy you just said?

krrood/doc/eql/user/probabilistic_queries.md
tomsch420: I think the story should be robotics based, as in the other tutorials of KRROOD

krrood/doc/eql/user/probabilistic_queries.md
tomsch420: Fair — implemented it. `Probability` now has a real `_evaluate_natively_`: it counts matching rows via `entity(count(root)).where(condition)` (the exact pattern `test_aggregations.py::test_count` already exercises), so `probability_of(...)` now works under any backend — e.g. plain `EntityQueryLanguageBackend`, no `ProbabilisticBackend` needed — as long as the referenced variable has an enumerable domain to count over. The closed-form path via `ProbabilisticBackend` stays exact and unchanged; this just adds the counting path alongside it, picked automatically by whichever backend is given. `distribution_of` stays `ProbabilisticBackend`-only — there's no row-counting equivalent for a whole distribution. Added to the doc and to `test_probability.py` (counting, explicit backend, conjunction, and the content-free-condition failure case). In `fb3438736`.

krrood/doc/eql/user/probabilistic_queries.md
tomsch420: Agreed — rewrote it around a robotics story to match the other KRROOD tutorials: Nova operates a warehouse robot fleet, and instead of a Coin with a/b/c fields there's a `Robot` dataclass (`temperature`, `speed`, `charge`), with the same three questions (`average` for fleet speed, `probability_of` for how many robots run dangerously hot, `distribution_of` for the shape of readings given today's charge level and a speed screen). All code blocks re-verified by actually running them end to end. In `fb3438736`.

krrood/doc/eql/user/probabilistic_queries.md
```diff
@@ -0,0 +1,205 @@
...
+
+temperature_range = SimpleEvent.from_data({var_temperature: closed(0, 0.5)}).as_composite_set()
+print(result.probability(temperature_range))
+# 0.5 -- unchanged: temperature never depended on speed or charge to begin with
+```
+
+To narrow down to temperature specifically:
+
+```python
+narrowed = distribution_of(match, match.variable.temperature).first(backend=backend)
```
AbdelrhmanBassiouny: why not make the second argument a keyword argument so we know what you are setting here?

krrood/src/krrood/entity_query_language/operators/probabilistic_queries.py
```diff
@@ -0,0 +1,225 @@
...
+        rather than reimplementing counting itself. Unlike
+        :meth:`_resolve_`'s exact, closed-form answer, this is only as good as the
+        variable's domain (all of it must be enumerable) -- it's the estimate
+        :meth:`_resolve_` computes exactly instead, when a
+        :class:`~krrood.entity_query_language.backends.ProbabilisticBackend` is used.
+
+        :raises JointQueryAcrossClassesNotSupported: If the condition references
+            attributes reached from more than one ``variable(...)`` root, or none.
+        """
+        from krrood.entity_query_language.factories import count, entity
```
AbdelrhmanBassiouny: if these are to avoid cyclic import, add a comment here on that such that we know

krrood/src/krrood/entity_query_language/operators/probabilistic_queries.py
```diff
@@ -0,0 +1,225 @@
...
+        Resolve this query against a model obtained from ``model_registry``.
+
+        :param model_registry: The registry to resolve a model with.
+        :return: Whatever the underlying ``ProbabilisticModel`` operation naturally
+            returns -- a plain value, unwrapped.
+        """
+
+
+@dataclass(eq=False, repr=False)
+class Probability(ProbabilisticQuery):
```
AbdelrhmanBassiouny, line 81: the doc string is sooo big, summarize and simplify it please, this verposity can be in the doc files instead.

krrood/src/krrood/entity_query_language/operators/probabilistic_queries.py
```diff
@@ -0,0 +1,225 @@
...
+    condition: ConditionType
+    """
+    The condition to compute the probability of.
+    """
+
+    def __post_init__(self):
+        if not isinstance(self.condition, SymbolicExpression):
+            self.condition = Literal(_value_=self.condition)
+
+    def _evaluate_natively_(self) -> Iterator[float]:
```
AbdelrhmanBassiouny, line 101: same very big doc string comment

krrood/src/krrood/entity_query_language/operators/probabilistic_queries.py
```diff
@@ -0,0 +1,225 @@
...
+            self.condition
+        ).variables.keys()
+        roots = {attribute._chain_root_ for attribute in referenced_attributes}
+        if len(roots) != 1:
+            raise JointQueryAcrossClassesNotSupported(
+                {root._type_ for root in roots}
+            )
+        [root_variable] = roots
+
+        matching_count = entity(count(root_variable)).where(self.condition).first()
```
AbdelrhmanBassiouny, line 119: why not use python list comprehension directly here instead of eql?

krrood/src/krrood/entity_query_language/verbalization/grammar/probabilistic_queries/rules.py
```diff
@@ -0,0 +1,139 @@
...
+    >>> verbalize_expression(distribution_of(match))
+    "The distribution over a Robot given that its name is 'R2'"
+    >>> verbalize_expression(distribution_of(match, match.variable.battery))
+    "The distribution over the battery of a Robot given that its name is 'R2'"
+    """
+
+    construct = Distribution
+
+    def build(self, node: Distribution, context: RuleContext) -> VerbalizationFragment:
+        context.services.performative_override = Directive.DISTRIBUTION_OVER
```
AbdelrhmanBassiouny, line 80: what do you think about making Distribution am EQL Backend, then actially you can there say the Directive of that backend as a property of the Backend, also it fits how you use distribution and propability_of now where they take a query basically and evaluate it.

krrood/src/krrood/entity_query_language/verbalization/pipeline.py
AbdelrhmanBassiouny: match...case...?

krrood/src/krrood/entity_query_language/verbalization/verbalizer.py
```diff
@@ -59,10 +63,19 @@ def build(
...
-        # resolved query expression.
-        scan_target = (
-            expression.expression if isinstance(expression, Match) else expression
-        )
+        # resolved query expression. Distribution/Probability are Evaluable-only (not
+        # SymbolicExpression), so the referring-expression scan needs the SymbolicExpression they
+        # actually wrap instead -- the underlying match's resolved query, or the condition
+        # (Probability.__post_init__ already normalizes a bare bool into a Literal, so
+        # .condition is always a real SymbolicExpression here).
+        if isinstance(expression, Match):
```
AbdelrhmanBassiouny: ok it is starting to feel like we have an open/closed problem here. Do you think we can make a mixin super class that is called HasExpression and adds a _get_expression_ abstract method and all those classes implement and thus we do not need this if conditions here and in the previous file where I suggested doin match...case...

krrood/src/krrood/entity_query_language/backends.py
```diff
@@ -307,6 +313,67 @@ class ProbabilisticBackend(GenerativeBackend):
...
+        aggregator = expression.selected_aggregator
+        if not isinstance(aggregator, Average):
+            return None
+        if aggregator._leaf_attribute_ is None:
+            return None
+        if aggregator._distinct_:
+            # native evaluation deduplicates values before averaging; the closed-form
+            # expectation has no notion of that, so it must not silently stand in
+            return None
+        if any(
```
AbdelrhmanBassiouny, line 350: Can the backend override the evaluate of symbolic expressions? If so, do you think it would be better to do that and raise in the expressions this cannot handle instead of doing these if conditions here?

AbdelrhmanBassiouny (changes requested):

krrood/doc/eql/user/probabilistic_queries.md
tomsch420: agreed, claude make it tuple of eql variable and name it marginalize_for

krrood/src/krrood/entity_query_language/operators/probabilistic_queries.py
tomsch420: agreed. Claude write a short factual description with less prose

krrood/src/krrood/entity_query_language/operators/probabilistic_queries.py
tomsch420: agreed. Claude write a short factual description with less prose

krrood/src/krrood/entity_query_language/operators/probabilistic_queries.py
```diff
@@ -0,0 +1,225 @@
...
+    ``distribution_of(match, match.variable.outcome)``. Without them, every one of the
+    match's free variables is kept.
+    """
+
+    match: Match
+    """
+    The match whose conditions describe the distribution.
+    """
+
+    variables: Tuple[Attribute, ...] = ()
```
tomsch420: make this a default factory

krrood/src/krrood/entity_query_language/operators/probabilistic_queries.py
tomsch420: is there any benefit of doing so?

krrood/src/krrood/entity_query_language/verbalization/verbalizer.py
tomsch420: i think we can, claude do it!

krrood/src/krrood/entity_query_language/verbalization/pipeline.py
tomsch420: ok

krrood/src/krrood/entity_query_language/backends.py
tomsch420: as discussed in person, nothing needs to be done here

krrood/doc/eql/user/probabilistic_queries.md
tomsch420: Agreed and done. `distribution_of`'s second argument is now keyword-only: `marginalize_for: Tuple[Attribute, ...]`, so a call site reads `distribution_of(match, marginalize_for=(match.variable.temperature,))` instead of a bare positional. `Distribution.marginalize_for` mirrors it, with a `field(default_factory=tuple)` default per the note below. In `c3cce99d0`.

krrood/src/krrood/entity_query_language/operators/probabilistic_queries.py
tomsch420: Right, they are -- added a comment. The module-level comment (right above ProbabilisticQuery) now explains the root cause for all of them: krrood.entity_query_language.factories imports Distribution/Probability from this module, so anything imported here at module level that reaches back to factories -- directly (factories.count/entity, used at this line) or indirectly (krrood.parametrization.parameterizer, which itself imports from factories) -- would cycle. Each individual local-import site now also has a one-line pointer back to that explanation. In `c3cce99d0`.

krrood/src/krrood/entity_query_language/operators/probabilistic_queries.py
tomsch420: Trimmed it down to a short factual description, with a pointer to the user doc instead of restating the whole design rationale inline. In `c3cce99d0`.

krrood/src/krrood/entity_query_language/operators/probabilistic_queries.py
tomsch420: Same treatment -- shortened to a short factual description (what it does and how it's counted), with the design rationale left to the user doc instead of the docstring. In `c3cce99d0`.

krrood/src/krrood/entity_query_language/verbalization/pipeline.py
tomsch420: Superseded by the mixin from the comment below (verbalizer.py:71) -- added a `HasExpression` ABC with an abstract `_get_expression_()` that `SymbolicExpression`/`Match`/`Distribution`/`Probability` all implement, so this whole isinstance chain collapses to `target = expression._get_expression_()` (only the `isinstance(target, Query)` build-guard remains, which is a different, orthogonal check -- whether the thing has anything to build, not which wrapper type routed here). No match/case needed since there's nothing left to dispatch on. In `c3cce99d0`.

krrood/src/krrood/entity_query_language/verbalization/verbalizer.py
tomsch420: Done. Added `HasExpression(ABC)` to core/base_expressions.py with an abstract `_get_expression_() -> SymbolicExpression`. `SymbolicExpression` implements it as `return self` (covers `Query` and any plain expression node for free), `Match` as `return self.expression`, and `Distribution`/`Probability` (both already required to implement `_resolve_` on the shared `ProbabilisticQuery` base, so `_get_expression_` joins it the same way) as `return self.match._get_expression_()` / `return self.condition`. Both this file's scan_target and pipeline.py's build-target now just call `expression._get_expression_()` -- no isinstance chain in either place, and a future `HasExpression` implementer needs no changes to this file or pipeline.py at all. In `c3cce99d0`.

krrood/src/krrood/entity_query_language/operators/probabilistic_queries.py
tomsch420: Done -- `marginalize_for: Tuple[Attribute, ...] = field(default_factory=tuple)`. In `c3cce99d0`.

krrood/src/krrood/entity_query_language/operators/probabilistic_queries.py
tomsch420: im still unhappy here. The exception should just import this using if type checking, as the exception never constructs an instance of Probability(). Check if its also possible for the WhereExpressionToRandomEventsTranslator to only depend on this using type checking

tomsch420: You were right -- both were fixable at the root. `krrood.parametrization.exceptions` and `krrood.parametrization.random_events_translator` each imported `factories` at module level purely for the `ConditionType` annotation on a dataclass field; neither ever constructs a `Probability`/`Distribution`, so that import only exists for type checkers. Moved both under `TYPE_CHECKING` (adding `from __future__ import annotations` to exceptions.py, which random_events_translator.py already had), so neither reaches back to `factories.py` at runtime any more.

That let me hoist `JointQueryAcrossClassesNotSupported`/`WhereExpressionToRandomEventTranslator` out of `_evaluate_natively_` into normal top-level imports here. Only two local imports are left, and both are genuine runtime dependencies that can't be fixed the same way: `factories.count/entity` itself (this module IS what factories.py imports Distribution/Probability from, so that's an inherent two-way coupling), and `krrood.parametrization.parameterizer`, which imports `factories.and_` for real, not just a type hint.

Verified with a fresh-process import test that all four modules import cleanly regardless of which one loads first, and full regression is still green (1325 passed, 3 skipped). In `ea24a9d88`.

krrood/src/krrood/parametrization/exceptions.py
```diff
@@ -1,12 +1,21 @@
...
-from typing_extensions import Any, List, Set, Type
+from typing_extensions import Any, List, Set, TYPE_CHECKING, Type
 
 import random_events.variable
 from krrood.entity_query_language.core.variable import Variable
-from krrood.entity_query_language.factories import ConditionType
 from krrood.exceptions import DataclassException, InputError
 
+if TYPE_CHECKING:
+    # None of these exceptions construct a Probability/Distribution -- ConditionType is
```
tomsch420: remove these big comments

krrood/src/krrood/parametrization/random_events_translator.py
```diff
@@ -30,6 +29,14 @@
 from random_events.interval import closed_open, closed, open
 from random_events.product_algebra import Event, SimpleEvent
 
+if TYPE_CHECKING:
+    # ConditionType is only ever a dataclass field annotation here (see
+    # WhereExpressionToRandomEventTranslator.conditions_root below) -- never
```
tomsch420: remove these big comments

krrood/src/krrood/entity_query_language/operators/probabilistic_queries.py
```diff
@@ -43,15 +47,18 @@
...
-# Every local import below (inside a method body, not at module level) exists to avoid
-# a circular import: krrood.entity_query_language.factories imports Distribution/
-# Probability from this module, so anything imported here at module level that reaches
-# back to factories -- directly (factories.count/entity) or indirectly
-# (krrood.parametrization.parameterizer, which itself imports from factories) -- would
-# cycle. Deferring those imports into each method body instead (matching
-# operators/causal.py's existing leaf-module pattern) delays them until a backend
-# actually evaluates the query, by which point every module involved has finished
-# loading.
+# krrood.entity_query_language.factories imports Distribution/Probability from this
```
tomsch420: remove these big comments

krrood/src/krrood/parametrization/exceptions.py
tomsch420: Done. In `1e90a3df9`.

krrood/src/krrood/parametrization/random_events_translator.py
tomsch420: Done. In `1e90a3df9`.

krrood/src/krrood/entity_query_language/operators/probabilistic_queries.py
tomsch420: Done -- removed the module-level block and cut the three local-import pointer comments down to one line each. In `1e90a3df9`.

krrood/src/krrood/entity_query_language/operators/probabilistic_queries.py
```diff
@@ -160,12 +122,13 @@ def _evaluate_natively_(self) -> Iterator[float]:
...
 
     def _resolve_(self, model_registry: ModelRegistry) -> float:
-        from krrood.parametrization.parameterizer import ConditionParameters
-
         parameters = ConditionParameters(self.condition)
         model = model_registry.get_model(parameters)
         return model.probability(parameters.event)
 
+    def _get_expression_(self) -> SymbolicExpression:
+        return self.condition
```
AbdelrhmanBassiouny, line 129: no need for that, it will already return self as the default implementation from SymbolicExpression.

AbdelrhmanBassiouny (changes requested):

krrood/src/krrood/entity_query_language/operators/probabilistic_queries.py
tomsch420: agrre

tomsch420: Checked this before applying it -- it doesn't hold here, and removing the override actually breaks things. `Probability` isn't a `SymbolicExpression` subclass, only `ProbabilisticQuery(Evaluable, HasExpression, ABC)`, so there's no `SymbolicExpression._get_expression_` default anywhere in its MRO to fall back on. I tested removing it directly: `Probability(condition=...)` then raises `TypeError: Can't instantiate abstract class Probability without an implementation for abstract method '_get_expression_'`, since `HasExpression._get_expression_` stays abstract with nothing concrete implementing it.

What *is* true, and maybe what you're pointing at: `self.condition` itself is always a real `SymbolicExpression` (guaranteed by `__post_init__`'s bool->Literal normalization), so `self.condition._get_expression_()` already resolves to `self.condition` via `SymbolicExpression`'s own default (`return self`) -- no override needed *on the condition*. But something still has to route a `Probability` instance itself to that condition, since `Probability` is a separate wrapper object, not a `SymbolicExpression`. I left `Probability._get_expression_` as `return self.condition` (kept as is, since it's already correct and equivalent) -- let me know if you meant something else and I'm missing it.

AbdelrhmanBassiouny (approved):

krrood/src/krrood/entity_query_language/operators/probabilistic_queries.py
```diff
@@ -0,0 +1,195 @@
...
+    See :doc:`/krrood/doc/eql/user/probabilistic_queries` for the full walkthrough.
+    """
+
+    condition: ConditionType
+    """
+    The condition to compute the probability of.
+    """
+
+    def __post_init__(self):
+        if not isinstance(self.condition, SymbolicExpression):
```
LucaKro, line 98: why not just type for SymbolicExpression

LucaKro (approved):

krrood/src/krrood/entity_query_language/operators/probabilistic_queries.py
tomsch420: makes sense actually. Claude!

tomsch420: Narrowed the field to `SymbolicExpression` (`condition: SymbolicExpression` instead of `ConditionType`) and dropped the now-unused `ConditionType` import — `__post_init__` already guarantees it's a `SymbolicExpression` by construction time, so this reflects the actual invariant. `394b47e2b`. Full regression green (1325 passed, 3 skipped).

PR #576, opened 2026-08-28
Add `combined_mesh` property to merge collision geometries in semantic annotations; implement related tests.

LucaKro:
not used outside of tests yet, will be used by @Tigul in his current PR

### DIFF
--- a/semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
+++ b/semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
@@ -6,6 +6,7 @@ from typing import Tuple
 
 import numpy as np
 import trimesh
+from trimesh.util import concatenate
 from krrood.class_diagrams.class_diagram import WrappedClass
 from krrood.entity_query_language.factories import variable_from, entity, variable, an
 from krrood.ormatic.utils import classproperty
@@ -140,6 +141,23 @@ class HasRootKinematicStructureEntity(
     """
     The root kinematic structure entity of the semantic annotation.
     """
+
+    @property
+    def combined_mesh(self) -> trimesh.Trimesh:
+        """
+        The collision geometry of every body of this annotation, merged into a single
+        mesh expressed in the frame of :attr:`root`.
+
+        ..note:: Rebuilt on every access, since the bodies move relative to each other
+            with the world state.
+        """
+        return concatenate(
+            [
+                shape.mesh_in_frame(self.root)
+                for body in self.bodies_with_collision
+                for shape in body.collision
+            ]
+        )
 
     @property
     def scale(self) -> Scale:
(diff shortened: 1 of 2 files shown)


Tigul (commented):
Would be cool to move this to world as combined_mesh_of_branch

LucaKro:
> Would be cool to move this to world as combined_mesh_of_branch

I think this is the correct place for it, inside of has a semantic meaning, so i think it makes a lot of sense that it will take semantic annotations (or for now HasRootKinematicStructureEntities) as an input. Whenever you  would ask the question if a collection of bodies (like the branch) is inside another Set, both sides will have a semantic meaning attached to it

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
```diff
@@ -141,6 +142,23 @@ class HasRootKinematicStructureEntity(
...
+        :return: The collision geometry of every body of this annotation, merged into a single
+        mesh expressed in the frame of :attr:`root`.
+
+        ..note:: Rebuilt on every access, since the bodies move relative to each other
+            with the world state.
+        """
+        return concatenate(
+            [
+                shape.mesh_in_frame(self.root)
+                for body in self.bodies_with_collision
```
tomsch420, line 157: Does this traverse the kinematic structure or the semantic structure?

tomsch420 (changes requested):

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
LucaKro: it basically traverses nested semantic annotations and collects their kse, then filters to bodies and then filters to bodies with collision

tomsch420 (approved):

Tigul (approved):

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
```diff
@@ -149,6 +142,23 @@ class HasRootKinematicStructureEntity(
...
+    @property
+    def combined_mesh(self) -> trimesh.Trimesh:
+        """
+        :return: The collision geometry of every body of this annotation, merged into a single
+        mesh expressed in the frame of :attr:`root`.
+
+        ..note:: Rebuilt on every access, since the bodies move relative to each other
+            with the world state.
+        """
+        return concatenate(
```
ichumuh, line 154: concatenate is something i usually now from numpy so i was confused until i saw the import. i wonder if that's something claude could also stumble over and writing trimesh.concatenate would be better. idk your call

ichumuh (approved):

semantic_digital_twin/src/semantic_digital_twin/semantic_annotations/mixins.py
LucaKro: i think its fine, whether its not something thats visible from the outside, and i dont think the the method will change a lot. also this was also used in ShapeCollection.combined_mesh, and so far it didnt really cause any problems ^^

PR #578, opened 2026-08-31
Speed up NygaInduction: vectorize split search, fix O(leaves^2) root lookup

tomsch420:
## Summary
- `InductionStep.compute_best_split` scored every candidate split with a per-candidate Python loop (two calls to `log_likelihood_of_split_side` each). Replaced with one vectorized numpy pass over all candidates at once — same formula, bit-identical output (factored into a shared `_vectorized_log_likelihood_of_split_side` helper for the left/right sides). The now-dead scalar `log_likelihood_of_split_side` was removed; its notebook-matched reference-value test now calls the vectorized helper directly.
- `InductionStep.induce` attached each new leaf via `probabilistic_circuit.root.add_subcircuit(...)`. The `.root` property rescans every node in the graph whenever its cache is stale, and every leaf attach (a new edge) invalidates that cache — so every fit was O(leaves²) regardless of data or hyperparameters. `NygaInduction` now caches a direct reference to the root `SumUnit` (`root_unit`, set once in `fit()`), and `induce()` uses that instead.
- `SumUnit.forward`/`log_forward` built a full Python list of every subcircuit's weighted result before reducing it with `np.sum`/`logsumexp`. For array-valued queries (e.g. one log-likelihood per sample), that list is an implicit `(n_subcircuits, n_samples)` matrix, materialized several times over (the list itself, `logsumexp`'s internal `np.asarray` copy, its subtraction/exp buffers) on top of the memory the subcircuits' own results already occupy. Replaced both with a streaming accumulation (plain running sum, `np.logaddexp`) that folds in one subcircuit at a time — same math, same result, and general to any `SumUnit`, not specific to Nyga's deterministic partitions.

## Benchmarks

All numbers below are freshly measured on the same machine: `cram2/main` (merge-base `8c40d6f03`) as "before" vs. this branch's tip as "after", same synthetic 5-mode Gaussian mixture data and same seeds for both. Leaf counts are identical between before/after in every row, confirming the vectorization is bit-for-bit equivalent — only the timing changes.

**Fit time** (`NygaInduction.fit`, vectorization + O(leaves²) fix):

| n | min_samples_per_quantile | min_likelihood_improvement | leaves | before | after | speedup |
|---:|---:|---:|---:|---:|---:|---:|
| 1,000 | 10 | 0.01 | 59 | 0.097s | 0.005s | 18.6x |
| 10,000 | 10 | 0.01 | 631 | 2.042s | 0.051s | 40.1x |
| 50,000 | 10 | 0.01 | 3,162 | 14.629s | 0.319s | 45.9x |
| 100,000 | 10 | 0.01 | 6,315 | 33.328s | 0.775s | 43.0x |
| 1,000 | 2 | 0.01 | 354 | 0.157s | 0.024s | 6.6x |
| 5,000 | 2 | 0.01 | 1,756 | 1.479s | 0.134s | 11.0x |
| 10,000 | 2 | 0.01 | 3,497 | 3.088s | 0.312s | 9.9x |

**Peak memory during `circuit.log_likelihood(...)`** (SumUnit streaming fix) — these are the configs that hit a 12GB memory cap in benchmarking before this fix; fitting itself was already cheap (e.g. the first row fits in 82MB/0.8s), the blowup was entirely in evaluating the fitted circuit:

| n_train | min_samples_per_quantile | min_likelihood_improvement | leaves | before | after |
|---:|---:|---:|---:|---:|---:|
| 80,000 | 10 | 0.01 | 6,315 | exceeds 12GB cap | 3.9GB, 12.3s |
| 80,000 | 10 | 0.001 | 6,298 | exceeds 12GB cap | 4.9GB, 12.0s |
| 40,000 | 10 | 0.05 | 3,170 | exceeds 12GB cap | 4.9GB, 2.6s |
| 80,000 | 10 | 0.05 | 6,305 | exceeds 12GB cap | 5.1GB, 11.1s |

The new peak sits right at the theoretical floor of `leaves * samples * 8 bytes` (e.g. 6,315 × 80,000 × 8B ≈ 4.0GB), instead of several times that.

## Test plan
- [x] Full `probabilistic_model_test` suite (excluding an unrelated pre-existing Qt/GUI test crash): 389 passed, 1 skipped, on every commit in this PR.
- [x] Re-ran the size/hyperparameter benchmark grid before and after each change; leaf counts and log-likelihoods match exactly, only `fit_time_s` / peak memory change.

### DIFF
--- a/probabilistic_model/src/probabilistic_model/learning/nyga_induction.py
+++ b/probabilistic_model/src/probabilistic_model/learning/nyga_induction.py
@@ -218,38 +218,70 @@ class InductionStep:
         The best split of the data is computed by evaluating the log likelihood of every possible split and memorizing
         the best one.
 
+        This evaluates every candidate split at once with vectorized numpy operations
+        (mirroring what two calls to :meth:`log_likelihood_of_split_side` per candidate
+        would compute) instead of a per-candidate Python loop, since that loop is the
+        dominant cost of :meth:`NygaInduction.fit` on large datasets.
+
         :return: The maximum log likelihood and the best split index.
         """
 
-        # initialize the maximum likelihood and the best split index
-        maximum_log_likelihood = -float("inf")
-        best_split_index = None
+        first_split_index = self.begin_index + self.min_samples_per_quantile
+        last_split_index = self.end_index - self.min_samples_per_quantile + 1
 
-        # calculate the connecting points
-        right_connecting_point = self.right_connecting_point()
+        # no candidate split respects min_samples_per_quantile on both sides
+        if first_split_index >= last_split_index:
+            return -float("inf"), None
+
+        split_indices = np.arange(first_split_index, last_split_index)
+        split_values = (
+            self.data[split_indices - 1] + self.data[split_indices]
+        ) / 2
+
         left_connecting_point = self.left_connecting_point()
+        right_connecting_point = self.right_connecting_point()
+        log_weight_sum = np.log(self.total_weights)
 
-        # for every possible splitting index
-        for split_index in range(
-            self.begin_index + self.min_samples_per_quantile,
-            self.end_index - self.min_samples_per_quantile + 1,
-        ):
+        # left-hand side: same formula as log_likelihood_of_split_side(split_index,
+        # left_connecting_point), which always resolves to the "is_left" branch since
+        # every split_value lies to the right of left_connecting_point
+        number_of_samples_left = split_indices - self.begin_index
+        log_density_left = np.log(split_values - left_connecting_point)
+        log_weight_sum_of_split_left = np.log(
+            self.cumulative_weights[split_indices]
+            - self.cumulative_weights[self.begin_index]
+        )
+        sum_of_log_weights_left = (
+            self.cumulative_log_weights[split_indices]
+            - self.cumulative_log_weights[self.begin_index]
+        )
+        left_hand_side = (
+            number_of_samples_left
+            * (log_weight_sum_of_split_left - log_weight_sum - log_density_left)
+            + sum_of_log_weights_left
+        )
 
-            # calculate log likelihoods
-            log_likelihood_left = self.log_likelihood_of_split_side(
-                split_index, left_connecting_point
-            )
-            log_likelihood_right = self.log_likelihood_of_split_side(
-                split_index, right_connecting_point
-            )
-            log_likelihood = log_likelihood_left + log_likelihood_right
+        # right-hand side: same formula as log_likelihood_of_split_side(split_index,
+        # right_connecting_point), which always resolves to the "is_right" branch
+        number_of_samples_right = self.end_index - split_indices
+        log_density_right = np.log(right_connecting_point - split_values)
+        log_weight_sum_of_split_right = np.log(
+            self.cumulative_weights[self.end_index]
+            - self.cumulative_weights[split_indices]
+        )
+        sum_of_log_weights_right = (
+            self.cumulative_log_weights[self.end_index]
+            - self.cumulative_log_weights[split_indices]
+        )
+        right_hand_side = (
+            number_of_samples_right
+            * (log_weight_sum_of_split_right - log_weight_sum - log_density_right)
+            + sum_of_log_weights_right
+        )
 
-            # update the maximum likelihood and the best split index
-            if log_likelihood > maximum_log_likelihood:
-                maximum_log_likelihood = log_likelihood
-                best_split_index = split_index
-
-        return maximum_log_likelihood, best_split_index
+        log_likelihoods = left_hand_side + right_hand_side
+        best = int(np.argmax(log_likelihoods))
+        return float(log_likelihoods[best]), int(split_indices[best])
 
     def log_likelihood_without_split(self) -> float:
         """
@@ -382,7 +414,12 @@ class InductionStep:
             # mount a uniform distribution
             distribution = self.create_uniform_distribution()
 
-            self.nyga_induction.probabilistic_circuit.root.add_subcircuit(
+            # uses the cached root reference rather than `probabilistic_circuit.root`:
+            # that property rescans every node in the graph for the in-degree-0 node
+            # whenever its cache is stale, and every prior leaf attach (a new edge)
+            # invalidates it - going through it here would make each leaf attach cost
+            # O(current graph size), i.e. O(leaves^2) for the whole fit.
+            self.nyga_induction.root_unit.add_subcircuit(
                 UnivariateContinuousLeaf(
                     distribution,
                     probabilistic_circuit=self.nyga_induction.probabilistic_circuit,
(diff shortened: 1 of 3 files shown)


probabilistic_model/src/probabilistic_model/probabilistic_circuit/rx/probabilistic_circuit.py
```diff
@@ -481,19 +481,31 @@ def latent_variable(self) -> Symbolic:
...
 
     def forward(self, *args, **kwargs):
-        self.result_of_current_query = np.sum(
-            [
-                np.exp(weight) * subcircuit.result_of_current_query
-                for weight, subcircuit in self.log_weighted_subcircuits
-            ],
-            axis=0,
-        )
+        # Accumulated one subcircuit at a time instead of building the full
```
tomsch420: remove the big comments

probabilistic_model/src/probabilistic_model/probabilistic_circuit/rx/probabilistic_circuit.py
```diff
@@ -481,19 +481,31 @@ def latent_variable(self) -> Symbolic:
...
+        for weight, subcircuit in self.log_weighted_subcircuits:
+            result = result + np.exp(weight) * subcircuit.result_of_current_query
+        self.result_of_current_query = result
 
     def log_forward(self, *args, **kwargs):
-        result = [
-            lw + s.result_of_current_query for lw, s in self.log_weighted_subcircuits
-        ]
-        self.result_of_current_query = logsumexp(result, axis=0)
+        # See the note in forward(): accumulate with np.logaddexp instead of
```
tomsch420: remove this big comment

probabilistic_model/src/probabilistic_model/learning/nyga_induction.py
```diff
@@ -382,7 +414,12 @@ def induce(self) -> List[Self]:
             # mount a uniform distribution
             distribution = self.create_uniform_distribution()
 
-            self.nyga_induction.probabilistic_circuit.root.add_subcircuit(
+            # uses the cached root reference rather than `probabilistic_circuit.root`:
```
tomsch420: remove this big comment

probabilistic_model/src/probabilistic_model/learning/nyga_induction.py
```diff
@@ -218,38 +218,70 @@ def compute_best_split(self) -> Tuple[float, Optional[int]]:
...
-        maximum_log_likelihood = -float("inf")
-        best_split_index = None
+        first_split_index = self.begin_index + self.min_samples_per_quantile
+        last_split_index = self.end_index - self.min_samples_per_quantile + 1
 
-        # calculate the connecting points
-        right_connecting_point = self.right_connecting_point()
-        left_connecting_point = self.left_connecting_point()
+        # no candidate split respects min_samples_per_quantile on both sides
+        if first_split_index >= last_split_index:
```
tomsch420, line 265: how can this happen?

probabilistic_model/src/probabilistic_model/learning/nyga_induction.py
```diff
@@ -218,38 +218,70 @@ def compute_best_split(self) -> Tuple[float, Optional[int]]:
...
-            # update the maximum likelihood and the best split index
-            if log_likelihood > maximum_log_likelihood:
-                maximum_log_likelihood = log_likelihood
-                best_split_index = split_index
+        # left-hand side: same formula as log_likelihood_of_split_side(split_index,
+        # left_connecting_point), which always resolves to the "is_left" branch since
+        # every split_value lies to the right of left_connecting_point
+        number_of_samples_left = split_indices - self.begin_index
+        log_density_left = np.log(split_values - left_connecting_point)
+        log_weight_sum_of_split_left = np.log(
```
tomsch420: why is this so much more complicated than before?

probabilistic_model/src/probabilistic_model/learning/nyga_induction.py
tomsch420: Whenever this induction step's own segment is already too small to split while keeping min_samples_per_quantile points on both sides — e.g. a segment of 15 points with min_samples_per_quantile=10 (first_split_index = begin+10, last_split_index = end-10+1, and end-begin=15 makes first_split_index > last_split_index). This is the recursion's base case: no candidate clears the floor, so -inf is returned, improvement_is_good_enough(-inf) is always False, and induce() falls through to mounting a leaf instead of splitting further.

The old per-candidate loop handled this implicitly (an empty range just means the loop body never runs, leaving the sentinel default). The vectorized version needs it explicit: np.argmax on an empty array raises, since np.arange(first_split_index, last_split_index) would be empty.

probabilistic_model/src/probabilistic_model/learning/nyga_induction.py
tomsch420: It's not more complicated in substance — it's the exact same formula as log_likelihood_of_split_side, just evaluated for every candidate split at once instead of once per candidate in a Python loop (which was the actual cost: this function is called on every recursion step, over every candidate split index).

I've factored it into a shared _vectorized_log_likelihood_of_split_side helper (pushed) so the left/right calls are now one line each and the mirrored structure with the original method is explicit — should read closer to before now.

probabilistic_model/src/probabilistic_model/learning/nyga_induction.py
```diff
@@ -211,45 +211,88 @@ def sum_log_weights(self):
...
+        """
+        Log likelihood of one side of a candidate split, evaluated for every candidate
+        split index of :meth:`compute_best_split` at once. Which side is being evaluated
+        is passed in directly (`is_left`) rather than inferred from the data, since
+        :meth:`compute_best_split` already knows it statically: every `split_value` lies
+        to the right of the left connecting point and to the left of the right
+        connecting point.
+        """
+        log_weight_sum = np.log(self.total_weights)
+        if is_left:
```
LucaKro: this is basically duplicated code on both branches no? abstract it into its own method please and call that here

LucaKro (changes requested):

probabilistic_model/src/probabilistic_model/learning/nyga_induction.py
tomsch420: i had the same thought but i thought yagni :D anyway now that there are 2 people with that thought ill tell claude to do it :D

tomsch420: Fair, done — pushed. Both branches only differed in which (begin, end) index pair fed number_of_samples and the two cumulative-array lookups, so I collapsed them into picking (begin, end) once and reusing sum_weights_from_indices/sum_log_weights_from_indices (already generic over scalar or array indices), plus abs() for the density sign the same way the original scalar method did instead of branching on it. Down to 9 lines from 22, same output (leaf counts/log-likelihoods unchanged on the benchmark grid, full test suite passes).

LucaKro (approved):

ichumuh (approved):

PR #579, opened 2026-08-31
Plan Transformations

Tigul:
# Descripiton 
Adds Rules for transforming a plan after it is expanded. 
Allows the plan to be adjusted to specific scenarios beyond the generic action definition. 

First use-case scenario is the DetectBeforeGrasp where a LookAtAction and DetectAction into a Reach Action


sunava (approved):

LucaKro:
interesting, but i think this should wait until at least after tomorrows cramgineers meeting :pray:

coraplex/examples/plan_transformations.md
tomsch420, line 1: it gives me RDR/Rule Trees vibes so perhaps use rdrs?

tomsch420 (changes requested):
This doesnt need to be addressed in this PR but here is one more problem i can see: The rules make assumptuons more implicit in my opinion. It give me "just preprocess a bit more and it will work" vibes.

coraplex/examples/plan_transformations.md
```diff
@@ -0,0 +1,254 @@
...
+
+`InsertionTransformation` takes the position the nodes are inserted at: `BEFORE` or `AFTER` the
+anchor make them its siblings, `BELOW` makes them its last children. The same transformation with
+another position parks the arms once the robot has arrived instead:
+
+```python
+from coraplex.datastructures.enums import InsertionPosition
+
+context.plan_transformations = [
+    ParkArmsBeforeNavigating(position=InsertionPosition.AFTER)
```
LucaKro: i think this should be encoded inside the transformation, not as a parameter. This now conflicts with the name imo, as it says "ParkArms*Before*Navigating"

coraplex/examples/plan_transformations.md
```diff
@@ -0,0 +1,331 @@
...
+## Where the Nodes Land
+
+`InsertionTransformation` takes the position the nodes are inserted at: `BEFORE` or `AFTER` the
+anchor make them its siblings, `BELOW` makes them its last children. The same transformation with
+another position parks the arms once the robot has arrived instead:
+
+```python
+from coraplex.datastructures.enums import InsertionPosition
+
+context.plan_transformations = [
```
AbdelrhmanBassiouny: If the same transformation class can have different positions, then you should not name the transformation class according to one specific position, like here `ParkArmsBeforeNavigating` is now positioned after navigating, right?

coraplex/examples/plan_transformations.md
```diff
@@ -0,0 +1,331 @@
...
+drives that are free to move them:
+
+```python
+@dataclass
+class ParkArmsBeforeFreeDrives(ParkArmsBeforeNavigating):
+    """
+    Parks the arms only before drives that do not have to hold the joints where they are.
+    """
+
+    def applies_to(self, plan_node: PlanNode) -> bool:
```
AbdelrhmanBassiouny: that indeed looks like rule based to me

coraplex/src/coraplex/datastructures/enums.py
```diff
@@ -325,6 +325,27 @@ class MonitorBehavior(Enum):
...
+    """
+    As the left neighbour of the anchor node.
+    """
+
+    AFTER = auto()
+    """
+    As the right neighbour of the anchor node.
+    """
+
+    BELOW = auto()
```
AbdelrhmanBassiouny: Should this be named AS_CHILD instead?

coraplex/src/coraplex/plans/plan.py
```diff
@@ -242,15 +244,68 @@ def add_nodes_from(self, nodes_for_adding: Iterable[PlanNode]):
...
+        return self.context.plan_transformations if self.context else []
+
+    def apply_plan_transformations(self, node: PlanNode):
+        """
+        Rewrites the plan with every transformation that applies to the given node.
+
+        :param node: The node that was just expanded
+        """
+        for transformation in self.plan_transformations:
+            if transformation.applies_to(node):
```
AbdelrhmanBassiouny: the classic collision/conflict of rules will likely happend where two rules both apply to the same node which will you apply? or if both in which order? and is that correct?

AbdelrhmanBassiouny: you should at least check for conflict and report that as an error if this is not correct behaviour. Actually there's a pattern made exactly for such situations in krrood/patterns about specificity and finding the best match and reporting conflicts

coraplex/src/coraplex/plans/plan_node.py
```diff
@@ -542,9 +576,19 @@ def __hash__(self):
...
+        designator_field = [
+            f"{field.name}: {getattr(self.designator, field.name)}"
+            for field in self.designator.fields
+        ]
+        parent_infos.append(
+            "---------------- Designator Parameter --------------------"
+        )
+        parent_infos.extend(
+            [
+                f"Designator Type: {self.designator.__class__.__name__}",
```
AbdelrhmanBassiouny: make these string keys an StrEnum instead or if possible make a dataclass that is convertible to Dict

AbdelrhmanBassiouny: so something like a NodeInfo class.

coraplex/src/coraplex/robot_plans/plan_transformations.py
```diff
@@ -0,0 +1,114 @@
...
+from krrood.entity_query_language.factories import a, variable
+from semantic_digital_twin.reasoning.predicates import InsideOf
+from semantic_digital_twin.semantic_annotations.semantic_annotations import Drawer
+from semantic_digital_twin.spatial_types.spatial_types import Pose
+
+# %% perceiving before a grasp
+
+
+@dataclass
+class DetectBeforeGrasp(InsertionTransformation, ActionTransformation[ReachAction]):
```
AbdelrhmanBassiouny: DetectBeforeReach would be better here imo

AbdelrhmanBassiouny: because what if reaching is used for other tasks other than grasping like pushing or cutting

AbdelrhmanBassiouny (changes requested):
Nice work

coraplex/examples/plan_transformations.md
LucaKro: yea and i think these rules should be an explicit part of the class. then there we can also say conditions like "if the endeffector of the robot has a camera..." for plan transformation that enables visual servoing for example

coraplex/src/coraplex/plans/plan.py
LucaKro: whats that called? or is it what rdr use?

coraplex/src/coraplex/robot_plans/plan_transformations.py
LucaKro: then DetectAfterReach would be more fitting imo, the usecase of this (the stretch demo) needs it between the end of reach and the start of grasp

coraplex/src/coraplex/plans/plan_node.py
```diff
@@ -341,6 +342,24 @@ def notify(self):
...
+        """
+        # Keyed by identity, holding the node itself: expanding a child simplifies the
+        # plan, which can remove nodes and free their graph index for a later one.
+        notified: Dict[int, PlanNode] = {}
+        pending = self.children
+        while pending:
+            notified.update({id(child): child for child in pending})
+            for child in pending:
+                child.notify()
+            pending = [child for child in self.children if id(child) not in notified]
```
LucaKro, line 374: node id as UUID? 👉 👈

coraplex/src/coraplex/plans/plan_transformation.py
```diff
@@ -0,0 +1,138 @@
...
+class PlanTransformation(Generic[NodeType], SubClassSafeGeneric, ABC):
+    """
+    Rewrites the part of a plan that a node expanded into.
+
+    A transformation is applied to every node it applies to, right after that node has
+    been expanded and before the nodes below it are expanded in turn.
+    """
+
+    @property
+    def node_type(self) -> Type[PlanNode]:
```
LucaKro: type for your generic

coraplex/src/coraplex/plans/plan_transformation.py
```diff
@@ -0,0 +1,138 @@
...
+        """
+        :return: The type of node this transformation rewrites.
+        """
+        return get_generic_type_parameters(
+            type(self), PlanTransformation, include_root_generic_base=False
+        )[0]
+
+    def applies_to(self, plan_node: PlanNode) -> bool:
```
LucaKro: i think there should be a more general condition method that may be implemented.

coraplex/src/coraplex/plans/plan_transformation.py
```diff
@@ -0,0 +1,138 @@
...
+):
+    """
+    Rewrites the plan of actions of the bound action type.
+    """
+
+    @property
+    def node_type(self) -> Type[PlanNode]:
+        # Concrete transformations of this family bind the action type, so the node
+        # type is read from what the family itself binds.
+        return get_generic_type_parameters(
```
LucaKro: i think you should be able to have one generic implementation of this that works for all future subclasses, so that you dont need to override it for each subclass, which kind of defeats the purpose of this abstraction imo

coraplex/src/coraplex/plans/plan_transformation.py
```diff
@@ -0,0 +1,138 @@
...
+    The nodes are built anew on every application, since a node belongs to the one plan
+    it was inserted into.
+    """
+
+    position: InsertionPosition = InsertionPosition.BEFORE
+    """
+    Where the inserted nodes are placed relative to the anchor node.
+    """
+
+    _insertion_methods: ClassVar[Dict[InsertionPosition, Callable[..., None]]] = {
```
LucaKro: Enum?

coraplex/src/coraplex/plans/plan_transformation.py
```diff
@@ -0,0 +1,138 @@
...
+        return super().applies_to(plan_node) and isinstance(
+            plan_node.designator, self.action_type
+        )
+
+
+# %% rewriting
+
+
+@dataclass
+class InsertionTransformation(ABC):
```
LucaKro: I think if we are going for CRUD behaviors, then we will need a superclass for them

coraplex/src/coraplex/robot_plans/plan_transformations.py
```diff
@@ -0,0 +1,114 @@
...
+                accept_first_if_multiple=True,
+            ),
+        ]
+
+
+# %% opening what the object lies in
+
+
+@dataclass
+class OpenDrawerBeforePickUp(
```
LucaKro: yea i think its sus that the names specify where to insert, but you can override the insert position

LucaKro: also should a transform like this also include that its closed again afterwards?

LucaKro (changes requested):
very cool 💪

coraplex/src/coraplex/robot_plans/plan_transformations.py
Tigul: I thought about adding the closing action but this is more complicated, because you can close with the other arm but what is with robots that only have one arm? They need to finish the transport first and then return. So I kept it simple for now

coraplex/src/coraplex/robot_plans/plan_transformations.py
Tigul: I think both are wrong because that is inserting something into a Reach action so saying before or after suggest it is executed before or after the whole action and not during the action

coraplex/examples/plan_transformations.md
Tigul: @LucaKro But isn't that exactly like it is now?

coraplex/src/coraplex/plans/plan.py
Tigul: Yes that will definitely something to look out for. Finding conflicting transformations will not be trivial because applying two different transformations to the same action can be an intendet behaviour. 
But you are right that we should at least warn people

coraplex/src/coraplex/robot_plans/plan_transformations.py
```diff
@@ -0,0 +1,114 @@
...
+    Opens the drawers an object lies in before the robot picks it up, so that it reaches
+    into an open drawer instead of a closed one.
+    """
+
+    minimum_containment_ratio: float = 0.9
+    """
+    How much of the object has to lie within a drawer for it to count as being in it.
+    """
+
+    def containing_drawers(self, pick_up: PickUpAction) -> List[Drawer]:
```
LucaKro: i think this is an implicit condition that should be explicit to be checked before the the transformation is applied. right now its always applied if its the correct node, but if this condition "doesnt hold" the transformation just doesnt do anything

coraplex/src/coraplex/plans/plan.py
AbdelrhmanBassiouny: @LucaKro yes rdr's detect these at the moment a rule is added and prompts expert with the conflict and asks for a differentiating condition.

What I was referring to though is another approach that can be easily done here without the rdrs (it is not a solution though, it is a more correct implementation by choosing most specific not just what applies (if that's the intended behaviour here ofc) and a conflict detection mechanism where you choose to raise or report the conflict. 

RDR's prevent conflicts from existing in the code because it adds rules incrementatly and catches conflicts on addition. But they do not solve them automatically (sometimes it is possible automatically, the new rdrs can do that), they still need the expert.

coraplex/src/coraplex/robot_plans/plan_transformations.py
LucaKro: but it is actually execute after the reach action no?

coraplex/examples/plan_transformations.md
Tigul: Removed that option

coraplex/src/coraplex/robot_plans/plan_transformations.py
Tigul: No, the reach includes the tcp moving to the pose of the object, after the pre-pose, and this is right in between

coraplex/src/coraplex/plans/plan.py
Tigul: The future plan for this would be to use rdrs as a selection mechanism for the plan transformations and let the Plan Transformations be the atomic operations on the plan

coraplex/src/coraplex/plans/plan.py
```diff
@@ -242,15 +244,71 @@ def add_nodes_from(self, nodes_for_adding: Iterable[PlanNode]):
...
+
+    @property
+    def plan_transformations(self) -> List[PlanTransformation]:
+        """
+        The transformations that rewrite this plan while it is expanded.
+
+        :return: The transformations of this plan's context; a plan without a context
+            has none.
+        """
+        return self.context.plan_transformations if self.context else []
```
LucaKro: i think this should fail, as usage errors like this:
```python
 plan.plan_transformations.append(transformation) 
```
where context isnt set would be swallowed silently

coraplex/src/coraplex/plans/plan_transformation.py
```diff
@@ -0,0 +1,170 @@
...
+
+NodeType = TypeVar("NodeType", bound=PlanNode)
+ActionType = TypeVar("ActionType", bound=ActionDescription)
+
+
+# %% transformations
+
+
+@runtime_checkable
+class PlanTransformation(Protocol):
```
LucaKro, line 29: hmm this is sus, whats the reason you dont want to inherit from this directly?

the more i learn about protocols the more sus i feel like they are...i think explicit inheritance is better.

also runtime_checkable apparently only matches names of the methods, not signature, which smells funny as well idk

coraplex/src/coraplex/plans/plan_transformation.py
```diff
@@ -0,0 +1,170 @@
...
+        Reports whether the case the node describes needs the transformation, which
+        every case does unless a transformation says otherwise.
+
+        It is asked only about nodes :meth:`applies_to_node` selected, so the node can
+        be read as the type this is bound to.
+
+        :param plan_node: A node this selects
+        :return: Whether the transformation is needed here.
+        """
+        return True
```
LucaKro, line 93: why? i think when it comes to transformations a True as default feels dangerous, worst case we apply transformations we dont want

coraplex/src/coraplex/plans/plan_transformation.py
```diff
@@ -0,0 +1,170 @@
...
+        return super().applies_to_node(plan_node) and isinstance(
+            plan_node.designator, self.action_type
+        )
+
+
+# %% rewriting
+
+
+@dataclass
+class PlanRewrite(ABC):
```
LucaKro, line 119: why did you chose rewrite instead of transformation?

coraplex/src/coraplex/plans/plan_transformation.py
LucaKro, line 1: chose either rewrite or transformation and stick to it i think

coraplex/src/coraplex/plans/underspecified.py
```diff
@@ -171,6 +178,16 @@ class UnderspecifiedNode(ExecutionBoundaryNode):
     On failure, `advance` replaces it with the next candidate.
     """
 
+    current_attempt: Optional[SequentialNode] = field(
```
LucaKro: i think the anme is really confusing, i would have expected that this is already what "current_candidate" would do

coraplex/src/coraplex/robot_plans/plan_transformations.py
```diff
@@ -0,0 +1,250 @@
...
+        :return: The reach's last motion, which brings the tool center point onto the
+            object.
+        """
+        motions = [
+            node for node in plan_node.descendants if isinstance(node, MotionNode)
+        ]
+        return motions[-1]
+
+    def anchor(self, plan_node: ActionNode) -> PlanNode:
+        return self.final_approach(plan_node)
```
LucaKro, line 60: why not just implement the body of final approach in here? i dont see any benefit from extracting it

coraplex/src/coraplex/robot_plans/plan_transformations.py
```diff
@@ -0,0 +1,250 @@
...
+
+    def nodes_to_insert(self, plan_node: ActionNode) -> List[ActionLike]:
```
LucaKro, line 62: cant you type for the generic here, so that you dont need to cast?

coraplex/src/coraplex/robot_plans/plan_transformations.py
```diff
@@ -0,0 +1,250 @@
...
+    def final_approach(self, plan_node: ActionNode) -> MotionNode:
+        """
+        :param plan_node: The node of the reach
+        :return: The reach's last motion, which brings the tool center point onto the
+            object.
+        """
+        motions = [
+            node for node in plan_node.descendants if isinstance(node, MotionNode)
+        ]
+        return motions[-1]
```
LucaKro, line 57: this seems a bit fragile to me, what if another transformation already modified this action?

coraplex/src/coraplex/robot_plans/plan_transformations.py
```diff
@@ -0,0 +1,250 @@
...
+    minimum_opening_ratio: float = 0.9
+    """
+    How far along its travel a drawer has to stand pulled out to count as open already.
+    """
+
+    @property
+    def position(self) -> InsertionPosition:
+        return InsertionPosition.BEFORE
+
+    def closed_drawers_holding(
```
LucaKro: holding? 😅

coraplex/src/coraplex/robot_plans/plan_transformations.py
```diff
@@ -0,0 +1,250 @@
...
+        :param arm: The arm that opens it
+        :param context: The context the drive to the handle is grounded against
+        :return: The drive that makes the handle reachable and the opening itself.
+        """
+        handle = drawer.handle.root
+        return [
+            a(NavigateAction)(
+                target_location=variable(
+                    Pose,
+                    domain=reachability_location(handle.global_pose, context, arm),
```
LucaKro, line 130: shouldnt this use accessing location?

coraplex/src/coraplex/robot_plans/plan_transformations.py
```diff
@@ -0,0 +1,250 @@
...
+                ),
+                keep_joint_states=True,
```
LucaKro, line 132: why?

coraplex/src/coraplex/robot_plans/plan_transformations.py
```diff
@@ -0,0 +1,250 @@
...
+                accept_first_if_multiple=True,
+            ),
+        ]
+
+
+# %% opening what the object lies in
+
+
+@dataclass
+class DrawerOpening(InsertionRewrite):
```
LucaKro, line 80: is this abc? i feel like there are methods missing no?

coraplex/src/coraplex/robot_plans/plan_transformations.py
```diff
@@ -0,0 +1,250 @@
...
+    """
+    Opens the drawers an object lies in before the robot picks it up, so that it reaches
+    into an open drawer instead of a closed one.
+
+    Nothing else positions the robot for a pick-up of its own, and opening a drawer
+    leaves the robot standing at its handle, so the rewrite ends by parking and driving
+    to a pose the object itself can be reached from.
+    """
+
+    def is_applicable(self, plan_node: PlanNode) -> bool:
```
LucaKro: generic typing?

coraplex/src/coraplex/robot_plans/plan_transformations.py
LucaKro: also i think those methods should be private, it seems like they should only be used inside the class

coraplex/src/coraplex/robot_plans/plan_transformations.py
```diff
@@ -0,0 +1,250 @@
...
+        nodes.extend(
+            [
+                ParkArmsAction(Arms.BOTH),
+                a(NavigateAction)(
+                    target_location=variable(
+                        Pose,
+                        # Built when the drive is grounded rather than now: the drawers
+                        # this rewrite opens stand open by then, and a pose the object
+                        # can be reached from only exists once they do.
+                        domain=DeferredLocation(
```
LucaKro, line 174: i dont think you need a deferred location bc of the underspecified no?

coraplex/src/coraplex/robot_plans/plan_transformations.py
```diff
@@ -0,0 +1,250 @@
...
+        ):
+            nodes.extend(self.opening_nodes(drawer, transport.arm, transport.context))
+        return nodes
+
+
+# %% parking before anything else
+
+
+@dataclass
+class ParkArmsBeforeFirstAction(InsertionRewrite, PlanMatch[ActionNode]):
```
LucaKro, line 222: i feel like this is a bit pointless, because its ever only apply to exactly one node 😅 at this point just insert part arms in the beginning no?

LucaKro: plus you cant really enforce this because transformations kinda depend on the order of application right? so the only way to guarantee this is to apply it as the very last transformation, and not allow any others afterwards, otherwise this may lie to you

krrood/src/krrood/patterns/subclass_safe_generic.py
```diff
@@ -120,6 +123,23 @@ def get_generic_type_parameters(cls) -> List[Type]:
         """
         return get_generic_type_parameters(cls, SubClassSafeGeneric)
 
+    @classmethod
+    def get_type_of_generic_parameter(cls, parameter: TypeVar) -> Type:
```
LucaKro: i *think* the return type is wrong, because it may also be a tuple bc typevartuple

LucaKro (changes requested):
nice, taking good shape i think! 💪

coraplex/src/coraplex/plans/plan_transformation.py
Tigul: I wanted to do inheritance first but that fails a bit it would need to inherit a PlanMatch and a PlanRewrite but then there are different cases you want to check, like PlanNodeMatch vs. ActionMatch and then inheritence becomes ugly. 
I think this is a fine solution since it is only used for Typing anyway

coraplex/src/coraplex/plans/plan_transformation.py
Tigul: Because I wanted to seperate the different concepts. A Transformation consists of a Match and a Rewrite. I could name either a Transformation but that would be confusing

coraplex/src/coraplex/robot_plans/plan_transformations.py
Tigul: I can tighter define the condition here but realisticly the problem will always remain. If two Transformations modify the same sort of the plan that that is something that needs to be fixed a layer above

coraplex/src/coraplex/robot_plans/plan_transformations.py
Tigul: No, accessing location is just a weaker version of reachability and I should probably get rid of it

coraplex/src/coraplex/robot_plans/plan_transformations.py
Tigul: No  there is nothing missing

coraplex/src/coraplex/robot_plans/plan_transformations.py
Tigul: I do need it since the factory is build at binding time only the sampling is done in the underspecified. So if I would not use a deferred location the Costmap would be build on wrong assumptions

coraplex/src/coraplex/robot_plans/plan_transformations.py
Tigul: Yeah I could do that but that is a common use-case so I thought having a Transformation that handles that would be nice.

Regarding the enforcement, that is similar to the case where multiple Transforms act on the same piece of the Plan. You can't really say what would happen then. So I would argue a Transformation does exactly what it says in the name _at the moment it is applied_

coraplex/src/coraplex/robot_plans/plan_transformations.py
Tigul: You mean the keep_joint_states?

coraplex/src/coraplex/plans/plan_transformation.py
Tigul: This is a PlanMatch, that should match on each node type that is given. If an implementation wants to apply only on certain nodes of this type the method needs to be overridden but otherwise this is the intended behavior.

coraplex/src/coraplex/robot_plans/plan_transformations.py
Tigul: That is already the generic in this case

Tigul: The cast is for the action in the Node not the node itself

coraplex/src/coraplex/robot_plans/plan_transformations.py
Tigul: The closed drawers that hold the annotation

Tigul: maybe contain could be better? :thinking:

coraplex/src/coraplex/plans/plan_transformation.py
LucaKro: no i dont think we should use protocols in our own library, as we have full control of our class structure here. if inheritance turns ugly i think we should take a look at the class structure in question again.

So far everything we needed could be done via inheritance or association, i dont see why this would be different here

coraplex/src/coraplex/plans/plan_transformation.py
LucaKro: hmm okay

coraplex/src/coraplex/robot_plans/plan_transformations.py
LucaKro: then just type plan_node to the correct type directly so that the static type checker catches it

coraplex/src/coraplex/robot_plans/plan_transformations.py
LucaKro: yea there will always be conflicts, but i think transforms should at least guarantee the correct transform at the time of application. if afterwards another transform changes sth again thats a different story, but then a transform is applied, it should always be applied correctly. in this case you could match to the correct motion type and then  use the last one or sth no?

coraplex/src/coraplex/robot_plans/plan_transformations.py
LucaKro: drawer_containing_annotation?

coraplex/src/coraplex/robot_plans/plan_transformations.py
LucaKro: yes

LucaKro: as far as i can tell keep_joint_states doesnt even do anything in the plan

```python
@dataclass
class MoveMotion(BaseMotion):
    """
    Moves the robot to a designated location.
    """

    target: Pose
    """
    Location to which the robot should be moved
    """

    keep_joint_states: bool = False
    """
    Keep the joint states of the robot during/at the end of the motion.
    """

    def perform(self):
        return

    @property
    def _motion_chart(self):
        return (
            SetOdometry(
                base_pose=self.target.to_homogeneous_matrix(),
                odom_connection=self.robot.root.parent_connection,
            )
            if GiskardExecutable.execution_type == ExecutionType.SIMULATED and self.context.teleport_as_navigation_in_simulation
            else CartesianPose(
                root_link=self.world.root,
                tip_link=self.robot.root,
                goal_pose=self.target,
            )
        )

```

coraplex/src/coraplex/robot_plans/plan_transformations.py
LucaKro: yes there is, it is missing nodes_to_insert no? or am i missing sth?

coraplex/src/coraplex/robot_plans/plan_transformations.py
LucaKro: idk i dont like accepting that the transform can lie to you.

if we say "detectafterreach", and something inserts a node between reach and detect, then the detect is still after the reach, so the name of the transformation still conveyed the correct thing.

if we say ParkArmsBeforeFirstAction and something inserts before that, it just doesnt hold anymore, which feels like the transform is lying

coraplex/src/coraplex/plans/plan_transformation.py
LucaKro: but this sounds like a transformation is a rewrite with extra stuff no? basically inheritance?

coraplex/src/coraplex/plans/plan.py
```diff
@@ -295,19 +295,38 @@ def plan_transformations(self) -> List[PlanTransformation]:
         :return: The transformations of this plan's context; a plan without a context
             has none.
         """
-        return self.context.plan_transformations if self.context else []
+        if not self.context:
+            raise ContextIsUnavailable
```
LucaKro, line 299: i dont think this raises correctly, did you test it?

coraplex/src/coraplex/plans/underspecified.py
```diff
@@ -178,7 +178,7 @@ class UnderspecifiedNode(ExecutionBoundaryNode):
     On failure, `advance` replaces it with the next candidate.
     """
 
-    current_attempt: Optional[SequentialNode] = field(
+    current_candidate_sequence: Optional[SequentialNode] = field(
```
LucaKro, line 181: is it always a sequence? i thought we also have other control structures

LucaKro (changes requested):

coraplex/src/coraplex/robot_plans/plan_transformations.py
Tigul: It is

coraplex/src/coraplex/robot_plans/plan_transformations.py
Tigul: That is implemented in the sub-classes

coraplex/src/coraplex/plans/plan_transformation.py
Tigul: A transformation is a rewrite with a Match but the match can't be an inherintence. Let's look at this tomorrow and then I show you what I mean

coraplex/src/coraplex/robot_plans/plan_transformations.py
Tigul: Yes that would be possible

coraplex/src/coraplex/robot_plans/plan_transformations.py
Tigul: changed it to closed_drawer_containing(annotation)

coraplex/src/coraplex/robot_plans/plan_transformations.py
Tigul: Hmm, I can check that. keep_joints is an old remnent from when Bass integrated PyBullet and Mujuco

coraplex/src/coraplex/robot_plans/plan_transformations.py
Tigul: I wouldn't say the Transformation is lying to you, It does what it says but only in the moment and not overall. 

We could think about re-ordering the transformations such that they do not interfere with each other but that will probably not be possible in every case and is defenitly out of scope for now. Alternatively, we don't allow two to modify the same node which I don't like

coraplex/src/coraplex/plans/underspecified.py
Tigul: Not in this case, this statically links a sequence to the current candidate

coraplex/src/coraplex/robot_plans/plan_transformations.py
Tigul: I thought about this a bit more and have a few ideas how to handle that but I think most of them are very similar to what RDRs already handle. 
On the other hand is there no current case where PlanTransformation act on the same node, so I call yagni until there is a case where this happens.

coraplex/src/coraplex/robot_plans/plan_transformations.py
Tigul: final approach is used two times and ensures that the correct MoveTCP motion is found

coraplex/src/coraplex/robot_plans/plan_transformations.py
LucaKro: no i would say that ParkArmsBeforeFirstAction should just be removed for now, I dont see a big benefit of it anyways and but a clear downside if people try to extend transformations and suddenly get a sideeffect like this. i dont see the this conceptual problem for the other transformations yet, as ParkArmsBeforeFirstAction makes a very strong claim on the position of the action, whereas the others dont as much

coraplex/src/coraplex/plans/plan.py
LucaKro: i unresolved this, as it wasnt answered nor changed, just resolved

coraplex/src/coraplex/robot_plans/plan_transformations.py
```diff
@@ -0,0 +1,260 @@
...
+            if isinstance(node, MotionNode)
+            and isinstance(node.motion, MoveToolCenterPointMotion)
+        ]
+        return motions[-1]
+
+    def anchor(self, plan_node: ActionNode) -> PlanNode:
+        return self.final_approach(plan_node)
+
+    def nodes_to_insert(self, plan_node: ActionNode) -> List[ActionLike]:
+        reach = cast(ReachAction, plan_node.action)
```
LucaKro, line 71: can it happen that plan_node.action is not actually reach? idk i find this a bit sus

test/coraplex_test/test_designator/test_multi_robot_action_designator.py
```diff
@@ -803,6 +807,8 @@ def test_transport_open_container(mutable_multiple_robot_apartment, rclpy_node):
 
     if isinstance(robot, HSRB):
         return
+    context.plan_transformations.append(OpenDrawerBeforeTransport())
```
LucaKro, line 810: why do we need both? i feel like we should only need one here no?

LucaKro (changes requested):
gj, almost done! I also replied on two existing threads

coraplex/src/coraplex/plans/plan.py
Tigul: wdym, I changed that

test/coraplex_test/test_designator/test_multi_robot_action_designator.py
Tigul: Yes that was my initial thought as well but the structure for transport is:
Navigate -> PickUp -> Navigate -> Place 
But the first navigate searches for a pose when the drawer is closed, so inserting before PickUp does not work in this context

coraplex/src/coraplex/robot_plans/plan_transformations.py
Tigul: No, the transformation defines Reach in the generic and check on isinstance not issubclass. So plan_node.action can only be Reach. I could remove the cast but that would only introduce more yellow lines in the code for something we know we enforce

coraplex/src/coraplex/robot_plans/plan_transformations.py
Tigul: No I disagree, there is already a mechanism that warns if two transformations apply to the same node and this makes a distinction of nodes that are added to the plan to make it work vs. nodes that the define the behaviour we want to achieve

LucaKro: I dont think it makes that particular distinction. but I talked to vanessa and simon about it, and although i still think this transformation should be deleted, now its mostly because i think its too simple and useless, because it doesnt actually save you any work compared to just writing it into the beginning of your plan, and just makes it look more "fancy" and complicated than it actually is. But they did convince me that the most sustainable path forward is just assuming that the transformation does exactly what you want at application time, and that we have the same problem with ordering in the collision avoidance rules, and there we accept it. So im okay keeping it as is for now

coraplex/src/coraplex/plans/plan.py
LucaKro: im pretty sure ContextIsUnavailable is still raised just as a class, not an instance, which results in python under the hood doing ContextIsUnavailable(), which isnt allowed because it needs a `instance: Designator`, which is not supplied. The result is (i think), that the exception isnt raised correctly, and thus cant be caught correctly, and doesnt show the correct error message.

coraplex/src/coraplex/robot_plans/plan_transformations.py
LucaKro: hmm okay i see...i think this can be improved, but its fine for now. Keep the cast :+1:

test/coraplex_test/test_designator/test_multi_robot_action_designator.py
LucaKro: that sounds like the transformation is not working as intended no?. So at least the OpenDrawerBeforePickup can be removed here right?

coraplex/src/coraplex/plans/plan.py
Tigul: Oh thats what you mean ... I thought you mean the check in the if that is not raising

test/coraplex_test/test_designator/test_multi_robot_action_designator.py
Tigul: No that means the two actions have different structures that have to be transformed differentlzy

tomsch420 (approved):

LucaKro (approved):

PR #581, opened 2026-08-31
Explicit failure: a node's terminal state says how it ended

ichumuh:
## Summary

A motion statechart node used to end in a single terminal state, `DONE`, whatever had
happened to it. This branch replaces it with three verdicts — `SUCCEEDED`, `FAILED` and
`INTERRUPTED` — decided by what the node observes at the moment it is ended, and makes
those verdicts readable by other nodes, so a transition can be written on how a node
finished instead of on what it happens to be observing.

`FAILED` already existed in the enum on `main` and even had a plot colour, but no
transition could produce it.

## What to look at

**`data_types.py` — the verdicts.** `verdict_for(observation)` decides which terminal state
a node reaches: observing true succeeds it, observing false fails it, and an observation
with no answer yet is no basis for a judgement. A node taken down by an ancestor ending is
`INTERRUPTED` — it never got judged on its own terms.

**`motion_statechart.py` — only a running node observes.** An observation is now evaluated
only while `RUNNING`, held while `PAUSED`, and unknown everywhere else, terminal states
included. This is the change with the widest blast radius: a condition can no longer read a
reading that stopped being taken, which is what pushes conditions onto the verdicts.

**`graph_node.py` — reading another node's state.** `node.is_succeeded` / `is_failed` /
`is_interrupted` / `is_terminated` / … are trinary tests on the life cycle state; the
verdict ones stay *unknown* until the node is judged. `node.goal_reached` is what most
callers want: the observation while the node runs, the verdict once it has ended.

**`NextLifeCycle` — same-cycle resolution.** Predicates resolve against the state a node
reaches *this* control cycle, so a node waiting on another's verdict starts on the cycle
that verdict arrives rather than the one after. Cycles between nodes are caught at compile
time.

**Templates.** `Sequence` chains on `previous.is_succeeded`; `Parallel`, `TryAll`, `Open`
and `Close` count `goal_reached`; `TryInOrder` advances only once an alternative has
actually failed, not merely while it is still short of its goal.

## Breaking changes

- `LifeCycleValues.DONE` is gone — use `SUCCEEDED`, or `is_terminal` when any ending will do.
- An ended node's `observation_variable` now reads unknown instead of its last value.
  Conditions that read it across a node's end must move to `goal_reached` or a verdict
  predicate.
- `ProgressStalled` → `StillProgressing`, **with inverted polarity**: it is true while
  progress is being made. A node's observation is its goal, so a monitor has to observe the
  good state or it would be recorded as succeeding when it detects trouble.
- `NonObservationVariableError` → `UnsupportedConditionVariableError`.
- `trinary_logic_or` was filtering out its `const_true` arguments where it meant
  `const_false`; fixed in krrood.


coraplex/src/coraplex/datastructures/enums.py
```diff
@@ -130,6 +134,21 @@ def color(self) -> str:
             TaskStatus.PAUSE: "yellow",
         }[self]
 
+    @classmethod
+    def from_life_cycle_state(cls, life_cycle_state: LifeCycleValues) -> TaskStatus:
+        """
+        :param life_cycle_state: The life cycle state of a motion statechart node.
+        :return: The status that state stands for.
+        """
+        return {
```
LucaKro: why create the dict everytime this method is called?

coraplex/src/coraplex/plans/executables.py
```diff
@@ -247,7 +248,9 @@ def _add_pause_interrupt(self, tasks: List[Task]) -> List[ObservationStateValues
             )
             self._current_motion_state_chart.add_node(interrupt_monitor)
             if index > 0:
-                previous_done = tasks[index - 1].observation_variable
+                # the verdict, not the observation behind it, which is gone once the
```
LucaKro: remove, this sounds like claude reasoning for the change

coraplex/src/coraplex/language_giskard_templates.py
```diff
@@ -54,41 +61,70 @@ class TryInOrder(Goal):
...
     The child nodes tried one after another, in order.
     """
 
+    _alternatives: List[MotionStatechartNode] = field(default_factory=list, init=False)
+    """
+    The nodes that were passed in, captured before :meth:`expand` adds a progress
+    monitor per alternative to :attr:`nodes` alongside them.
+    """
+
+    give_up_after: float = field(default=DEFAULT_STALL_TIMEOUT, kw_only=True)
```
tomsch420: timedetla?

giskardpy/src/giskardpy/motion_statechart/monitors/progress_monitors.py
```diff
@@ -177,29 +174,34 @@ def build_artifacts(self, context: MotionStatechartContext) -> NodeArtifacts:
 
 # %% watching a whole goal
 
+DEFAULT_STALL_TIMEOUT = 5.0
```
tomsch420: grrrr, no module level variables

giskardpy/src/giskardpy/motion_statechart/monitors/progress_monitors.py
```diff
@@ -177,29 +174,34 @@ def build_artifacts(self, context: MotionStatechartContext) -> NodeArtifacts:
...
+    the negation of its observation to a node's end condition to give up on that node.
     """
 
     monitored_node: MotionStatechartNode = field(kw_only=True)
     """
     The task or goal whose progress is watched.
     """
 
-    timeout: float = field(default=5.0, kw_only=True)
+    timeout: float = field(default=DEFAULT_STALL_TIMEOUT, kw_only=True)
```
tomsch420: just put the default here. Also timedelta

giskardpy/src/giskardpy/motion_statechart/data_types.py
```diff
@@ -23,11 +105,116 @@ class FloatEnum(float, Enum):
...
+            a=life_cycle,
+            b_result_cases=[
+                (int(state), Scalar(float(self.truth_value(state))))
+                for state in sorted(LifeCycleValues)
+            ],
+            else_result=Scalar.const_trinary_unknown(),
+        )
+
+
+class LifeCyclePredicate(Enum):
```
tomsch420: Subclass of LifeCyclePredicateDefinition?

giskardpy/src/giskardpy/motion_statechart/graph_node.py
```diff
@@ -378,47 +381,149 @@ def _from_json(
...
 
-    def __init__(self, name: str, motion_statechart_node: MotionStatechartNode):
-        super().__init__(name)
-        self.motion_statechart_node = motion_statechart_node
+    predicate: LifeCyclePredicate = field(kw_only=True)
+    """
+    The test this variable holds the value of.
+    """
+
+    def __init__(
```
tomsch420, line 498: why?

tomsch420 (changes requested):

coraplex/src/coraplex/plans/executables.py
```diff
@@ -337,14 +340,15 @@ def _execute_simulation(self) -> None:
         executor.context.cleanup()
 
         if not executor.motion_statechart.is_end_motion():
-            failed_nodes = [
+            unfinished_nodes = [
```
Tigul, line 267: I would go with failed nodes here not unfinished

coraplex/src/coraplex/language_giskard_templates.py
Tigul: Why is this only here and not on all templates?

Sanic (approved):

coraplex/src/coraplex/plans/executables.py
ichumuh: you don't care about interrupted?

coraplex/src/coraplex/language.py
```diff
@@ -216,10 +216,19 @@ def add_to_motion_state_chart(
...
             target=self.maximum_repetitions,
         )
+        # A template that derives its own failure monitor, such as RepeatOnStall,
+        # excludes failure_monitor from its constructor entirely, so it must only be
+        # passed on when one was actually given.
+        failure_monitor_kwargs = (
+            {"failure_monitor": self.failure_monitor}
+            if self.failure_monitor is not None
+            else {}
+        )
```
ichumuh, line 226: the alternative is worse, trust @LucaKro and it will probably get refactored soon anyway again

giskardpy/src/giskardpy/motion_statechart/data_types.py
ichumuh: why?

giskardpy/src/giskardpy/motion_statechart/graph_node.py
ichumuh: you mean why this init is needed? because its ancestor is FloatVariable which overwrites `__init__` so every child must do the field assignments. refactoring FloatVariable is out of scope here

coraplex/src/coraplex/language_giskard_templates.py
ichumuh: should be using timedelta everywhere now

giskardpy/src/giskardpy/motion_statechart/data_types.py
tomsch420: because its exactly defined as this but you would need to call .value when you want the usable enum value instead of using the enum value directly.

semantic_digital_twin/src/semantic_digital_twin/world_description/geometry.py
```diff
@@ -86,6 +88,16 @@ class Color:
     Opacity of the color.
     """
 
+    HEX_COLOR_PREFIX: ClassVar[str] = field(default="#", init=False, repr=False)
+    """
+    The character a hex color is conventionally written behind.
+    """
+
+    HEX_COLOR_PATTERN: ClassVar[str] = field(default="(?:[0-9a-fA-F]{2}){3,4}", init=False, repr=False)
```
tomsch420: these two both look to me as they should be parameters of the to/from_hex functions that have these values as default

tomsch420 (changes requested):

tomsch420 (approved):

coraplex/src/coraplex/plans/executables.py
Tigul: Ok yeah also true

Tigul (approved):

PR #585, opened 2026-09-01
Confidence-aware OOD check for EQL rule evaluation

ganeshyadavyatham:
this PR generates synthetic data that looks like real SDT objects using the EQL probabilistic backend, converts each object with `to_dao`, and extracts mass and class features with `FeatureExtractor`. the class is preprocessed as an enum, then variables are inferred and a `JointProbabilityTree` is fitted, giving the likelihood of objects. a threshold is calibrated from a percentile of the training likelihoods. when an EQL rule reaches a conclusion, the observer fires — it takes the concluded object, scores its likelihood against the trained model, and raises an exception if the object is unfamiliar.

depends on #584 — 3 tests require the symbolic-distribution fix there; without it, symbolic scoring returns -inf.


experiments/src/experiments/confidence_aware_eql/confidence_guard.py
```diff
@@ -0,0 +1,98 @@
...
+from typing_extensions import Any, Iterator
+
+from experiments.confidence_aware_eql.confidence_model import ConfidenceModel
+
+
+@dataclass
+class UnfamiliarObjectError(DataclassException):
+    """Raised when a rule concludes on an object the confidence model judges unfamiliar."""
+
+    obj: Any
```
tomsch420: no abbrivation and use the word 'instance' instead

experiments/src/experiments/confidence_aware_eql/confidence_guard.py
```diff
@@ -0,0 +1,98 @@
...
+    ) -> None:
+        """
+        Check every object this rule just concluded and raise on the first unfamiliar one.
+
+        :param expression: The rule whose conclusions were processed.
+        :param result: The result carrying the bindings the conclusions just updated.
+        :raises UnfamiliarObjectError: If a conclusion's bound object is unfamiliar.
+        """
+        for conclusion in expression._conclusions_:
+            obj = result.bindings.get(conclusion.variable._id_)
```
tomsch420: no abbrivations

experiments/src/experiments/confidence_aware_eql/confidence_model.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+@dataclass
+class ConfidenceModel:
+    """A fitted joint probability tree scoring the familiarity of an object."""
+
+    circuit: ProbabilisticCircuit
+    """The tractable circuit learned from the familiar objects."""
+
+    column_order: List[str]
```
tomsch420: this order is literally [v.name for v in circuit.variables]

tomsch420: dont make this a field is what i want to say, just make it a property

experiments/src/experiments/confidence_aware_eql/confidence_model.py
```diff
@@ -0,0 +1,84 @@
...
+        """Whether an object is familiar under the model.
+
+        :param obj: The object to judge.
+        :return: ``True`` when the object's log-likelihood is at or above the
+            familiarity threshold, ``False`` otherwise.
+        """
+        return self.log_likelihood_of(obj) >= self.threshold
+
+
+def fit_confidence_model(objects: List[Any]) -> ConfidenceModel:
```
tomsch420: make this a classmethod as factory of ConfidenceModel and call it fit_from_instances

experiments/src/experiments/confidence_aware_eql/data_generation.py
```diff
@@ -0,0 +1,95 @@
...
+from krrood.entity_query_language.factories import a
+from krrood.parametrization.model_registries import DictRegistry
+from krrood.parametrization.parameterizer import UnderspecifiedParameters
+from probabilistic_model.probabilistic_circuit.rx.helper import fully_factorized
+from semantic_digital_twin.world_description.inertial_properties import Inertial
+from semantic_digital_twin.world_description.world_entity import Body
+from typing_extensions import Any, List, Type
+
+
+class MassVariableMissing(KeyError):
```
tomsch420: we have mass in the class Intertial. Type for that instead of key errors here

experiments/src/experiments/confidence_aware_eql/data_generation.py
```diff
@@ -0,0 +1,95 @@
...
+    :param mass_distributions: The per-class distributions to sample objects from.
+    :return: The generated objects, grouped by distribution in the given order.
+    """
+    generated_objects = []
+    for distribution in mass_distributions:
+        generated_objects.extend(_generate_objects_for_distribution(distribution))
+    return generated_objects
+
+
+def _mass_variable_name(object_class: Type) -> str:
```
tomsch420: this should be done using an eql statement and shouldnt have its own funiction

experiments/src/experiments/confidence_aware_eql/data_generation.py
```diff
@@ -0,0 +1,95 @@
...
+    mass_variable = parameters.variables[mass_variable_name]
+
+    circuit = fully_factorized(
+        parameters.variables.values(),
+        means={mass_variable: distribution.mean},
+        variances={mass_variable: distribution.standard_deviation},
+    )
+    registry = DictRegistry({distribution.object_class: circuit})
+    backend = ProbabilisticBackend(
+        registry, number_of_samples=distribution.number_of_samples
```
tomsch420: you can realize this better by using a .limit() call on the query

experiments/src/experiments/confidence_aware_eql/data_generation.py
```diff
@@ -0,0 +1,95 @@
...
+
+def _mass_variable_name(object_class: Type) -> str:
+    """
+    :param object_class: The semantic-annotation class whose mass leaf is queried.
+    :return: The dotted access-path name EQL assigns to that class' root inertial mass.
+    """
+    return f"{object_class.__name__}.root.inertial.mass"
+
+
+def _generate_objects_for_distribution(distribution: MassDistribution) -> List[Any]:
```
tomsch420: this whole data generation is done way better then you did it in the last try. Well done!

experiments/src/experiments/confidence_aware_eql/feature_pipeline.py
```diff
@@ -0,0 +1,80 @@
...
+import enum
+
+import pandas as pd
+from krrood.ormatic.data_access_objects.dao import to_dao
+from krrood.parametrization.feature_extraction.feature_extractor import FeatureExtractor
+from probabilistic_model.learning.jpt.variables import infer_variables_from_dataframe
+from random_events.variable import Variable
+from typing_extensions import Any, List
+
+MASS_FEATURE = "mass"
```
tomsch420: no module level variables

experiments/src/experiments/confidence_aware_eql/feature_pipeline.py
```diff
@@ -0,0 +1,80 @@
...
+            MASS_FEATURE: extracted[mass_column].to_numpy(),
+            CLASS_FEATURE: [
+                ObjectClass(type(instance).__name__) for instance in objects
+            ],
+        }
+    )
+    return dataframe
+
+
+def infer_feature_variables(dataframe: pd.DataFrame) -> List[Variable]:
```
tomsch420: this is just an alias, remove the method

tomsch420 (changes requested):

experiments/src/experiments/confidence_aware_eql/data_generation.py
```diff
@@ -0,0 +1,95 @@
...
+    """
+    Sample ``distribution.number_of_samples`` objects of ``distribution.object_class``.
+
+    :param distribution: The mass distribution to sample objects from.
+    :return: The generated objects.
+    :raises MassVariableMissing: If the object class' query has no
+        ``root.inertial.mass`` variable to parameterize.
+    """
+    query = a(distribution.object_class)(root=a(Body)(inertial=a(Inertial)(mass=...)))
```
AbdelrhmanBassiouny: an(Inertial)*

experiments/src/experiments/confidence_aware_eql/feature_pipeline.py
AbdelrhmanBassiouny: maybe put them in a `Feature` StrEnum

test/experiments_test/confidence_aware_eql/test_confidence_guard.py
```diff
@@ -0,0 +1,63 @@
...
+        for mass in np.random.normal(0.25, 0.03, 40)
+    ]
+    return fit_confidence_model(cups)
+
+
+def _concluding_rule_for(cup: Cup) -> SymbolicExpression:
+    """A trivial always-true rule whose conclusion binds ``cup`` to a result variable."""
+    cup_variable = variable(Cup, domain=[cup])
+    condition = HasType(cup_variable, Cup)
+    concluded = variable(Cup, domain=[])
```
AbdelrhmanBassiouny: that should be a deduced_variable()

test/experiments_test/confidence_aware_eql/test_data_generation.py
```diff
@@ -0,0 +1,101 @@
...
+    ]
+
+
+def test_generate_familiar_objects_returns_the_requested_count_per_class(
+    mass_distributions,
+):
+    """One generated object per requested sample, of the matching class, in order."""
+    np.random.seed(0)
+    generated = generate_familiar_objects(mass_distributions)
+    assert [type(obj) for obj in generated] == [Cup] * 20 + [Pot] * 20
```
AbdelrhmanBassiouny: where did you specify that it should generate 20 of each?

AbdelrhmanBassiouny (changes requested):
welld done!

experiments/src/experiments/confidence_aware_eql/confidence_guard.py
ganeshyadavyatham: I have replaced obj with instance wherever  it is present , also scanned my code again to see if there are any other abbreviations. found excinfo which is pytest's own

experiments/src/experiments/confidence_aware_eql/confidence_guard.py
ganeshyadavyatham: replaced with instance

experiments/src/experiments/confidence_aware_eql/confidence_model.py
ganeshyadavyatham: for this instead of storing it as a list, now I have made it a property which can be used whenever it is required by calling that property

experiments/src/experiments/confidence_aware_eql/feature_pipeline.py
ganeshyadavyatham: removed the function [def infer_feature_variables(dataframe: pd.DataFrame) -> List[Variable]: ] which does nothing and copies to a different function infer_variables_from_dataframe(dataframe) . so directly used infer_variables_from_dataframe(dataframe) function

experiments/src/experiments/confidence_aware_eql/feature_pipeline.py
ganeshyadavyatham: changed the scope the variables from global to inline local function . declared only in the functions they are required as per Tom's comment . later again changed it to feature strEnum as per Abdelrhman's comment.

experiments/src/experiments/confidence_aware_eql/data_generation.py
ganeshyadavyatham: yes  instead of passing no of samples in backend we can use query.expression.limit(n) . made edits accordingly

experiments/src/experiments/confidence_aware_eql/data_generation.py
ganeshyadavyatham: removed [class MassVariableMissing(KeyError):] and function [def _mass_variable_name(object_class: Type) -> str: ] as eql query itself is enough to get the info related to mass variable and raise an error if something is incorrect

experiments/src/experiments/confidence_aware_eql/data_generation.py
ganeshyadavyatham: removed [class MassVariableMissing(KeyError):] and function [def _mass_variable_name(object_class: Type) -> str: ] as eql query itself is enough to get the info related to mass variable and raise an error if something is incorrect

experiments/src/experiments/confidence_aware_eql/confidence_model.py
ganeshyadavyatham: I have changed this to class method instead of function and called it as fit_from_instances

experiments/src/experiments/confidence_aware_eql/data_generation.py
ganeshyadavyatham: done

experiments/src/experiments/confidence_aware_eql/data_generation.py
ganeshyadavyatham: Thank you sir

test/experiments_test/confidence_aware_eql/test_data_generation.py
ganeshyadavyatham: it comes from mass_distributions fixture. there I have defined mass for cup and one pot along with number_of_samples=20 . generate_familiar_objects reads count per class and passes it to query.expression.limit. so 20 cups + 20 pots are generated.

test/experiments_test/confidence_aware_eql/test_confidence_guard.py
ganeshyadavyatham: changed the concluded variable to deduced_variable and I have kept cup_variable as variable(Cup,domain=[cup]) as it is HasType's subject which needs a domain to enumerate

experiments/src/experiments/confidence_aware_eql/confidence_guard.py
```diff
@@ -0,0 +1,98 @@
...
+        for conclusion in expression._conclusions_:
+            instance = result.bindings.get(conclusion.variable._id_)
+            if instance is None:
+                continue
+            if not self.confidence_model.is_familiar(instance):
+                raise UnfamiliarObjectError(instance)
+
+
+def evaluate_with_confidence_guard(
+    expression: SymbolicExpression, confidence_model: ConfidenceModel
```
tomsch420, line 82: i think the typing should be something like experission: SymbolicExpression[T] -> Iterator[T]

experiments/src/experiments/confidence_aware_eql/confidence_model.py
```diff
@@ -0,0 +1,85 @@
...
+from typing_extensions import Any, List
+
+from experiments.confidence_aware_eql.feature_pipeline import extract_feature_dataframe
+
+
+@dataclass
+class ConfidenceModel:
+    """A fitted joint probability tree scoring the familiarity of an object."""
+
+    circuit: ProbabilisticCircuit
```
tomsch420: i think you should use a RelationProbabilisticCircuit here

experiments/src/experiments/confidence_aware_eql/data_generation.py
```diff
@@ -0,0 +1,75 @@
...
+    :param mass_distributions: The per-class distributions to sample objects from.
+    :return: The generated objects, grouped by distribution in the given order.
+    """
+    generated_objects = []
+    for distribution in mass_distributions:
+        generated_objects.extend(_generate_objects_for_distribution(distribution))
+    return generated_objects
+
+
+def _generate_objects_for_distribution(distribution: MassDistribution) -> List[Any]:
```
tomsch420: i think this should be a method of MassDistribution

experiments/src/experiments/confidence_aware_eql/feature_pipeline.py
```diff
@@ -0,0 +1,73 @@
...
+
+import enum
+
+import pandas as pd
+from krrood.ormatic.data_access_objects.dao import to_dao
+from krrood.parametrization.feature_extraction.feature_extractor import FeatureExtractor
+from typing_extensions import Any, List
+
+
+class ObjectClass(enum.Enum):
```
tomsch420: inherit from StrEnum

tomsch420 (changes requested):

experiments/src/experiments/confidence_aware_eql/feature_pipeline.py
ganeshyadavyatham: Done

experiments/src/experiments/confidence_aware_eql/data_generation.py
ganeshyadavyatham: I have removed the separate standalone function _generate_objects_for_distribution and moved this code logic into MassDistribution as generate_objects() method. now we call generate_objects() method to generate objects for that distribution

experiments/src/experiments/confidence_aware_eql/confidence_guard.py
ganeshyadavyatham: I have looked into this and conducted audits using Claude and found that it has no typing parameter Generic[T] declared on SymbolicExpression .so subscripting it as SymbolicExpression[T] is not possible and fails.

experiments/src/experiments/confidence_aware_eql/confidence_model.py
ganeshyadavyatham: I have gone through example classes  in semantic_annotations.py . in our case we are doing OOD/familiarity  on end graspable objects . the data for these objects is flat and also in synthetic data I am using flat data(mass+class) and relationalprobabilisticcircuit is needed for something like wall has doors. I have doubt about the relationalprobabilisticcircuit scope here,so in our case do we need to check something relational like if a cup has water vs empty cups  something like that . or  kitchen -> cupboard->cups vs livingroom->table-> cups. or is it just a flat single object .

experiments/src/experiments/confidence_aware_eql/confidence_guard.py
```diff
@@ -0,0 +1,98 @@
...
+    set_evaluation_context,
+)
+from krrood.exceptions import DataclassException
+from typing_extensions import Any, Iterator
+
+from experiments.confidence_aware_eql.confidence_model import ConfidenceModel
+
+
+@dataclass
+class UnfamiliarObjectError(DataclassException):
```
tomsch420: *UnfamiliarInstanceError

experiments/src/experiments/confidence_aware_eql/confidence_guard.py
tomsch420: sadness but ok, thanks for looking

test/experiments_test/confidence_aware_eql/test_feature_pipeline.py
tomsch420, line 1: i think these tests are not needed

experiments/src/experiments/confidence_aware_eql/confidence_guard.py
tomsch420: can you also add information about the log-likelihood of the instance here and the threshhold that was required?

experiments/src/experiments/confidence_aware_eql/confidence_model.py
tomsch420: In the minimal example you added this is true. However, the general scope is to be able to use this with abitrary classes that also have assoications. It should work for your example no different than its currently working already. if not, we can go back to ordinary probabilistic circuit, but in the research agenda of the institute, the RelationalProbabilisticCircuit is the new future of querying, so id like to have this work already.

tomsch420 (changes requested):

experiments/src/experiments/confidence_aware_eql/confidence_guard.py
ganeshyadavyatham: Done

ganeshyadavyatham: I have added instance's loglikelihood and the threshold to the error message

test/experiments_test/confidence_aware_eql/test_feature_pipeline.py
ganeshyadavyatham: I have removed the redundant test (test_dataframe_labels_each_object_with_its_class). the other two tests are for mass value and type-inference . shall I remove those tests as well .

experiments/src/experiments/confidence_aware_eql/confidence_model.py
```diff
@@ -0,0 +1,237 @@
...
+    query = a(domain_class)(**kwargs)
+    query.resolve()
+    return query
+
+
+# %% per-class model
+
+
+@dataclass
+class PerClassConfidenceModel:
```
tomsch420, line 102: interesting, i like

test/experiments_test/confidence_aware_eql/test_feature_pipeline.py
tomsch420, line 1: delete this entire file, its tested elsewhere already

tomsch420 (changes requested):

experiments/src/experiments/confidence_aware_eql/confidence_model.py
ganeshyadavyatham: thank you sir, with per-class approach where N models needs to be built for n types of objects . if we have many types of objects then is it ok to have  many models or  shall I work on building with shared base class approach ? also I have a question related to testing part do we get any objects with missing class labels / properties during real time . or just impossible properties . if we need to check object with missing class label or unseen class then shared base class architecture is good option I think . if this per-class approach is fine , shall I start working on traceability the per-feature attribution using marginal likelihood to raise a warning which feature caused the OOD.

test/experiments_test/confidence_aware_eql/test_feature_pipeline.py
ganeshyadavyatham: done

experiments/src/experiments/confidence_aware_eql/confidence_model.py
tomsch420: We will never have the case where there is an unknown class, as OOP kind of doesnt really allow for this. But the problem you are hinting on here is very real. I have been thinking about how to model this for a couple of month now, and what i found out so far is that it boils down to the 3 representations that SQLAlchemy offers here: Concrete table inheritance (yours), joined table inheritance and single table inheritence. But i have not reached a conclusion whats a good approach here. Also its out of scope for this PR, but if you want to work on that in your masters thesis or some other project i would be happy to have that

tomsch420 (approved):

experiments/src/experiments/confidence_aware_eql/confidence_model.py
ganeshyadavyatham: thank you so much sir. i'd be very interested in doing this as my master's thesis under your supervision, after i complete my guided research. i don't officially meet the criteria to start the thesis yet, so i'll begin it as a voluntary extra-credits project in the meantime, and make some progress before i can formally start the thesis.

as the next step in the guided research, shall i work on traceability first (per-feature attribution to flag which features caused the ood),  then replace the synthetic data with real sdt data and shall i submit that as a new pr?

AbdelrhmanBassiouny (approved):

experiments/src/experiments/confidence_aware_eql/confidence_model.py
tomsch420: sounds good. The scope of this PR is good as it is. Make a new PR based on this where you try to apply this to real sdt data. You can load worlds from the existing resources and try to score every object there. I think attribution to why its unlikely is pretty hard, but you can try to come up with a scheme, which would be pretty cool if you actually find a way. @Narenvasant is looking into causality, perhaps you should sit together for ideas. 

 For the polymorphic stuff, you can write me an email if you want to do that in independent study or as master thesis and we can see what fits best both from the regulations of the university and the scope of the task. Also i think when you tried it on real sdt data, we can conclude the guided research and you get your credits for it.

PR #586, opened 2026-09-01
Added shutdown procedure for Giskard, when using KeyboardInterrupt

MideDay:
- GracefulShutdownSignals: Handles signal replacement. Replaces old signals on entry into with-Block and changes back on exit.
- Giskard: Uses GracefulShutdownSignals to replace standard ROS2 signal handlers.
- MotionServer: Handles KeyboardInterrupt by shutting down ControlLoop, which stops robot as well.

### DIFF
--- a/giskardpy/src/giskardpy/middleware/ros2/graceful_shutdown.py
+++ b/giskardpy/src/giskardpy/middleware/ros2/graceful_shutdown.py
@@ -1,15 +1,17 @@
 from __future__ import annotations
 
 import signal
-from collections.abc import Callable
+from contextlib import AbstractContextManager
 from dataclasses import dataclass, field
 from types import FrameType
+from typing import TYPE_CHECKING
 
-_SignalHandler = Callable[[int, FrameType | None], None] | int | None
+if TYPE_CHECKING:
+    from signal import _HANDLER as SignalHandler
 
 
 @dataclass
-class GracefulShutdownSignals:
+class GracefulShutdownSignals(AbstractContextManager):
     """
     Turns SIGINT and SIGTERM into a :class:`KeyboardInterrupt` for the duration of a
     ``with`` block, and restores whatever handlers were installed before on exit.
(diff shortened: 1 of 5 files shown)


giskardpy/src/giskardpy/middleware/ros2/graceful_shutdown.py
```diff
@@ -0,0 +1,48 @@
...
+import signal
+from collections.abc import Callable
+from dataclasses import dataclass, field
+from types import FrameType
+
+_SignalHandler = Callable[[int, FrameType | None], None] | int | None
+
+
+@dataclass
+class GracefulShutdownSignals:
```
tomsch420: subclass of context manager

giskardpy/src/giskardpy/middleware/ros2/graceful_shutdown.py
```diff
@@ -0,0 +1,48 @@
+from __future__ import annotations
+
+import signal
+from collections.abc import Callable
+from dataclasses import dataclass, field
+from types import FrameType
+
+_SignalHandler = Callable[[int, FrameType | None], None] | int | None
```
tomsch420: hmm is there no type like this already in some existing python types?

giskardpy/src/giskardpy/middleware/ros2/graceful_shutdown.py
```diff
@@ -0,0 +1,48 @@
...
+        self._original_sigint_handler = signal.signal(
+            signal.SIGINT, self._raise_keyboard_interrupt
+        )
+        self._original_sigterm_handler = signal.signal(
+            signal.SIGTERM, self._raise_keyboard_interrupt
+        )
+        return self
+
+    def __exit__(self, exc_type, exc, tb) -> bool:
+        signal.signal(signal.SIGINT, self._original_sigint_handler)
```
tomsch420, line 44: i think here you have to check if actually the signal you are listening for was raised. Otherwise, this will raise a SIGINT/SIGTERM even if the exit block was reached without any problem

tomsch420 (changes requested):

LucaKro (approved):
toms comments, otherwise gucci

giskardpy/src/giskardpy/middleware/ros2/graceful_shutdown.py
MideDay: Added AbstractContextManager

giskardpy/src/giskardpy/middleware/ros2/graceful_shutdown.py
MideDay: there is a signal._HANDLER type. Changed to using that one now

giskardpy/src/giskardpy/middleware/ros2/graceful_shutdown.py
MideDay: This replaces the custom signal handler with the original one from ROS2 and does not raise any signal itself. I added a test to make sure that no SIGINT/SIGTERM is raised, when there was no issue.

tomsch420 (approved):

PR #587, opened 2026-09-02
Failure handling

Tigul:
# Description 
My first version of the failure handling Module. This is more a concept and nothing is really finished or polished here. I'm mainly looking for feedback on the general idea. 

Future Work coudl include: 
* including more EQL to evaluate the conditions of failed actions and their successors 
* A more structured approach to failure hirachies 
* Maybe using RDRs for refinement and strategy selection


coraplex/src/coraplex/failure_handling/failure_handling_strategy.py
```diff
@@ -0,0 +1,236 @@
...
+            return
+        self.propagate(node)
+
+
+@dataclass
+class RetryNode(TargetedResolution):
+    """
+    Run the target's work again as it is, typically after a recovery sub-plan repaired
+    the situation the failure described.
+    """
```
ichumuh: why no implementation?

coraplex/src/coraplex/failure_handling/failure_refiner.py
```diff
@@ -0,0 +1,175 @@
...
+        :return: The refined failure, or the given failure itself if there is nothing to
+            refine.
+        """
+
+
+# %% refinement
+
+
+@dataclass
+class FailureRefiner:
```
ichumuh: when RDR?

PR #590, opened 2026-09-03
Add MarkovChainDistributionTemplate for HMM-shaped probabilistic circuits

tomsch420:
## Summary
- Adds `MarkovChainDistributionTemplate`, a second relational template alongside `ExchangeableDistributionTemplate`: grounds one `RelationalProbabilisticCircuit` per position in a sequence via EQL query parts, then chains the grounded copies into a hidden Markov model instead of combining them under a product.
- Each grounded part's root `SumUnit` interprets its children as the hidden state (per the "Latent Variable Interpretation" theorem in `doc/circuits.md`); a `starting_distribution` and a shared `transition_model` connect consecutive positions.
- Builds the chain directly with `ProductUnit`/`SumUnit` rather than via `SumUnit.mount_with_interaction_terms`, which was found not to preserve a node's own children order once it has been used as its first argument -- unsafe for chaining more than one pairwise combination.
- Promotes the part-prefix renaming helper from `rspn.py` to `helper.py` so both templates can share it (behavior unchanged).

## Example: expressivity gain over `ExchangeableDistributionTemplate`

`ExchangeableDistributionTemplate` grounds every part of a one-to-many relation from the *same* i.i.d. per-object distribution and combines the parts under a `ProductUnit` -- so it cannot express any correlation between positions in a sequence, no matter how it was fit. `MarkovChainDistributionTemplate` grounds the same kind of EQL query parts but chains them, so consecutive positions stay coupled through `transition_model`.

To make the gap measurable, both templates are grounded on a query for a 3-object sequence of `SceneObject`s from a ground-truth 2-state hidden Markov chain (states persist with probability 0.8-0.9 -- e.g. a robot's belief about which room it's scanning tends to stay the same between consecutive observed objects). Emissions are dirac (each state deterministically emits one `SceneObjectType`), so the true joint probability of every 3-object type sequence is exactly computable by hand and can be checked against what each grounded circuit reports:

```python
type_variable = Symbolic(name="type", domain=Set.from_iterable(SceneObjectType))
state_domain = Set.from_iterable(range(2))

starting_probabilities = np.array([0.3, 0.7])
transition_probabilities = np.array([[0.9, 0.1], [0.2, 0.8]])  # states persist

markov_chain_template = MarkovChainDistributionTemplate(
    template_distribution=markov_emission_template,  # dirac emission per hidden state
    starting_distribution=SymbolicDistribution(
        variable=Symbolic(name="state", domain=state_domain),
        probabilities=MissingDict(float, dict(enumerate(starting_probabilities))),
    ),
    transition_model=MultinomialDistribution(
        distribution_variables=(
            Symbolic(name="state", domain=state_domain),
            Symbolic(name="next_state", domain=state_domain),
        ),
        probabilities=transition_probabilities,
    ),
)

# the pre-existing RSPN mechanism: only ever sees one object at a time, so the best
# it can be fit to is the (correct) unconditional marginal type frequency.
exchangeable_template = ExchangeableDistributionTemplate(
    template_distribution=exchangeable_marginal_template,  # same emissions, weighted by starting_probabilities
    latent_variables=[],
)

parts = _scene_object_parts(3)  # a(SceneRoom)(objects=[a(SceneObject)(type=...) for _ in range(3)]).resolve()
grounded_markov = markov_chain_template.ground(parts)
grounded_exchangeable = exchangeable_template.ground(parts, aggregation_statistics={})
```

Querying `.probability(...)` on both grounded circuits for all 8 possible 3-object type sequences against the true forward-algorithm probability `start[i] * trans[i, j] * trans[j, k]`:

| sequence | true P | markov chain | exchangeable | markov err | exchangeable err |
|---|---|---|---|---|---|
| TTT | 0.2430 | 0.2430 | 0.0270 | 0.0000 | 0.2160 |
| TTC | 0.0270 | 0.0270 | 0.0630 | 0.0000 | 0.0360 |
| TCT | 0.0060 | 0.0060 | 0.0630 | 0.0000 | 0.0570 |
| TCC | 0.0240 | 0.0240 | 0.1470 | 0.0000 | 0.1230 |
| CTT | 0.1260 | 0.1260 | 0.0630 | 0.0000 | 0.0630 |
| CTC | 0.0140 | 0.0140 | 0.1470 | 0.0000 | 0.1330 |
| CCT | 0.1120 | 0.1120 | 0.1470 | 0.0000 | 0.0350 |
| CCC | 0.4480 | 0.4480 | 0.3430 | 0.0000 | 0.1050 |

Total variation distance to the ground truth, summed over all 8 sequences: **`MarkovChainDistributionTemplate` = 0.0000** (exact -- expected, since it's literally the forward algorithm) vs **`ExchangeableDistributionTemplate` = 0.3840**. The exchangeable template systematically underestimates persistent sequences (`TTT`, `CCC`) and overestimates sequences that switch type every step (`TCT`, `CTC`), because it has no mechanism to represent that consecutive objects' states are correlated -- it can only ever multiply the same marginal into itself.

### Same numbers, asked through EQL's `probability_of`

The table above calls `.probability(...)` directly on the grounded circuit. The idiomatic EQL entry point is the `probability_of(...)` directive (`Probability`, added in #575): it resolves via a `ProbabilisticBackend`/`ModelRegistry` instead of hand-built `SimpleEvent`s, so the same query reads the same way whether it's backed by a plain fitted circuit or -- as here -- a template freshly grounded from EQL parts.

```python
# align the grounded chain's variables to the qualified names an EQL condition on
# variable(SceneRoom).objects[i].type produces, then register the grounded circuit as
# SceneRoom's model.
grounded_markov = markov_chain_template.ground(parts)
grounded_markov.update_variables({
    old: variable_from_name_and_type(f"SceneRoom.objects[{i}].type", SceneObjectType)
    for i, old in enumerate(old_type_variables)
})
backend = ProbabilisticBackend(model_registry=DictRegistry({SceneRoom: grounded_markov}))

room = variable(SceneRoom)
condition = and_(
    room.objects[0].type == SceneObjectType.CHAIR,
    room.objects[1].type == SceneObjectType.CHAIR,
    room.objects[2].type == SceneObjectType.CHAIR,
)
probability_of(condition).first(backend=backend)  # -> 0.4480, matches the CCC row above
```

Run over all 8 sequences, `probability_of(...)` reproduces the `.probability(...)` column exactly (total variation distance 0.0000) -- same model, same answer, through the declarative query surface.

## Test plan
- [x] `test/krrood_test/test_feature_extraction/test_markov_chain.py` (new, 9 tests): `__post_init__` validation, `ground([])`, single-part grounding, and a 3-part chain checked against the textbook forward-algorithm formula for a hand-built 2-state dirac-emission model.
- [x] `test/krrood_test/test_feature_extraction/` (53 tests, includes the above) -- no regressions.
- [x] `test/probabilistic_model_test/test_bayesian_networks`, `test/probabilistic_model_test/test_rx` (68 tests) -- no regressions.

https://claude.ai/code/session_018ZXsrfn3nxKZjZw98wDeWL

### DIFF
--- a/probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/helper.py
+++ b/probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/helper.py
@@ -3,11 +3,37 @@ from __future__ import annotations
 import rustworkx
 from sortedcontainers import SortedSet
 
+from krrood.utils import get_class_and_attribute_name
 from probabilistic_model.probabilistic_circuit.rx.probabilistic_circuit import (
     ProbabilisticCircuit,
     ProductUnit,
 )
 from random_events.variable import Variable
+
+
+def rename_variables_with_part_prefix(
+    circuit: ProbabilisticCircuit,
+    prefix: str,
+    excluded_variables: list[Variable],
+) -> None:
+    """
+    Rename each variable in the circuit to include ``prefix`` as a namespace.
+
+    Produces names of the form ``"{prefix}.{variable.name}"``. Variables listed in
+    ``excluded_variables`` are left unchanged.
+
+    :param circuit: The circuit whose variables are renamed in-place.
+    :param prefix: String prefix to prepend to every variable name.
+    :param excluded_variables: Variables that should keep their current names.
+    """
+    variable_renames = {
+        variable: type(variable)(
+            get_class_and_attribute_name(prefix, variable.name), domain=variable.domain
+        )
+        for variable in circuit.variables
+        if variable not in excluded_variables
+    }
+    circuit.update_variables(variable_renames)
 
 
 def find_lowest_product_nodes_that_model_variables(
--- a/probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/helper.py
+++ b/probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/helper.py
@@ -3,37 +3,11 @@ from __future__ import annotations
 import rustworkx
 from sortedcontainers import SortedSet
 
-from krrood.utils import get_class_and_attribute_name
 from probabilistic_model.probabilistic_circuit.rx.probabilistic_circuit import (
     ProbabilisticCircuit,
     ProductUnit,
 )
 from random_events.variable import Variable
-
-
-def rename_variables_with_part_prefix(
-    circuit: ProbabilisticCircuit,
-    prefix: str,
-    excluded_variables: list[Variable],
-) -> None:
-    """
-    Rename each variable in the circuit to include ``prefix`` as a namespace.
-
-    Produces names of the form ``"{prefix}.{variable.name}"``. Variables listed in
-    ``excluded_variables`` are left unchanged.
-
-    :param circuit: The circuit whose variables are renamed in-place.
-    :param prefix: String prefix to prepend to every variable name.
-    :param excluded_variables: Variables that should keep their current names.
-    """
-    variable_renames = {
-        variable: type(variable)(
-            get_class_and_attribute_name(prefix, variable.name), domain=variable.domain
-        )
-        for variable in circuit.variables
-        if variable not in excluded_variables
-    }
-    circuit.update_variables(variable_renames)
 
 
 def find_lowest_product_nodes_that_model_variables(
--- a/probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/markov_chain.py
+++ b/probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/markov_chain.py
@@ -1,0 +1,200 @@
+"""
+A relational, EQL-grounded template for hidden Markov models expressed as probabilistic
+circuits.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+
+import numpy as np
+import numpy.typing as npt
+
+from krrood.entity_query_language.query.match import AbstractMatchExpression
+from probabilistic_model.exceptions import ShapeMismatchError
+from probabilistic_model.probabilistic_circuit.relational.exceptions import (
+    EmptyMarkovChainError,
+    NotAProbabilityDistributionError,
+)
+from probabilistic_model.probabilistic_circuit.relational.helper import (
+    rename_variables_with_part_prefix,
+)
+from probabilistic_model.probabilistic_circuit.relational.rspn import (
+    RelationalProbabilisticCircuit,
+)
+from probabilistic_model.probabilistic_circuit.rx.probabilistic_circuit import (
+    ProbabilisticCircuit,
+    ProductUnit,
+    SumUnit,
+)
+
+
+@dataclass
+class MarkovChainDistributionTemplate:
+    """
+    A fitted distribution template for one sequential (Markov chain) relation.
+
+    Grounds one ``RelationalProbabilisticCircuit`` per position in a sequence, the
+    same way :class:`~probabilistic_model.probabilistic_circuit.relational.rspn.ExchangeableDistributionTemplate`
+    grounds one per exchangeable child, and chains the grounded copies into a
+    hidden Markov model instead of combining them under a product.
+
+    Each grounded copy's root is a :class:`~probabilistic_model.probabilistic_circuit.rx.probabilistic_circuit.SumUnit`;
+    its latent-variable interpretation (see the "Latent Variable Interpretation"
+    theorem in ``doc/circuits.md``) is the chain's hidden state at that position:
+    the root's children are the states, and ``root.subcircuits[s]`` is
+    ``P(observation | state = s)``.
+    """
+
+    template_distribution: RelationalProbabilisticCircuit
+    """
+    One time step's fitted relational circuit.
+
+    Its class-level circuit's root is a ``SumUnit``; that sum unit's latent-variable
+    interpretation is the chain's hidden state, shared across every grounded part.
+    """
+
+    starting_distribution: npt.NDArray
+    """
+    1D array of length K (K = number of hidden states); the probability of each
+    hidden state for the first grounded part.
+    """
+
+    transition_model: npt.NDArray
+    """
+    (K, K) row-stochastic array; ``transition_model[i, j]`` is the probability of
+    transitioning to hidden state ``j`` given hidden state ``i``, applied between
+    every consecutive pair of grounded parts.
+    """
+
+    def __post_init__(self):
+        expected_transition_shape = (len(self.starting_distribution),) * 2
+        if self.transition_model.shape != expected_transition_shape:
+            raise ShapeMismatchError(
+                self.transition_model.shape, expected_transition_shape
+            )
+        if np.any(self.starting_distribution < 0) or not np.isclose(
+            self.starting_distribution.sum(), 1.0
+        ):
+            raise NotAProbabilityDistributionError(
+                "starting_distribution", self.starting_distribution
+            )
+        if np.any(self.transition_model < 0) or not np.allclose(
+            self.transition_model.sum(axis=1), 1.0
+        ):
+            raise NotAProbabilityDistributionError(
+                "transition_model", self.transition_model
+            )
+
+    def ground(self, parts_to_ground: list) -> ProbabilisticCircuit:
+        """
+        Ground one copy of the template per part and chain them into a hidden Markov
+        model.
+
+        :param parts_to_ground: The query parts, one per position in the sequence, in
+            sequence order.
+        :return: A circuit over all variables implied by the query, with the per-
+            position hidden states marginalized out.
+        """
+        if len(parts_to_ground) == 0:
+            raise EmptyMarkovChainError()
+
+        result = ProbabilisticCircuit()
+        roots = [
+            self._ground_and_mount_part(result, index, part)
+            for index, part in enumerate(parts_to_ground)
+        ]
+
+        state_count = len(roots[0].subcircuits)
+        expected_shape = (state_count,)
+        if self.starting_distribution.shape != expected_shape:
+            raise ShapeMismatchError(self.starting_distribution.shape, expected_shape)
+
+        # Read every position's emission branches once, up front: these are plain
+        # leaves/subtrees ("P(observation at t | state = s)"), never mutated below, so
+        # reading them here fixes what "state s" means at each position once and for
+        # all -- the chain built below only ever adds new nodes, it never re-derives a
+        # position's state ordering from a circuit structure that could have changed.
+        emission_given_state = [list(root.subcircuits) for root in roots]
+
+        self._point_root_at_hidden_markov_chain(result, roots[0], emission_given_state)
+
+        result.remove_unreachable_nodes(roots[0])
+        result.simplify()
+        return result
+
+    def _ground_and_mount_part(self, result: ProbabilisticCircuit, index: int, part):
+        """
+        Ground one part of the sequence and mount it into ``result``.
+
+        :param result: The circuit the grounded part is mounted into.
+        :param index: Position of the part in the sequence; used as fallback prefix when
+            ``part`` does not carry a symbolic variable.
+        :param part: The query part (a ``Match`` or a concrete domain object).
+        :return: The root of the mounted part, owned by ``result``.
+        """
+        part_circuit = self.template_distribution.ground(part)
+        prefix = (
+            str(part.variable)
+            if isinstance(part, AbstractMatchExpression)
+            else str(index)
+        )
+        rename_variables_with_part_prefix(part_circuit, prefix, [])
+        part_root_index = part_circuit.root.index
+        node_index_map = result.mount(part_circuit.root)
+        return node_index_map[part_root_index]
+
+    def _point_root_at_hidden_markov_chain(
+        self,
+        result: ProbabilisticCircuit,
+        first_root: SumUnit,
+        emission_given_state: list,
+    ) -> None:
+        """
+        Rebuild ``first_root``'s edges so it becomes the root of the whole chain.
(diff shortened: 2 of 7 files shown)


probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/markov_chain.py
```diff
@@ -0,0 +1,200 @@
...
+)
+from probabilistic_model.probabilistic_circuit.rx.probabilistic_circuit import (
+    ProbabilisticCircuit,
+    ProductUnit,
+    SumUnit,
+)
+
+
+@dataclass
+class MarkovChainDistributionTemplate:
```
tomsch420: create a common base here with the EDT, such that the interface which these template collection distributions need to implement is unified

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/markov_chain.py
```diff
@@ -0,0 +1,200 @@
...
+    interpretation is the chain's hidden state, shared across every grounded part.
+    """
+
+    starting_distribution: npt.NDArray
+    """
+    1D array of length K (K = number of hidden states); the probability of each
+    hidden state for the first grounded part.
+    """
+
+    transition_model: npt.NDArray
```
tomsch420: can the be done using the multinomial distribution?

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/markov_chain.py
```diff
@@ -0,0 +1,200 @@
...
+
+    template_distribution: RelationalProbabilisticCircuit
+    """
+    One time step's fitted relational circuit.
+
+    Its class-level circuit's root is a ``SumUnit``; that sum unit's latent-variable
+    interpretation is the chain's hidden state, shared across every grounded part.
+    """
+
+    starting_distribution: npt.NDArray
```
tomsch420: can this be done using symbolic distirubtion or integer distribution?

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/markov_chain.py
```diff
@@ -0,0 +1,200 @@
...
+        """
+        Ground one copy of the template per part and chain them into a hidden Markov
+        model.
+
+        :param parts_to_ground: The query parts, one per position in the sequence, in
+            sequence order.
+        :return: A circuit over all variables implied by the query, with the per-
+            position hidden states marginalized out.
+        """
+        if len(parts_to_ground) == 0:
```
tomsch420, line 89: wouldnt you in this case just return an empty circuit?

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/markov_chain.py
```diff
@@ -0,0 +1,200 @@
...
+            self._ground_and_mount_part(result, index, part)
+            for index, part in enumerate(parts_to_ground)
+        ]
+
+        state_count = len(roots[0].subcircuits)
+        expected_shape = (state_count,)
+        if self.starting_distribution.shape != expected_shape:
+            raise ShapeMismatchError(self.starting_distribution.shape, expected_shape)
+
+        # Read every position's emission branches once, up front: these are plain
```
tomsch420: if you would use proper classes instead of arrays, the interface would speak for itself without this explanation needed

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/markov_chain.py
```diff
@@ -0,0 +1,200 @@
...
+            the root of the whole chain.
+        :param emission_given_state: ``emission_given_state[t][s]`` is
+            ``P(observation_t | state_t = s)``, read once before any mutation.
+        """
+        state_count = len(self.starting_distribution)
+        continuation_given_state = emission_given_state[-1]
+        for position in reversed(range(len(emission_given_state) - 1)):
+            next_continuation_given_state = continuation_given_state
+            continuation_given_state = []
+            for state in range(state_count):
```
tomsch420: too deep nesting

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/markov_chain.py
tomsch420: Done in 3f47fbc9d: added `RelationalDistributionTemplate` (`relational/template.py`), an ABC that both `ExchangeableDistributionTemplate` and `MarkovChainDistributionTemplate` now inherit from. It owns `template_distribution`, the part-prefix resolution (`_prefix_for_part`), and the ground+rename+mount helpers, so both templates implement the same interface.

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/markov_chain.py
tomsch420: Done in 3f47fbc9d: `transition_model` is now a `MultinomialDistribution` over `(state, next_state)` instead of a raw `(K, K)` ndarray. Its own `__post_init__` validates the shape against the two variables' domains, so the manual shape check for it is gone; row-stochasticity is still checked explicitly since `MultinomialDistribution` doesn't assume that on its own.

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/markov_chain.py
tomsch420: Done in 3f47fbc9d: `starting_distribution` is now a `SymbolicDistribution` over a single "state" variable (domain `0..K-1`, the same indices `root.subcircuits` uses) instead of a raw length-`K` ndarray.

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/markov_chain.py
tomsch420: Done in 3f47fbc9d: `ground([])` now returns an empty `ProbabilisticCircuit()` instead of raising. Removed `EmptyMarkovChainError` (it had no other callers).

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/markov_chain.py
tomsch420: Done in 3f47fbc9d, together with the multinomial/symbolic-distribution change above: the state count now comes straight from `starting_distribution.variable.domain`, so there's no separate shape tuple to construct and compare by hand, and the explanatory comment shrank to a one-line mutation-safety note.

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/markov_chain.py
tomsch420: Done in 3f47fbc9d: split the loop into `_continuation_given_state_at` (one loop over state) and `_joint_of_emission_and_transition` (one loop over next_state), so no method has more than one loop nested inside the other.

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/markov_chain.py
```diff
@@ -0,0 +1,249 @@
...
+        if transition_domains != expected_domains:
+            raise ShapeMismatchError(transition_domains, expected_domains)
+
+        starting_probabilities = np.array(
+            [
+                self.starting_distribution.probabilities[hash(state)]
+                for state in state_domain.simple_sets
+            ]
+        )
+        if np.any(starting_probabilities < 0) or not np.isclose(
```
tomsch420: remove these checks and the exception

tomsch420: Done in ac8218ddd: removed the non-negative/sums-to-1 checks and `NotAProbabilityDistributionError` from `__post_init__` (and the now-unused exception class + the 3 tests that only exercised it) — `starting_distribution`/`transition_model` are `SymbolicDistribution`/`MultinomialDistribution` now, so that's their own class's concern, not this template's. Kept the domain-shape check, since that's checking agreement between the two fields, not either one's internal validity.

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/helper.py
```diff
@@ -3,13 +3,39 @@
...
 
+from krrood.utils import get_class_and_attribute_name
 from probabilistic_model.probabilistic_circuit.rx.probabilistic_circuit import (
     ProbabilisticCircuit,
     ProductUnit,
 )
 from random_events.variable import Variable
 
 
+def rename_variables_with_part_prefix(
```
LucaKro: method of ProbabilisticCircuit?

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/template.py
```diff
@@ -0,0 +1,105 @@
...
+        return (
+            str(part.variable)
+            if isinstance(part, AbstractMatchExpression)
+            else str(index)
+        )
+
+    @staticmethod
+    def _ground_and_rename_part(
+        template_distribution: RelationalProbabilisticCircuit,
+        part,
```
LucaKro: typing

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/template.py
```diff
@@ -0,0 +1,105 @@
...
+    sequential one.
+    """
+
+    template_distribution: RelationalProbabilisticCircuit
+    """
+    The fitted ``RelationalProbabilisticCircuit`` shared by every grounded part.
+    """
+
+    @staticmethod
+    def _prefix_for_part(part, index: int) -> str:
```
LucaKro: part typing

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/template.py
```diff
@@ -0,0 +1,105 @@
...
+        :param result: The circuit ``part_circuit`` is mounted into.
+        :param part_circuit: A self-contained circuit for one grounded part.
+        :return: The root of the mounted part, owned by ``result``.
+        """
+        part_root_index = part_circuit.root.index
+        node_index_map = result.mount(part_circuit.root)
+        return node_index_map[part_root_index]
+
+    @abstractmethod
+    def ground(self, parts_to_ground: list, *args, **kwargs) -> ProbabilisticCircuit:
```
LucaKro: list of what?

LucaKro (changes requested):

AbdelrhmanBassiouny (approved):
Answer @LucaKro comments, and I am gucci

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/helper.py
tomsch420: agreed. Claude, throw "part" out of the name of this method and move it to probabilistic circuit

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/template.py
tomsch420: agreed. Claude add the correct type

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/template.py
tomsch420: agreed. Claude add the correct type

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/template.py
tomsch420: agreed. Claude add the correct type

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/helper.py
tomsch420: Done in e37be855b: moved it onto `ProbabilisticCircuit` as `rename_variables_with_prefix` (dropped "part" from the name since the method itself has no notion of parts) and dropped the free function from `relational/helper.py`. Both callers (`RelationalDistributionTemplate._ground_and_rename_part`, `ExchangeableDistributionTemplate._ground_part_circuit`) now call `part_circuit.rename_variables_with_prefix(prefix, excluded_variables)`.

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/template.py
tomsch420: Done in e37be855b: added `part: Match` to `_prefix_for_part`.

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/template.py
tomsch420: Done in e37be855b: added `part: Match` to `_ground_and_rename_part`, imported alongside `AbstractMatchExpression`.

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/template.py
tomsch420: Done in e37be855b: `ground`'s abstract signature is now `parts_to_ground: list[Match]`, and carried the same into the concrete overrides in `MarkovChainDistributionTemplate`/`ExchangeableDistributionTemplate` (plus the `part`/`part_to_ground` helper params along the way) so the interface and its implementations agree.

LucaKro (approved):

PR #592, opened 2026-09-03
Gcs navigate

Tigul:
# Description 
Added a new Navigation Action which uses the GCS for planning a path and then follows it. 
Some small error fixed on the way: 
    * Transformation for Point2 
    * GCS now throws an error if a point is outside the search space
    * Removed global_transform call from DiffDriveGoal


semantic_digital_twin/src/semantic_digital_twin/world_description/graph_of_convex_sets/exceptions.py
```diff
@@ -128,3 +129,32 @@ def error_message(self) -> str:
...
+
+
+@dataclass
+class PointOutsideSearchSpaceError(UsageError):
+    """
+    Raised when a queried point lies beyond the region a graph of convex sets was built
+    over, so nothing is known about whether it is free.
+    """
+
+    point: Point3 | Point2
```
tomsch420: they have a common superclass Point

semantic_digital_twin/src/semantic_digital_twin/world_description/shape_collection.py
```diff
@@ -289,6 +293,15 @@ def event(self) -> Event:
             *[box.simple_event for box in self.bounding_boxes]
         )
 
+    def contains(self, point: Point2 | Point3) -> bool:
```
tomsch420: Point

coraplex/src/coraplex/robot_plans/actions/core/navigation.py
```diff
@@ -105,6 +122,106 @@ def _action_plan(self) -> PlanNode:
         return execute_single(LookingMotion(target=self.target, camera=camera))
 
 
+@dataclass
+class GCSNavigateAction(ActionDescription):
```
tomsch420, line 118: if we want to apply the design paradigms discussed last cramgineers this should be like this imo: 

```python
@dataclass
class FollowTrajectory:
    trajectory: List[Pose2D]

FollowTrajectory(trajectory=Floor().planar_free_space().path_from_to(..., ...))
```
This can become a factory this class but this would be moving the reasoning outside of the definition

tomsch420: in short, no reasoning about the parameters in the action

tomsch420: only interpretation of what they mean

semantic_digital_twin/src/semantic_digital_twin/world_description/graph_of_convex_sets/boxes.py
```diff
@@ -268,9 +269,27 @@ def node_of_point(self, point: PointT) -> Optional[BoxT]:
                 return node
         return None
 
+    def free_node_of_point(self, point: PointT) -> BoxT:
```
tomsch420: i think a similar method exists already, make sure you are not duplicating things here

semantic_digital_twin/src/semantic_digital_twin/world.py
```diff
@@ -2533,6 +2536,13 @@ def transform(
                 reference_frame_R = spatial_object.to_rotation_matrix()
                 target_frame_R = target_frame_T_reference_frame @ reference_frame_R
                 return target_frame_R.to_quaternion()
+            case Point2():
```
tomsch420: O Violation? Perhaps this should be implemented by the classes

tomsch420 (changes requested):

coraplex/src/coraplex/robot_plans/actions/core/navigation.py
```diff
@@ -105,6 +122,106 @@ def _action_plan(self) -> PlanNode:
...
+        :return: The navigation map covering the whole environment.
+        """
+        origin = HomogeneousTransformationMatrix(reference_frame=self.world.root)
+        environment = SemanticEnvironmentAnnotation(
+            root=self.world.root, _world=self.world
+        )
+        extent = environment.as_bounding_box_collection_at_origin(origin).bounding_box()
+        search_space = BoundingBoxCollection(
+            [
+                VolumetricBoundingBox(
```
LucaKro: why volumetric bounding box? i thought thats exactly why we have planar bounding box, for navigation maps

coraplex/src/coraplex/robot_plans/actions/core/navigation.py
```diff
@@ -105,6 +122,106 @@ def _action_plan(self) -> PlanNode:
...
+    target: Pose
+    """
+    Where the robot should stand at the end of the path, with its base.
+    """
+
+    @property
+    def _action_plan(self) -> PlanNode:
+        return sequential([MoveMotion(waypoint) for waypoint in self._path()])
+
+    def _navigation_map(self, floor_level: float) -> PlanarGraphOfBoundingBoxes:
```
LucaKro: this is just duplicated code from HasSupportingSurface.planar_free_space no?

semantic_digital_twin/src/semantic_digital_twin/robots/robot_parts.py
```diff
@@ -637,6 +637,14 @@ def bounding_box(self) -> VolumetricBoundingBox:
...
         ).bounding_box()
 
+    @property
+    def base_radius(self) -> float:
+        """
+        Approximates the radius of the mobile base, as the average between the radius in the x and y axis.
+
+        :return: The approximate radius of the mobile base, in meters.
+        """
+        return (self.bounding_box.depth / 2 + self.bounding_box.width / 2) / 2
```
LucaKro, line 663: why not go with the max? being conservative would be more save here i think

semantic_digital_twin/src/semantic_digital_twin/world_description/graph_of_convex_sets/boxes.py
Tigul: Didn't find anything in the class, can you point me to it?

semantic_digital_twin/src/semantic_digital_twin/world_description/shape_collection.py
LucaKro: or typevar bound to point? does it make sense here to ask if a 2d box "contains" a 3d point? or the other way around?

LucaKro (changes requested):

coraplex/src/coraplex/robot_plans/actions/core/navigation.py
Tigul: Yes but I would like to take everything into account that the robot could collide with. If there is something above the floor but still something the robot could collide with I think a planar bounding box would not incorporate this

semantic_digital_twin/src/semantic_digital_twin/world_description/shape_collection.py
tomsch420: yes sounds right

semantic_digital_twin/src/semantic_digital_twin/world_description/graph_of_convex_sets/boxes.py
tomsch420: GraphOfBoundingBoxes.node_of_point

Tigul: node_of_point is used by this function, this only adds the validation and raises the exceptions

semantic_digital_twin/src/semantic_digital_twin/robots/robot_parts.py
Tigul: Can try that but that is also used in costmap calculation, overestimating the base size will lead to reachability issues

coraplex/src/coraplex/robot_plans/actions/core/navigation.py
Tigul: I will do this in another PR when I do all the actions at once

coraplex/src/coraplex/robot_plans/actions/core/navigation.py
LucaKro: yea i checked and i think its done the exact same way in semantic_digital_twin.semantic_annotations.mixins.HasSupportingSurface.planar_free_space

I think you should use that method, its basically the exact same

semantic_digital_twin/src/semantic_digital_twin/world_description/graph_of_convex_sets/boxes.py
```diff
@@ -96,7 +88,7 @@ class BoundingBoxAdjacency(Generic[BoxT]):
 @dataclass
 class GraphOfBoundingBoxes(
     Generic[BoxT, PointT],
-    GraphOfConvexSets[PointT, BoundingBoxCollection[BoxT]],
+    GraphOfConvexSets[PointT, BoundingBoxCollection[BoxT, PointT]],
```
LucaKro, line 91: why is this needed?

semantic_digital_twin/src/semantic_digital_twin/world_description/geometry.py
```diff
@@ -1376,7 +1442,7 @@ def intersection_with(self, other: Self) -> Optional[Self]:
 
 
 @dataclass(eq=False)
-class VolumetricBoundingBox(AxisAlignedBox):
+class VolumetricBoundingBox(AxisAlignedBox[Point3]):
```
LucaKro, line 1435: especially since we pass Point3 here, passing point3 and VolumetricBoundingBox in the shape collection thing is duplicated information that may contradict no?

LucaKro (changes requested):

semantic_digital_twin/src/semantic_digital_twin/world_description/graph_of_convex_sets/boxes.py
Tigul: Bounding Box collection is agnostic of the type of box is put in there, so to have the right typing for the methods we need to hand over both

Tigul: That is also an architectual problem, but I didn't want to change the architecture just so typing is a bit nicer

LucaKro: hmm okay maybe its a limitation of Generics in python atm

LucaKro (approved):

coraplex/src/coraplex/locations/costmaps.py
```diff
@@ -498,24 +510,35 @@ def create_ray_mask_around_origin(self):
...
+            unoccupied_entities.update(
+                self.world.get_kinematic_structure_entities_of_branch(
+                    self.robot_view.root
                 )
-                for i in range(len(r_t[1]))
-            ]
-        else:
-            res[r_t[1]] = 0
+            )
+        res[r_t[1]] = [
```
tomsch420: no abbrivations

coraplex/src/coraplex/exceptions.py
```diff
@@ -326,6 +326,27 @@ def suggest_correction(self) -> str:
         return "start the perception pipeline before running the plan."
 
 
+@dataclass
+class NoFloorBelowRobot(DataclassException):
+    """
+    Raised when a robot that has to plan its way over a floor stands over none.
+    """
+
+    robot_name: str
```
tomsch420: why just the name and not the actual robot?

semantic_digital_twin/src/semantic_digital_twin/world_description/graph_of_convex_sets/boxes.py
```diff
@@ -268,9 +263,27 @@ def node_of_point(self, point: PointT) -> Optional[BoxT]:
                 return node
         return None
 
+    def free_node_of_point(self, point: PointT) -> BoxT:
```
tomsch420: this is now confusing to use, free_node_of_point and node_of_point sound like they do the same thing. Call them the same, and make node of point protected, because i think your way here is the intended usage

semantic_digital_twin/src/semantic_digital_twin/world_description/graph_of_convex_sets/exceptions.py
tomsch420: its not resolved

semantic_digital_twin/src/semantic_digital_twin/exceptions.py
```diff
@@ -1754,3 +1754,28 @@ def error_message(self) -> str:
...
+
+
+@dataclass
+class NoSupportingSurfaceError(UsageError):
+    """
+    Raised when an annotation's geometry offers no surface anything could be supported
+    on.
+    """
+
+    annotation_name: PrefixedName
```
tomsch420: why the name and not the annotation itself?

tomsch420 (changes requested):

coraplex/src/coraplex/robot_plans/actions/core/navigation.py
```diff
@@ -105,6 +114,126 @@ def _action_plan(self) -> PlanNode:
         return execute_single(LookingMotion(target=self.target, camera=camera))
 
 
+@dataclass
+class GCSNavigateAction(ActionDescription):
```
tomsch420: no abbrivations

tomsch420 (changes requested):

coraplex/src/coraplex/robot_plans/actions/core/navigation.py
Tigul: No

PR #594, opened 2026-09-03
Add rollback system for world modification history

tomsch420:
Every WorldModification subclass gains an undo() method that reverses its effect (Add<->Remove pairs, per-dof hardware-interface state, and generic attribute diffs), and WorldModelModificationBlock.undo() runs them in reverse order. World.rollback_modification_blocks() undoes the most recently completed blocks using this, raising InsufficientModificationHistoryError if the history is too short.

Several Remove*Modification classes now also keep a reference to the removed entity (previously only the id) so they can be re-added by undo(). SetDofHasHardwareInterface now records each affected dof's previous value. AttributeUpdateModification.undo() reverses the recorded diff; for scalar attributes this required krrood's shallow-diff helper to also record the previous value, which it previously discarded.

Made with the help of Claude.

### DIFF
--- a/semantic_digital_twin/src/semantic_digital_twin/world.py
+++ b/semantic_digital_twin/src/semantic_digital_twin/world.py
@@ -59,6 +59,7 @@ from semantic_digital_twin.exceptions import (
     WorldContainsOrphanedDegreeOfFreedom,
     BrokenWorldModificationHistoryError,
     MismatchingWorld,
+    InsufficientModificationHistoryError,
 )
 from semantic_digital_twin.mixin import HasSimulatorProperties
 from semantic_digital_twin.spatial_computations.forward_kinematics import (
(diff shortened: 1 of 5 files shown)


semantic_digital_twin/src/semantic_digital_twin/world_description/world_modification.py
```diff
@@ -62,6 +62,18 @@ def apply(self, world: World):
         :param world: The world to modify.
         """
 
+    @abstractmethod
+    def undo(self, world: World):
```
tomsch420: rename this to revert

semantic_digital_twin/src/semantic_digital_twin/world_description/world_modification.py
```diff
@@ -265,25 +306,47 @@ class RemoveDegreeOfFreedomModification(WorldModification):
 
     dof_id: UUID
```
tomsch420: this is not your change, but not abbrivations

semantic_digital_twin/src/semantic_digital_twin/world.py
```diff
@@ -2613,6 +2614,36 @@ def reset_state_context(self) -> ResetStateContextManager:
     def get_world_model_manager(self) -> WorldModelManager:
         return self._model_manager
 
+    def rollback_modification_blocks(
```
LucaKro, line 2618: Isnt this the oppsite of what we talked about yesterday? :D
We said the reverse should be to a specific Version number, not "n blocks" no?

LucaKro (approved):

semantic_digital_twin/src/semantic_digital_twin/world.py
tomsch420: yes, do you want the other version?

tomsch420: added it

Tigul (approved):

PR #595, opened 2026-09-04
Do() is in a Relationship

Narenvasant:
### **Summary of the Relationship**
Adds relational causal reasoning: a relational model can now keep values it used to discard while grounding, so they can be registered as causes in a causal circuit and queried through cause/causes_effect.

- Lets a relational model (RSPN) keep values it used to throw away while grounding a query, so those values can be used as causes in a causal circuit.                                                    
- Adds a new setting, **_GroundingMode_**, with three choices: keep the old behavior (default), keep a value by sampling it, or keep a value using the exact split the model already learned from training data.                     
- Adds a new function that grounds a relational model for a query and returns a ready-to-use causal circuit in one step.                                                                                                       
- Adds a way to look up a variable in a grounded circuit by a short, readable name instead of its full internal name.                                                                                                          
- Extends the existing query system so **_cause/causes_effect_** questions now work on relational facts after this. 
- Fixes three bugs found while building and reviewing this: a grounding bug that produced a broken circuit, a bug where a single-item circuit failed to rename its own value, and a bug in the exact-split path that silently dropped the connection between a cause and its effect.                                                                                                                                                                       
- Added tests for the new settings, the new function, the query extension, and each bug fix, including tests that check results stay reproducible and that correlations are actually preserved.

### **Why this Relationship**

**_cause and causes_effect_** work only for flat, pre-built circuits registered per class via **_CausalCircuitRegistry_**, like a fixed class-to-circuit mapping. A relational fact (e.g., number of objects) doesn't have one fixed circuit; it depends on the specific scene being queried, and RSPN discarded that fact during grounding regardless of whether anyone wanted to query it. This PR gives relational queries a circuit to route through by grounding one for the specific query on demand, keeping the fact instead of dropping it.

### DIFF
--- a/krrood/src/krrood/parametrization/model_registries.py
+++ b/krrood/src/krrood/parametrization/model_registries.py
@@ -74,14 +74,16 @@ class RelationalCircuitRegistry(ModelRegistry):
     :class:`~krrood.parametrization.exceptions.RelationalCircuitRegistryRequiresMatch`
     rather than failing on a missing attribute.
 
-    When the query also declares ``cause``/``causes_effect``/``confounder`` markers,
-    grounding uses :attr:`causal_grounding_mode` instead of
+    When the query also declares a ``cause`` marker, grounding uses
+    :attr:`causal_grounding_mode` instead of
     :attr:`~probabilistic_model.probabilistic_circuit.relational.rspn.GroundingMode.PREDICTIVE`,
-    so the aggregation latents those markers name survive grounding instead of being
+    so the aggregation latents that marker names survive grounding instead of being
     integrated out, and the result is wrapped as a verified
     :class:`~probabilistic_model.probabilistic_circuit.causal.causal_circuit.CausalCircuit`
     instead of a plain circuit -- one more kind of cause this same registry can answer
-    about, not a parallel system. Non-causal queries are unaffected.
+    about, not a parallel system. Any accompanying ``causes_effect``/``confounder``
+    markers are read later, by ``ProbabilisticBackend``, once it has this
+    ``CausalCircuit`` in hand. Non-causal queries are unaffected.
     """
 
     relational_probabilistic_circuit: RelationalProbabilisticCircuit
--- a/krrood/src/krrood/parametrization/model_registries.py
+++ b/krrood/src/krrood/parametrization/model_registries.py
@@ -10,8 +10,10 @@ from krrood.parametrization.parameterizer import (
 from krrood.utils import get_class_and_attribute_name
 from probabilistic_model.probabilistic_circuit.causal.causal_circuit import (
     CausalCircuit,
+    MarginalDeterminismTreeNode,
 )
 from probabilistic_model.probabilistic_circuit.relational.rspn import (
+    GroundingMode,
     RelationalProbabilisticCircuit,
 )
 from probabilistic_model.probabilistic_circuit.rx.helper import fully_factorized
@@ -71,6 +73,15 @@ class RelationalCircuitRegistry(ModelRegistry):
     parameter classes don't carry -- resolving one of those raises
     :class:`~krrood.parametrization.exceptions.RelationalCircuitRegistryRequiresMatch`
     rather than failing on a missing attribute.
+
+    When the query also declares ``cause``/``causes_effect``/``confounder`` markers,
+    grounding uses :attr:`causal_grounding_mode` instead of
+    :attr:`~probabilistic_model.probabilistic_circuit.relational.rspn.GroundingMode.PREDICTIVE`,
+    so the aggregation latents those markers name survive grounding instead of being
+    integrated out, and the result is wrapped as a verified
+    :class:`~probabilistic_model.probabilistic_circuit.causal.causal_circuit.CausalCircuit`
+    instead of a plain circuit -- one more kind of cause this same registry can answer
+    about, not a parallel system. Non-causal queries are unaffected.
     """
 
     relational_probabilistic_circuit: RelationalProbabilisticCircuit
@@ -78,10 +89,23 @@ class RelationalCircuitRegistry(ModelRegistry):
     The trained relational probabilistic circuit to ground.
     """
 
+    causal_grounding_mode: GroundingMode = GroundingMode.CAUSAL_SAMPLED
+    """
+    Grounding mode used only when the query declares
+    ``cause``/``causes_effect``/``confounder`` markers. Non-causal queries always
+    ground with ``GroundingMode.PREDICTIVE``, unchanged from before.
+    """
+
     def get_model(self, parameters: ModelQueryParameters) -> ProbabilisticModel:
         if not isinstance(parameters, UnderspecifiedParameters):
             raise RelationalCircuitRegistryRequiresMatch(parameters)
-        grounded = self.relational_probabilistic_circuit.ground(parameters.statement)
+        is_causal_query = bool(parameters.search_cause_variables)
+        grounding_mode = (
+            self.causal_grounding_mode if is_causal_query else GroundingMode.PREDICTIVE
+        )
+        grounded = self.relational_probabilistic_circuit.ground(
+            parameters.statement, grounding_mode
+        )
         class_prefix = self.relational_probabilistic_circuit.class_.__name__
         rename_map = {}
         for circuit_var in grounded.variables:
@@ -91,7 +115,18 @@ class RelationalCircuitRegistry(ModelRegistry):
             if qualified_name in parameters.variables:
                 rename_map[circuit_var] = parameters.variables[qualified_name]
         grounded.update_variables(rename_map)
-        return grounded
+        if not is_causal_query:
+            return grounded
+
+        effect_variables = list(parameters.effect_variables_from_causes_effect)
+        tree = MarginalDeterminismTreeNode.from_causal_graph(
+            parameters.search_cause_variables, effect_variables
+        )
+        causal_circuit = CausalCircuit.from_probabilistic_circuit(
+            grounded, tree, parameters.search_cause_variables, effect_variables
+        )
+        causal_circuit.verify_support_determinism()
+        return causal_circuit
 
 
 @dataclass
--- a/krrood/src/krrood/parametrization/model_registries.py
+++ b/krrood/src/krrood/parametrization/model_registries.py
@@ -10,13 +10,18 @@ from krrood.parametrization.parameterizer import (
 from krrood.utils import get_class_and_attribute_name
 from probabilistic_model.probabilistic_circuit.causal.causal_circuit import (
     CausalCircuit,
-    MarginalDeterminismTreeNode,
+)
+from probabilistic_model.probabilistic_circuit.relational.causal import (
+    RelationalCausalCircuit,
 )
 from probabilistic_model.probabilistic_circuit.relational.rspn import (
     GroundingMode,
     RelationalProbabilisticCircuit,
 )
 from probabilistic_model.probabilistic_circuit.rx.helper import fully_factorized
+from probabilistic_model.probabilistic_circuit.rx.probabilistic_circuit import (
+    ProbabilisticCircuit,
+)
 from probabilistic_model.probabilistic_model import ProbabilisticModel
 
 
@@ -74,16 +79,16 @@ class RelationalCircuitRegistry(ModelRegistry):
     :class:`~krrood.parametrization.exceptions.RelationalCircuitRegistryRequiresMatch`
     rather than failing on a missing attribute.
 
-    When the query also declares a ``cause`` marker, grounding uses
-    :attr:`causal_grounding_mode` instead of
-    :attr:`~probabilistic_model.probabilistic_circuit.relational.rspn.GroundingMode.PREDICTIVE`,
-    so the aggregation latents that marker names survive grounding instead of being
-    integrated out, and the result is wrapped as a verified
+    Every query grounds the same way, with :attr:`grounding_mode`: undetermined
+    aggregation latents are always retained, never integrated out. Whether the query
+    also declares a ``cause`` marker only decides what happens to the grounded circuit
+    afterward -- a postprocessing step, not a different way of grounding. A causal
+    query is wrapped as a verified
     :class:`~probabilistic_model.probabilistic_circuit.causal.causal_circuit.CausalCircuit`
-    instead of a plain circuit -- one more kind of cause this same registry can answer
-    about, not a parallel system. Any accompanying ``causes_effect``/``confounder``
+    registering the marked variables; any accompanying ``causes_effect``/``confounder``
     markers are read later, by ``ProbabilisticBackend``, once it has this
-    ``CausalCircuit`` in hand. Non-causal queries are unaffected.
+    ``CausalCircuit`` in hand. A non-causal query gets the grounded circuit back
+    unchanged, retained latents included.
     """
 
     relational_probabilistic_circuit: RelationalProbabilisticCircuit
@@ -91,22 +96,34 @@ class RelationalCircuitRegistry(ModelRegistry):
     The trained relational probabilistic circuit to ground.
     """
 
-    causal_grounding_mode: GroundingMode = GroundingMode.CAUSAL_SAMPLED
+    grounding_mode: GroundingMode = GroundingMode.SAMPLED
     """
-    Grounding mode used only when the query declares
-    ``cause``/``causes_effect``/``confounder`` markers. Non-causal queries always
-    ground with ``GroundingMode.PREDICTIVE``, unchanged from before.
+    How undetermined aggregation latents are represented during grounding.
+
+    See
+    :class:`~probabilistic_model.probabilistic_circuit.relational.rspn.GroundingMode`.
     """
 
     def get_model(self, parameters: ModelQueryParameters) -> ProbabilisticModel:
         if not isinstance(parameters, UnderspecifiedParameters):
             raise RelationalCircuitRegistryRequiresMatch(parameters)
-        is_causal_query = bool(parameters.search_cause_variables)
-        grounding_mode = (
-            self.causal_grounding_mode if is_causal_query else GroundingMode.PREDICTIVE
-        )
+        grounded = self._ground_and_rename(parameters)
+        if not parameters.search_cause_variables:
+            return grounded
+        return self._as_causal_circuit(grounded, parameters)
+
+    def _ground_and_rename(
+        self, parameters: UnderspecifiedParameters
+    ) -> ProbabilisticCircuit:
+        """
+        Ground the relational circuit for ``parameters.statement`` and align its
+        variable names to the ``UnderspecifiedParameters`` convention.
+
+        :param parameters: The parameters extracted from the queried statement.
+        :return: The grounded, renamed circuit.
+        """
         grounded = self.relational_probabilistic_circuit.ground(
-            parameters.statement, grounding_mode
+            parameters.statement, self.grounding_mode
         )
         class_prefix = self.relational_probabilistic_circuit.class_.__name__
         rename_map = {}
@@ -117,18 +134,26 @@ class RelationalCircuitRegistry(ModelRegistry):
             if qualified_name in parameters.variables:
                 rename_map[circuit_var] = parameters.variables[qualified_name]
         grounded.update_variables(rename_map)
-        if not is_causal_query:
-            return grounded
+        return grounded
 
-        effect_variables = list(parameters.effect_variables_from_causes_effect)
-        tree = MarginalDeterminismTreeNode.from_causal_graph(
-            parameters.search_cause_variables, effect_variables
+    @staticmethod
+    def _as_causal_circuit(
+        grounded: ProbabilisticCircuit, parameters: UnderspecifiedParameters
+    ) -> CausalCircuit:
+        """
+        Wrap an already-grounded, already-renamed circuit as a verified
+        ``CausalCircuit`` registering the query's declared cause and effect variables.
+
+        :param grounded: The grounded, renamed circuit to wrap.
+        :param parameters: The parameters extracted from the queried statement, carrying
+            the declared cause and effect variables.
+        :return: A verified, support-deterministic ``CausalCircuit``.
+        """
+        return RelationalCausalCircuit.from_grounded_circuit(
+            grounded,
+            parameters.search_cause_variables,
+            list(parameters.effect_variables_from_causes_effect),
         )
-        causal_circuit = CausalCircuit.from_probabilistic_circuit(
-            grounded, tree, parameters.search_cause_variables, effect_variables
-        )
-        causal_circuit.verify_support_determinism()
-        return causal_circuit
 
 
 @dataclass
--- a/krrood/src/krrood/parametrization/model_registries.py
+++ b/krrood/src/krrood/parametrization/model_registries.py
@@ -149,7 +149,7 @@ class RelationalCircuitRegistry(ModelRegistry):
             the declared cause and effect variables.
         :return: A verified, support-deterministic ``CausalCircuit``.
         """
-        return RelationalCausalCircuit.from_grounded_circuit(
+        return RelationalCausalCircuit().from_grounded_circuit(
             grounded,
             parameters.search_cause_variables,
             list(parameters.effect_variables_from_causes_effect),
--- a/probabilistic_model/doc/causal_reasoning.md
+++ b/probabilistic_model/doc/causal_reasoning.md
@@ -49,7 +49,7 @@ an adjustment set on a `CausalCircuit` is the analyst's assertion, from domain k
 that such a $G$ exists and that the chosen adjustment set satisfies the backdoor criterion
 relative to it. What the package verifies instead is the circuit-side precondition the next
 section's polytime algorithm depends on -- support determinism -- which is a property of the
-probabilistic circuit itself, not of $G$; it is unrelated to the vtree used later for
+probabilistic circuit itself, not of $G$; it is unrelated to the variable tree used later for
 tractability, and to the `MarginalDeterminismTreeNode` structure introduced below, which only
 groups query variables for that check and does not represent $G$'s edges.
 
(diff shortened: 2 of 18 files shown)


krrood/src/krrood/parametrization/model_registries.py
```diff
@@ -71,17 +73,41 @@ class RelationalCircuitRegistry(ModelRegistry):
     parameter classes don't carry -- resolving one of those raises
     :class:`~krrood.parametrization.exceptions.RelationalCircuitRegistryRequiresMatch`
     rather than failing on a missing attribute.
+
+    When the query also declares a ``cause`` marker, grounding uses
+    :attr:`causal_grounding_mode` instead of
+    :attr:`~probabilistic_model.probabilistic_circuit.relational.rspn.GroundingMode.PREDICTIVE`,
```
tomsch420: hmm i think we should change this forever. They should never be integrated out, the sample arrays should just ignore the aggregation dimensions

krrood/src/krrood/parametrization/model_registries.py
```diff
@@ -91,7 +117,18 @@ def get_model(self, parameters: ModelQueryParameters) -> ProbabilisticModel:
             if qualified_name in parameters.variables:
                 rename_map[circuit_var] = parameters.variables[qualified_name]
         grounded.update_variables(rename_map)
-        return grounded
+        if not is_causal_query:
+            return grounded
```
tomsch420: feels like this should be two methods

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/causal.py
```diff
@@ -0,0 +1,200 @@
...
+    ProbabilisticCircuit,
+)
+from random_events.variable import Variable
+
+if TYPE_CHECKING:
+    from krrood.entity_query_language.query.match import Match
+
+logger = logging.getLogger(__name__)
+
+DEFAULT_ADJUSTMENT_REGION_COUNT_WARNING_THRESHOLD = 1000
```
tomsch420: no module level variables

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/causal.py
tomsch420, line 1: here are all methods without a class. I dont like

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/rspn.py
tomsch420, line 1: in my head, causality is something that is like a postprocessing step to grounding and conditioning and so on. Is this not the case? Because if it is the case, this can be implemented way cleaner

tomsch420 (changes requested):

krrood/src/krrood/parametrization/model_registries.py
Narenvasant: Agreed, changed it

krrood/src/krrood/parametrization/model_registries.py
Narenvasant: Done, splited

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/causal.py
Narenvasant: done

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/rspn.py
Narenvasant: yeah i see what you mean. Causal registration is now a separate step after grounding

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/causal.py
```diff
@@ -0,0 +1,255 @@
...
+class RelationalCausalCircuit:
+    """
+    Factory bridging ``RelationalProbabilisticCircuit`` grounding into ``CausalCircuit``
+    construction, mirroring how the rest of the ``relational`` package bridges
+    ``probabilistic_model`` and ``krrood``.
+
+    Holds no state of its own; :meth:`ground` returns a plain ``CausalCircuit``.
+    """
+
+    DEFAULT_ADJUSTMENT_REGION_COUNT_WARNING_THRESHOLD: ClassVar[int] = 1000
```
tomsch420: dont make this a classvar, make this a field with default

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/causal.py
```diff
@@ -0,0 +1,255 @@
...
+                else RelationalCausalCircuit.resolve_variable(circuit, variable)
+            )
+            for variable in variables
+        ]
+
+    @staticmethod
+    def ground(
+        relational_probabilistic_circuit: RelationalProbabilisticCircuit,
+        query: Match,
+        causal_variables: List[Union[Variable, str]],
```
tomsch420: why the unions?

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/causal.py
```diff
@@ -0,0 +1,255 @@
...
+        effect_variables: List[Union[Variable, str]],
+        adjustment_variables: Optional[List[Union[Variable, str]]] = None,
+        grounding_mode: GroundingMode = GroundingMode.SAMPLED,
+        adjustment_region_count_warning_threshold: int = DEFAULT_ADJUSTMENT_REGION_COUNT_WARNING_THRESHOLD,
```
tomsch420: if this would be a field you wouldnt need it here

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/causal.py
```diff
@@ -0,0 +1,255 @@
...
+        """
+        if len(adjustment_variables) < 2:
+            return
+        region_counts = [
+            len(grounded_circuit.marginal([variable]).leaves)
+            for variable in adjustment_variables
+        ]
+        region_product = math.prod(region_counts)
+        if region_product > threshold:
+            logging.getLogger(__name__).warning(
```
tomsch420: define the logger after the import for the entire module

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/causal.py
```diff
@@ -0,0 +1,255 @@
...
+            (
+                variable
+                if isinstance(variable, Variable)
+                else RelationalCausalCircuit.resolve_variable(circuit, variable)
+            )
+            for variable in variables
+        ]
+
+    @staticmethod
+    def ground(
```
tomsch420, line 112: looks like this should be a method pattern that should be called like RelationalCausalCircuit(adjustment_region_count_warning_threshhold=...).ground().

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/causal.py
```diff
@@ -0,0 +1,255 @@
...
+            adjustment_region_count_warning_threshold,
+        )
+
+    @staticmethod
+    def from_grounded_circuit(
+        grounded_circuit: ProbabilisticCircuit,
+        causal_variables: List[Union[Variable, str]],
+        effect_variables: List[Union[Variable, str]],
+        adjustment_variables: Optional[List[Union[Variable, str]]] = None,
+        adjustment_region_count_warning_threshold: int = DEFAULT_ADJUSTMENT_REGION_COUNT_WARNING_THRESHOLD,
```
tomsch420: Same as above

probabilistic_model/src/probabilistic_model/probabilistic_circuit/rx/probabilistic_circuit.py
```diff
@@ -286,7 +286,7 @@ def is_leaf(self):
 
     @property
     def leaves(self) -> List[LeafUnit]:
-        return []
+        return [self]
```
tomsch420, line 289: it do be like that :+1:

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/rspn.py
```diff
@@ -49,10 +53,46 @@
...
+
+class GroundingMode(enum.Enum):
+    """
+    Selects how ``RelationalProbabilisticCircuit.ground`` represents an exchangeable
+    relation's aggregation latents that a query leaves undetermined.
+
+    Undetermined latents are always retained as variables of the grounded circuit --
+    never integrated out -- so any caller that only wants the query's own variables
+    marginalizes the ones it does not need afterward, as a postprocessing step, rather
+    than grounding deciding that for it. Registering a retained latent as a cause or
```
tomsch420: this is a general true statement without the causal connection here. Just make this about the way aggregations are grounded here, not about causality.

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/rspn.py
```diff
@@ -49,10 +53,46 @@
...
     Unit,
+    leaf,
 )
 from random_events.interval import Interval
 from random_events.variable import Variable
 
+logger = logging.getLogger(__name__)
+
+
+class GroundingMode(enum.Enum):
```
tomsch420: also if you use auto, this is an IntEnum

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/rspn.py
```diff
@@ -359,28 +399,26 @@ def _condition_class_circuit(
...
         conditioning_result, _ = circuit.log_conditional_in_place(
             aggregation_statistics
         )
         if conditioning_result is None:
             circuit = self.class_probabilistic_circuit.__deepcopy__()
-            product_nodes_to_extend = find_lowest_product_nodes_that_model_variables(
-                circuit, SortedSet(latent_variables)
-            )
         if len(circuit.nodes()) == 0:
             raise ValueError("The grounding of the class failed.")
```
tomsch420: custom exception here

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/rspn.py
```diff
@@ -464,6 +511,27 @@ def _ground_exchangeable_part(
             )
             return circuit
 
+        if grounding_mode is GroundingMode.EXACT:
```
tomsch420: too deep nesting. Make to methods, ground_monte_carlo and ground_exact

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/rspn.py
```diff
@@ -640,6 +712,222 @@ def _attach_monte_carlo_mixture(
...
+        )
+        wrapper = ProductUnit(probabilistic_circuit=circuit)
+        wrapper.add_subcircuit(instance_root)
+        for variable in undetermined_latents:
+            wrapper.add_subcircuit(
+                leaf(make_dirac(variable, assignment[variable]), circuit)
+            )
+        return wrapper
+
+    def _attach_exact_partition_mixture(
```
tomsch420: can you give me some plot or math or so that explains me why this is possible?

tomsch420: if so, you could also start working on a doc page about how causal reasoning works in math in the probabilistc model package. Basically summarizing the literature there and showing how your stuff implements different parts of the literature. This is very useful for your (and my) scientific understanding and also is something you have to do anyways for your thesis at some point

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/rspn.py
tomsch420, line 1: there are many many static methods in here. Perhaps its then time for some other class where you actually need fields instead of passing the eversame arguments around?

tomsch420 (changes requested):

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/causal.py
Narenvasant: done

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/causal.py
Narenvasant: both are used, but now I kept accepting both, just gave the Union[Variable, str] a name (VariableReference) instead of repeating it three times.

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/causal.py
Narenvasant: done

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/causal.py
Narenvasant: done

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/rspn.py
Narenvasant: alright, done

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/rspn.py
Narenvasant: done

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/rspn.py
Narenvasant: done

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/rspn.py
Narenvasant: created a new doc causal_reasoning explaining pearls do calculus, and also relational causal circuits

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/rspn.py
Narenvasant: done

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/rspn.py
```diff
@@ -140,6 +179,384 @@ def ground(
...
+    """
+    The class circuit's product nodes that mount the grounded instance(s).
+    """
+
+    template: ExchangeableDistributionTemplate
+    """
+    The fitted template for this exchangeable relation.
+    """
+
+    query_parts: list
```
davidprueser: list of what

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/rspn.py
```diff
@@ -140,6 +179,384 @@ def ground(
...
+    """
+    The working class circuit being extended.
+    """
+
+    product_nodes_to_extend: list[ProductUnit]
+    """
+    The class circuit's product nodes that mount the grounded instance(s).
+    """
+
+    template: ExchangeableDistributionTemplate
```
tomsch420: do you need exchangability here? The future of this collection grounding is in a more abstract class. Check if you can rely on the superclass of ExchangeableDistributionTemplate instead

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/rspn.py
tomsch420: List of what?

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/rspn.py
```diff
@@ -140,6 +179,384 @@ def ground(
...
+    @staticmethod
+    def _undetermined_latents_partition_disjointly(
+        proposal: ProbabilisticCircuit,
+    ) -> bool:
+        """
+        Check whether ``proposal`` is a genuine, pairwise-disjoint partition over the
+        undetermined latents: a mixture of at least two branches, no two of which
+        overlap.
+
+        ``JointProbabilityTree`` does not retain which variables it actually split on
```
tomsch420: dont write something about JPT here, keep the docstring short and precise

probabilistic_model/doc/causal_reasoning.md
```diff
@@ -0,0 +1,256 @@
...
+writes this as $P(Y \mid do(X = x))$: the distribution of effect $Y$ once cause $X$ is set
+to $x$ by intervention, cutting $X$ off from whatever normally determines it. This differs
+from conditioning, $P(Y \mid X = x)$, whenever $X$ and $Y$ share a common cause: observing
+$X = x$ also tells you something about that common cause, which then leaks into what you
+infer about $Y$; intervening on $X$ does not.
+
+````{prf:definition} Backdoor Criterion
+:label: def-backdoor-criterion
+
+A set of variables $Z$ satisfies the backdoor criterion relative to an ordered pair
```
tomsch420, line 59: this needs first a definition of the graph you operate on (i think the MDVtree

probabilistic_model/doc/causal_reasoning.md
```diff
@@ -0,0 +1,256 @@
...
+
+## Relational causal reasoning
+
+The circuits above assume a fixed variable set, decided once and for all when the circuit
+is fitted. That does not hold for a scene with an unknown number of objects: how many
+objects there are is itself a fact you might want to intervene on or condition on, but no
+single flat circuit has "object count" as one of its variables ahead of time -- the variable
+set depends on which scene is being described.
+
+`RelationalProbabilisticCircuit` ({cite}`nath2015rspn`, "Relational Sum-Product Networks",
```
tomsch420, line 208: cite KRRUESER by Prüser instead

probabilistic_model/doc/causal_reasoning.md
```diff
@@ -0,0 +1,256 @@
...
+bridges into `krrood`) resolves this by *grounding*: given a query describing one concrete
+scene, it stamps out one instance of a fitted template circuit per object the query
+mentions, and combines the instances into a single circuit over that scene's actual
+variables. Where the query leaves some *aggregation statistic* undetermined -- e.g. it asks
+about "a room with some chairs" without saying how many -- grounding must decide how many
+child instances to stamp out and how to combine them, which requires resolving the
+statistic to a concrete value or distribution over values (the "undetermined latent") before
+grounding the exchangeable part.
+
+Prior to this package, that undetermined latent was resolved by marginalizing it out
```
tomsch420: this is developer notes, it doesnt belong here. Just write factual about the current state

probabilistic_model/doc/causal_reasoning.md
```diff
@@ -0,0 +1,256 @@
...
+The hypothesis -- that the $a_i$ are actually pairwise disjoint -- is where the two modes
+differ, and it is not free for the exact-partition case:
+
+- **`SAMPLED`** deduplicates its Monte-Carlo draws before grounding
+  (`_sample_undetermined_latents`), so the $a_i$ are distinct points -- trivially pairwise
+  disjoint regardless of what the underlying distribution over $L$ looks like. This is why
+  `SAMPLED` always succeeds: disjointness of single points needs no assumption about the
+  fitted model.
+- **`EXACT`** instead needs the fitted `JointProbabilityTree` {cite}`nyga2023joint` to have
+  *actually split* on $L$: a JPT is only guaranteed deterministic
```
tomsch420: dont talk about JPTs. this is general for probabilistic circuits

probabilistic_model/doc/causal_reasoning.md
```diff
@@ -0,0 +1,256 @@
...
+
+No prior work was found combining md-vtree-style circuit causal inference with relational
+grounding directly; {cite}`luttermann2024lifted` is the closest bridge, but on a different
+substrate (parametric factor graphs, not circuits) with its own, differently scoped
+tractability class (lifted over object symmetry, rather than exact per-query grounding).
+This package stays fully grounded per query rather than lifted: it answers exact causal
+queries at RSPN's existing per-query grounding granularity, not population-scale inference
+over many symmetric objects at once.
+
+## Open question: does the interventional distribution match reality?
```
tomsch420: this section is not needed

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/rspn.py
```diff
@@ -140,6 +179,384 @@ def ground(
...
+            raise UndeterminedLatentsNotModeledError(list(self.undetermined_latents))
+        if not self._undetermined_latents_partition_disjointly(proposal):
+            raise UndeterminedLatentsNotPartitionedError(
+                list(self.undetermined_latents)
+            )
+
+        # the precondition just verified guarantees proposal.root is a SumUnit with at
+        # least two branches
+        branches = proposal.root.log_weighted_subcircuits
+        _ = proposal.support
```
davidprueser: Why is that here?

probabilistic_model/doc/causal_reasoning.md
tomsch420, line 1: i think here it would be cool to really look into the math and show a small example with numbers and explain and show how your stuff transforms the circuits structure into the causal structure for a query

tomsch420 (changes requested):

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/rspn.py
```diff
@@ -451,223 +878,60 @@ def _ground_exchangeable_part(
...
+        try:
+            grounder.attach_exact_partition_mixture()
+            return circuit
+        except UndeterminedLatentsNotPartitionedError:
+            logger.warning(
+                "Exact-partition grounding for latents [%s] is not support-"
+                "deterministic; falling back to GroundingMode.SAMPLED.",
+                ", ".join(variable.name for variable in grounder.undetermined_latents),
             )
+            return self._ground_monte_carlo(grounder, circuit)
```
davidprueser, line 919: Feels a bit inconsistent here. In the try-block, you just call functionality on the grounder directly, but in the except-block you extract similar stuff into a separate function

davidprueser: Ah I see why you did it, the function is called somewhere else as well. I still think it is more readable if you just put the two-liner at each place instead of extracting it into a separate function, but I guess that is just personal preference

davidprueser (changes requested):
I like. Good work!

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/rspn.py
Narenvasant: done

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/rspn.py
Narenvasant: list[Match]. changed

probabilistic_model/doc/causal_reasoning.md
Narenvasant: done

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/rspn.py
Narenvasant: done

probabilistic_model/doc/causal_reasoning.md
Narenvasant: added this, check if its correct for citation, i am not sure

probabilistic_model/doc/causal_reasoning.md
Narenvasant: added a small example

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/rspn.py
Narenvasant: changed

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/rspn.py
Narenvasant: yeah i see. i left it same for now XD

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/rspn.py
Narenvasant: you're right. its not needed. removed it

probabilistic_model/doc/causal_reasoning.md
```diff
@@ -0,0 +1,308 @@
...
+The backdoor criterion below is stated over a causal graph: a directed acyclic graph
+$G = (V, E)$ whose edges represent direct causal influence, with $X$, $Y$, and every
+candidate adjustment variable among its nodes $V$. This package neither constructs nor
+verifies $G$: registering `causal_variables`, `effect_variables`, and an adjustment set on
+a `CausalCircuit` is the analyst's assertion, from domain knowledge, that such a $G$ exists
+and that the chosen adjustment set satisfies the backdoor criterion relative to it. What the
+package verifies instead is the circuit-side precondition the next section's polytime
+algorithm depends on -- support determinism -- which is a property of the probabilistic
+circuit itself, not of $G$; it is unrelated to the vtree used later for tractability, and to
+the `MarginalDeterminismTreeNode` structure introduced below, which only groups query
```
tomsch420: where formal definition?

probabilistic_model/doc/causal_reasoning.md
```diff
@@ -0,0 +1,308 @@
...
+randomized training data), this collapses to $P(Y \mid do(X = x)) = P(Y \mid X = x)$: the
+interventional and observational distributions coincide.
+
+## Tractable causal inference on probabilistic circuits
+
+Evaluating the backdoor formula requires computing $P(Y \mid X = x, Z = z)$ for every value
+$z$ in the adjustment set's domain, and summing. On an arbitrary joint distribution this can
+be as expensive as the domain of $Z$ is large. {cite}`wang2023compositional` shows that on a
+*structured-decomposable* probabilistic circuit -- one where every product unit's children
+partition the scope along a shared, recursively fixed variable tree (a vtree) -- backdoor
```
tomsch420: no abbrivations

probabilistic_model/doc/causal_reasoning.md
```diff
@@ -0,0 +1,308 @@
...
+A circuit is support-deterministic over $V$ if every sum unit that splits on any variable
+in $V$ is support-deterministic over $V$.
+````
+
+Support determinism over the cause and adjustment variables means each sum unit's children
+already partition the world by the value those variables take -- so answering "what is
+$P(Y \mid X = x, Z = z)$ under this branch" never requires mixing across branches, and the
+backdoor sum becomes a single weighted pass over the circuit rather than one circuit
+evaluation per $(x, z)$ pair. {cite}`wang2023compositional` formalizes the circuit family
+that guarantees this (*md-vtrees*, a generalization of probabilistic sentential decision
```
tomsch420: no abbrivations

tomsch420: also reference to PSDD paper?

probabilistic_model/doc/causal_reasoning.md
tomsch420: @davidprueser

tomsch420 (changes requested):

probabilistic_model/doc/causal_reasoning.md
Narenvasant: done

probabilistic_model/doc/causal_reasoning.md
```diff
@@ -67,11 +76,22 @@ interventional and observational distributions coincide.
...
-*structured-decomposable* probabilistic circuit -- one where every product unit's children
-partition the scope along a shared, recursively fixed variable tree (a vtree) -- backdoor
-adjustment is tractable in the size of the circuit whenever the circuit satisfies one
-additional structural property for the relevant variables: **support determinism**.
+be as expensive as the domain of $Z$ is large.
+
+````{prf:definition} Vtree
+:label: def-vtree
+
+A vtree over a variable set $X$ is a full, rooted binary tree whose leaves are in
```
tomsch420: no abbrivations

tomsch420 (changes requested):

davidprueser (approved):

tomsch420 (approved):

PR #599, opened 2026-09-04
Underspecified node real and sim compatible

LucaKro:
UnderspecifiedNode used to execute each grounded candidate directly against the plan's world, so a candidate that partially ran and then failed left its changes behind to poison every later attempt. 

Candidates are now rehearsed first against a disposable copy under a forced SIMULATED execution, and only one that survives is attached to the plan and executed for real. A real failure is deliberately left in place, since reality cannot be rolled back, and the next candidate rehearses from that new state. One copy serves every candidate: it is rolled back to the model version it was at and its state restored between attempts, and replaced only when the real world has itself moved on or when a BrokenWorldModificationHistoryError leaves it unrecoverable. Since the copy is never synchronized, nothing a rehearsal does is published.

Supporting changes: 
- World.rebind_world_entities re-points an object graph at a given world's own entities, which model modification requires
-  PlaceAction may now measures the grasp from the attachment (GraspDescription.from_attachment) rather than reading the preceding PickUpAction's intent, so it also works when placing is underspecified.


coraplex/src/coraplex/datastructures/grasp.py
LucaKro, line 1: fyi, most of this will be removed in a followup PR

coraplex/src/coraplex/plans/plan_node.py
```diff
@@ -392,6 +397,136 @@ class ExecutionBoundaryNode(ABC, PlanNode):
     """
 
 
+@dataclass
+class CandidateRehearsal:
```
Tigul: Might be nitpicky but shouldn't this be a noun

coraplex/src/coraplex/plans/plan_node.py
```diff
@@ -392,6 +397,136 @@ class ExecutionBoundaryNode(ABC, PlanNode):
...
+        and longer run of blocks, most of them already-undone ones.
+
+        :param action: The grounded action to rehearse.
+        :return: True if `action` runs to completion without raising a `PlanFailure`.
+        """
+        from coraplex.plans.plan import Plan
+
+        world = self._copy()
+        plan = Plan(context=self._context)
+        candidate = ActionNode(designator=world.rebind_world_entities(action))
```
Tigul: copying and then rebinding  would be cleaner in this case

coraplex/src/coraplex/plans/plan_node.py
```diff
@@ -392,6 +397,136 @@ class ExecutionBoundaryNode(ABC, PlanNode):
...
+        plan.add_node(candidate)
+        version = world.get_world_model_manager().version
+
+        with world.reset_state_context(), ExecutionEnvironment(
+            ExecutionType.SIMULATED,
+            collision_avoidance=GiskardExecutable.collision_avoidance,
+        ):
+            try:
+                candidate.notify()
+                candidate.parse().execute()
```
Tigul: use perform

coraplex/src/coraplex/plans/plan_node.py
```diff
@@ -392,6 +397,136 @@ class ExecutionBoundaryNode(ABC, PlanNode):
...
+                return True
```
Tigul: Also evaluate the post_condition

coraplex/src/coraplex/plans/plan_node.py
```diff
@@ -392,6 +397,136 @@ class ExecutionBoundaryNode(ABC, PlanNode):
...
+        :return: The copy to rehearse against, taken again if `context.world` has
+            changed since the current one was made.
+        """
+        versions = (
+            self.context.world.get_world_model_manager().version,
+            self.context.world.state.version,
+        )
+        if self._world is None or self._source_versions != versions:
+            self._world = deepcopy(self.context.world)
+            self._context = replace(
```
Tigul: TIL

Tigul (changes requested):

Tigul:
Good Job

coraplex/src/coraplex/plans/plan_node.py
LucaKro: is it not?

coraplex/src/coraplex/plans/plan_node.py
LucaKro: hmm why do you think so? i dont think it would be cleaner, because right now World.rebind_world_entities is technically a "pure" function, and doing it the way you propose would mean the method works in-place, which leaves open more possibilities for bugs imo.

coraplex/src/coraplex/plans/plan_node.py
LucaKro: is it not already, since the context should have "evaluate_conditions" set to true? just to be sure I now pass evaluate_conditions=True to the new context

coraplex/src/coraplex/plans/plan_node.py
```diff
@@ -392,6 +397,138 @@ class ExecutionBoundaryNode(ABC, PlanNode):
...
+        """
+        Run `action` against the disposable copy of the world.
+
+        :param action: The grounded action to rehearse.
+        :return: True if `action` runs to completion without raising a `PlanFailure`.
+        """
+        try:
+            return self._attempt(action)
+        except BrokenWorldModificationHistoryError:
+            # The attempt failed inside an open modification block, so what it already
```
tomsch420: i feel like if this happens its a legit bug that gets swallowed here

coraplex/src/coraplex/plans/plan_node.py
```diff
@@ -392,6 +397,138 @@ class ExecutionBoundaryNode(ABC, PlanNode):
...
+
+        The version to roll back to is read here rather than when the copy is taken, so
+        each attempt undoes only its own modifications. Reverting is itself recorded, so
+        rolling every attempt back to where the copy started would mean undoing a longer
+        and longer run of blocks, most of them already-undone ones.
+
+        :param action: The grounded action to rehearse.
+        :return: True if `action` runs to completion without raising a `PlanFailure`.
+        """
+        from coraplex.plans.plan import Plan
```
tomsch420: 🚨 🚨 🚨

coraplex/src/coraplex/plans/plan_node.py
```diff
@@ -392,6 +397,138 @@ class ExecutionBoundaryNode(ABC, PlanNode):
     """
 
 
+@dataclass
+class CandidateRehearsal:
```
tomsch420: i dont like the name rehearsal. I had to google what that means and i think there are easier to understand words like ActionTrial or so

coraplex/src/coraplex/exceptions.py
```diff
@@ -329,4 +330,30 @@ def error_message(self) -> str:
...
+    Raised when a grasp should be read off a body that no end effector is holding.
+    """
+
+    body: KinematicStructureEntity
+    """
+    The body that was expected to be held.
+    """
+
+    end_effector: SemanticAnnotation
+    """
```
tomsch420, line 356: is this type hint correct?

semantic_digital_twin/src/semantic_digital_twin/spatial_types/spatial_types.py
```diff
@@ -1943,6 +1943,19 @@ def dot(self, other: Quaternion) -> sm.Scalar:
             return sm.Scalar(ca.mtimes(self.casadi_sx.T, other.casadi_sx))
         return NotImplemented
 
+    def rotational_error(self, other: Quaternion) -> sm.Scalar:
```
tomsch420: shouldnt this be rotational_distance? Error sounds like something is wrong, which is context dependent outside of this function the case, but this is just distance

semantic_digital_twin/src/semantic_digital_twin/exceptions.py
```diff
@@ -400,6 +401,31 @@ def suggest_correction(self) -> str:
         return "did you forget to call self.delete_orphaned_dofs()?"
 
 
+@dataclass
+class WorldEntityBelongsToAnotherWorld(WorldValidationError):
+    """
+    Raised when a world's own lookup returns an entity that belongs to another world.
+    """
+
+    world_entity: WorldEntityWithID
```
tomsch420, line 454: why not type WorldEntity?

tomsch420 (changes requested):

coraplex/src/coraplex/plans/plan_node.py
LucaKro: hmm yea fair point, i guess we just fail so we fix the bug?

semantic_digital_twin/src/semantic_digital_twin/spatial_types/spatial_types.py
LucaKro: the name is copied from RotationMatrix, where @ichumuh named it rotational_error iirc

if you think it should be changed you should talk to him, i dont mind either way.

coraplex/src/coraplex/exceptions.py
LucaKro: fixed

semantic_digital_twin/src/semantic_digital_twin/exceptions.py
LucaKro: Because normal world entites are not looked up like where the exception is used. but i can see if i can make this more clear

semantic_digital_twin/src/semantic_digital_twin/spatial_types/spatial_types.py
LucaKro: im changing it to rotational_distance for now, bc i think i agree. if @ichumuh disagrees ill change it back

coraplex/src/coraplex/plans/plan_node.py
tomsch420: yes

semantic_digital_twin/src/semantic_digital_twin/exceptions.py
tomsch420: I see what you mean. Imo this error semantically means you were doing something where you assumed a world entity you got belongs to the world you are working with but it isnt. This can be checked whenever a world backreference is used, so world_entity. In your case its only by ID, but the semantics of the exception do not change if its noticed by backreference. Anyways, perhaps yagni

tomsch420 (approved):

sunava (approved):

coraplex/src/coraplex/plans/plan_node.py
Tigul: Ok true sounded like an adjective first. However, I'm not a huge fan of the Rehersal name, maybe rename it to ActionProspection because that is what we usually use to describe that process?

coraplex/src/coraplex/plans/underspecified.py
```diff
@@ -0,0 +1,282 @@
...
+    from coraplex.datastructures.dataclasses import Context
+    from coraplex.robot_plans.actions.base import ActionDescription
+    from semantic_digital_twin.world import World
+
+
+# %% trying a grounded action out before it is executed for real
+
+
+@dataclass
+class ActionTrial:
```
Tigul, line 28: Prospection?

coraplex/src/coraplex/plans/underspecified.py
```diff
@@ -0,0 +1,282 @@
...
+    context: Context
+    """
+    The context the candidates were grounded in.
+
+    Only ever read from: a trial never mutates it or the world it points at, and the
+    candidates themselves are left untouched too, so they can still be attached and
+    executed for real afterwards.
+    """
+
+    _world: Optional[World] = field(default=None, init=False, repr=False)
```
Tigul: Why are these duplicated if we have the context?

Tigul (changes requested):

coraplex/src/coraplex/plans/underspecified.py
LucaKro: no i think i prefer how it is rn. Id say the whole process could be called prospection, but within the prospection you do multiple action trails

semantic_digital_twin/src/semantic_digital_twin/exceptions.py
LucaKro: i think yagni for now

Tigul (approved):

PR #600, opened 2026-09-04
Lidar annotation

Tigul:
# Description 
Adds a lidar annotation to the robots we currently use: 
    * Pr2
    * HSRB
    * Tiago 
    * Stretch 

Values for the scans are taken from the real robots


semantic_digital_twin/src/semantic_digital_twin/adapters/ros/__init__.py
LucaKro, line 1: wait how do i not know this is not empty 😭

semantic_digital_twin/src/semantic_digital_twin/robots/robot_part_mixins.py
```diff
@@ -298,3 +322,24 @@ class HasNeck(Generic[TGenericNeck], SubClassSafeGeneric, RobotPartMixin, ABC):
 
     def validate(self):
         assert self.neck is not None, f"Expected neck, got None"
+
+
+@dataclass(eq=False)
+class HasLaser(Generic[TGenericLaser], SubClassSafeGeneric, RobotPartMixin, ABC):
```
LucaKro, line 328: maybe Lidar?

semantic_digital_twin/src/semantic_digital_twin/adapters/ros/laser.py
```diff
@@ -0,0 +1,108 @@
...
+from semantic_digital_twin.datastructures.scan_pattern import ScanPattern
+from semantic_digital_twin.exceptions import NoLaserScanReceived, UselessConceptError
+from semantic_digital_twin.robots.robot_parts import Laser
+from semantic_digital_twin.world_description.world_entity import (
+    KinematicStructureEntity,
+)
+
+
+@dataclass(eq=False)
+class SubscribedLaser(Laser):
```
LucaKro, line 24: hmm i think this and the subscribed laser should be in the same package. I think this should move to the adapters sensors package

semantic_digital_twin/src/semantic_digital_twin/adapters/sensors/lidar.py
```diff
@@ -0,0 +1,68 @@
...
+import math
+from abc import ABC
+from dataclasses import dataclass
+
+import numpy as np
+from typing_extensions import List
+
+from semantic_digital_twin.datastructures.joint_state import JointState
+from semantic_digital_twin.datastructures.laser_reading import LaserReading
+from semantic_digital_twin.robots.robot_parts import Laser
```
LucaKro, line 12: I think the robot part should actually be defined in here. I think whoever inherits from HasLaser should only ever inherit from simulated or real laser.

semantic_digital_twin/src/semantic_digital_twin/robots/hsrb.py
```diff
@@ -411,7 +414,33 @@ def setup_default_configuration_in_world_below_robot_root(
 
 
 @dataclass(eq=False)
-class HSRBMobileBase(MobileBase[OmniDrive], HasTorso[HSRBTorso]):
+class HSRBBaseLaser(SimulatedLaser):
```
LucaKro, line 417: hmm now that i think of it, i dont think the split between simulated laser and real laser is good here. I think it should just be laser  in the adapter module, and then we have something concrete to inherit from. How would the HSR now get its laser data from the real laser? another subclass of the "real laser" class? and does the user then need to change the assignment of the laser to the base then?

I think having just one laser class that handles both the real and the simulated laser is more clean. could be sth if we provide a rosnode to the robot when creating it, then the laser subscribes and uses the real one, otherwise it takes the fake one.

or sth like that

LucaKro (changes requested):
nice, cool idea! 💪

semantic_digital_twin/src/semantic_digital_twin/adapters/ros/__init__.py
tomsch420, line 1: make this file empty

tomsch420 (commented):
One step closer to slam in SemDT 😎

semantic_digital_twin/src/semantic_digital_twin/robots/hsrb.py
tomsch420: I agree. i think the RealWorldLaser should be a process that just creates LaserReadings from the topic for a given Scanner. So it has a Laser and not Is a laser.

semantic_digital_twin/src/semantic_digital_twin/robots/robot_part_mixins.py
```diff
@@ -58,6 +54,34 @@ def validate(self):
...
+
+    def _narrowest_mixins(self) -> list[Type[RobotPartMixin]]:
+        """
+        :return: The mixins stating this part's assumptions, leaving out every mixin
+            another one of them narrows.
+        """
+        mixins = [
+            ancestor
+            for ancestor in type(self).__mro__
+            if issubclass(ancestor, RobotPartMixin) and "validate" in vars(ancestor)
```
tomsch420, line 75: if you depend on a validation method, make a class that explicitly declares a validation method and dont check for it via string

semantic_digital_twin/src/semantic_digital_twin/robots/robot_part_mixins.py
tomsch420: Lidar is technically an abbrivation 🤓 But I agree, we dont mean any kind of laser here, we mean specific lidar technology

tomsch420 (changes requested):

semantic_digital_twin/src/semantic_digital_twin/robots/robot_part_mixins.py
LucaKro: true, but actually i think that this may be one of the rare cases where "Lidar" is possibly more informative to its non abbreviated form. I dont think "Light Detection and Ranging" would tell someone who doesnt know what a Lidar is what it does. There "LaserScanner" would probably more helpful. But i do think Lidar is more accurate

semantic_digital_twin/src/semantic_digital_twin/robots/robot_part_mixins.py
LucaKro: true, but that already exists. RobotPartMixin is exactly that class

semantic_digital_twin/src/semantic_digital_twin/adapters/ros/__init__.py
ichumuh, line 1: empty when?

semantic_digital_twin/src/semantic_digital_twin/adapters/ros/laser.py
ichumuh, line 24: this looks a lot like the input synchronization in giskard. maybe that should be unified and moved to semdt/adapaters/ros2? probably makes more sense if semdt owns the joint state etc synchronization anyway.

semantic_digital_twin/src/semantic_digital_twin/robots/robot_part_mixins.py
ichumuh: we live in a lidar society

semantic_digital_twin/src/semantic_digital_twin/datastructures/laser_reading.py
```diff
@@ -0,0 +1,29 @@
...
+
+from dataclasses import dataclass, field
+
+from typing_extensions import List
+
+from semantic_digital_twin.spatial_types import Vector3
+
+
+@dataclass
+class LaserReading:
```
ichumuh, line 11: `LidarReading`? and shouldn't it have a scan pattern?

semantic_digital_twin/src/semantic_digital_twin/datastructures/laser_reading.py
```diff
@@ -0,0 +1,29 @@
...
+    """
+    One sweep of a laser.
+
+    Both lists hold one entry per beam and share their order, so ``direction[i]`` is the
+    beam that measured ``distance[i]``.
+    """
+
+    direction: List[Vector3] = field(default_factory=list)
```
ichumuh, line 19: i don't like list of vector3 here, they all have their own frames and construction overhead. use numpy or a new class which wraps symbolic math matrix

ichumuh: also, should be called `directions`

ichumuh: hm what do you even need it for? couldn't find any real uses

semantic_digital_twin/src/semantic_digital_twin/adapters/sensors/lidar.py
```diff
@@ -0,0 +1,68 @@
...
+            hits, so the closest one is picked explicitly.
+        """
+        distances = np.full(self.scan_pattern.beam_count, math.inf)
+        if len(index_ray) == 0:
+            return distances.tolist()
+
+        hit_distances = np.linalg.norm(points - world_P_laser[index_ray], axis=1)
+        farthest_first = np.argsort(hit_distances)[::-1]
+        distances[index_ray[farthest_first]] = hit_distances[farthest_first]
+        return distances.tolist()
```
ichumuh, line 68: why don't you return numpy?

semantic_digital_twin/src/semantic_digital_twin/datastructures/scan_pattern.py
```diff
@@ -0,0 +1,108 @@
...
+        """
+        return self.minimum_angle + np.arange(self.beam_count) * self.angle_increment
+
+    @property
+    def beam_directions(self) -> np.ndarray:
+        """
+        :return: A unit vector along every beam, one per row, in the scanner's frame.
+        """
+        angles = self.beam_angles
+        return np.column_stack((np.cos(angles), np.sin(angles), np.zeros_like(angles)))
```
ichumuh, line 96: duplication with directions in laserscan

semantic_digital_twin/src/semantic_digital_twin/datastructures/scan_pattern.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def beam_directions_in_frame(
+            self, reference_frame: Optional[KinematicStructureEntity]
+    ) -> List[Vector3]:
```
ichumuh, line 100: same comments as for directions above apply

semantic_digital_twin/src/semantic_digital_twin/robots/hsrb.py
ichumuh: what they said

ichumuh (changes requested):
when labels?

semantic_digital_twin/src/semantic_digital_twin/datastructures/laser_reading.py
Tigul: There is no uese-case yet, I want to use this for the elevator Navigate to check if a door is open or not but this will be a different PR.

semantic_digital_twin/src/semantic_digital_twin/robots/robot_part_mixins.py
Tigul: Every Robot Part has a validate method but previously they shadowed each other. This method is for checking the validate methods of the Parts that are used in this robot

LucaKro: so its to find the ones that override?

semantic_digital_twin/src/semantic_digital_twin/datastructures/laser_reading.py
LucaKro: i think then it should be removed until you actually need it. We dont want actively add dead code

ichumuh: yagni. when you actually need it, i'm pretty sure that this will not be the best way to achieve it :D will see when you work on it

Tigul: Do you mean the laser reading or the whole laser because the laser reading is actively used

semantic_digital_twin/src/semantic_digital_twin/adapters/ros/laser.py
Tigul: But a subscribed laser depends on ROS and I would keep everything that depends on ROS in one package

semantic_digital_twin/src/semantic_digital_twin/robots/hsrb.py
Tigul: I see where you are coming from and I partially disagree. I think having one Laser class is more messy than having to explicit ones because then we can't inject dependencies for the laser. 
The idea for a Real laser is either to use the Subscribed Laser or make another sub-class. That would be in line with how robots are modeled now. But maybe we have to discuss more how to do the architecture there also for how to hanlde cameras.

semantic_digital_twin/src/semantic_digital_twin/adapters/ros/laser.py
Tigul: My idea was to attach this in Giskard anyway.

semantic_digital_twin/src/semantic_digital_twin/adapters/ros/__init__.py
Tigul: The ROS2SemDT message converter relies on that, I can remove the imports here but that would mean rewriting the message converter which I would say is out of scope for this PR

semantic_digital_twin/src/semantic_digital_twin/datastructures/lidar_reading.py
```diff
@@ -0,0 +1,29 @@
...
+@dataclass
+class LaserReading:
+    """
+    One sweep of a laser.
+
+    Both lists hold one entry per beam and share their order, so ``direction[i]`` is the
+    beam that measured ``distance[i]``.
+    """
+
+    direction: List[Vector3] = field(default_factory=list)
```
ichumuh: laser reading is fine, but the laser direction vectors are yagni

Tigul: Yep they be gone

semantic_digital_twin/src/semantic_digital_twin/robots/hsrb.py
Tigul: Ok new structure, we have a lidar base class and the specific implementatios for each robot which have the lidar parameter. The lidar base class now has a source again which provides the readings, there are two factory methods to create the lidar either with the simulated source or with a real one

semantic_digital_twin/src/semantic_digital_twin/datastructures/laser_reading.py
Tigul: The lidar has the scan pattern, this is the raw output of one swwp

semantic_digital_twin/src/semantic_digital_twin/robots/robot_part_mixins.py
Tigul: No, it finds the mixins that are make up this mixin so that the validate methods can be explicitly run

ichumuh (approved):

semantic_digital_twin/src/semantic_digital_twin/adapters/ros/input_synchronization.py
```diff
@@ -289,13 +239,18 @@ class TfFrameSynchronizer(InputSynchronizer):
     Maps each tracked connection to its tf parent and child frame.
     """
 
+    node: Node = field(kw_only=True)
```
tomsch420, line 242: node appears very often as field in these. Move it up the abstraction level if possible

semantic_digital_twin/src/semantic_digital_twin/adapters/ros/msg_converter.py
```diff
@@ -52,55 +55,128 @@ def suggest_correction(self) -> str:
 
 
 @dataclass
-class Ros2ToSemDTConverter(ABC, Generic[InputType, OutputType]):
+class LaserScanBeamCountMismatch(ROS2ConversionError):
```
tomsch420, line 58: make its own exceptions module for this adapter

semantic_digital_twin/src/semantic_digital_twin/adapters/ros/msg_converter.py
```diff
@@ -52,55 +55,128 @@ def suggest_correction(self) -> str:
...
+    """
+    Base class for converters between ROS2 messages and their semDT representation.
 
-    If you want to add a new converter, subclass this class and override the convert
-    method. No registration is necessary.
+    If you want to add a new converter, subclass one of the two directions below and
+    override the convert method. No registration is necessary.
+    """
+
+    converter_module_suffix: ClassVar[str] = "_converters"
```
tomsch420, line 93: this sounds very smelly

semantic_digital_twin/src/semantic_digital_twin/adapters/sensors/lidar.py
```diff
@@ -0,0 +1,176 @@
...
+    @abstractmethod
+    def get_lidar_reading(self, lidar: Lidar) -> LidarReading:
+        """
+        :param lidar: The lidar whose beams are read.
+        :return: The most recent sweep of that lidar.
+        """
+
+
+@dataclass
+class SimulatedLidarSource(LidarSource):
```
tomsch420, line 40: thats pretty cool

semantic_digital_twin/src/semantic_digital_twin/adapters/sensors/lidar.py
```diff
@@ -0,0 +1,176 @@
...
+    """
+    A source that measures the world's collision geometry by casting a ray along every
+    beam of the lidar's scan pattern.
```
tomsch420, line 43: should noise be calibratable here?

semantic_digital_twin/src/semantic_digital_twin/robots/robot_part_mixins.py
tomsch420: i dont get why this is needed. This changes the semantic of subclassing something and looks smelly to me. The subclasses themselves should know what to call for validation

semantic_digital_twin/src/semantic_digital_twin/robots/robot_part_mixins.py
```diff
@@ -298,3 +322,24 @@ class HasNeck(Generic[TGenericNeck], SubClassSafeGeneric, RobotPartMixin, ABC):
...
+    Mixin class for robots or robot parts that have a lidar as their direct child.
+    """
+
+    lidar: TGenericLidar = field(default=None, kw_only=True)
+    """
+    The lidar attached to the robot part.
+    """
+
+    def validate(self):
+        assert self.lidar is not None, "Expected lidar, got None"
```
tomsch420, line 339: use proper exceptions

tomsch420 (changes requested):

semantic_digital_twin/src/semantic_digital_twin/robots/robot_part_mixins.py
Tigul: This is needed because every robot part declares their own validate method but a robot part can be made up of other robot parts so the validate methods shadow each other. This method finds the robot parts that make up this part so the validate methods can be called explicitly

semantic_digital_twin/src/semantic_digital_twin/adapters/ros/msg_converter.py
Tigul: What else do you suggest? The modules need to be imported somehow in order to find them, the alternative would be to always hand them over explicitly or import everything in the module which is also messy

semantic_digital_twin/src/semantic_digital_twin/adapters/sensors/lidar.py
Tigul: Noise could be interesting but then we can make this in another PR and also introduce noise for Joints

semantic_digital_twin/src/semantic_digital_twin/robots/robot_part_mixins.py
LucaKro: okay i think i understand, but isnt the canonical way to fix it just to call super().validate() everywhere?

semantic_digital_twin/src/semantic_digital_twin/adapters/ros/ros2_node.py
```diff
@@ -0,0 +1,19 @@
...
+
+from dataclasses import dataclass, field
+
+from rclpy.node import Node
+
+# %% the node everything talks over
+
+
+@dataclass(eq=False)
+class Ros2Node:
```
LucaKro, line 11: HasROS2Node

semantic_digital_twin/src/semantic_digital_twin/datastructures/lidar_reading.py
```diff
@@ -0,0 +1,38 @@
...
+    """
+    The frame the sweep was measured in.
+    """
+
+    scan_pattern: ScanPattern
+    """
+    The directions the sweep covered and the distances it could measure.
+    """
+
+    ranges: np.ndarray
```
LucaKro, line 32: use numpy typing

LucaKro (changes requested):
i dont understand how you will inject the real lidar when needed.

semantic_digital_twin/src/semantic_digital_twin/robots/robot_part_mixins.py
Tigul: But that wouldn't solve this case?

tomsch420: the subclasses themselves should decide which validations are taking into account. I am still against this change

semantic_digital_twin/src/semantic_digital_twin/adapters/ros/msg_converter.py
tomsch420: ormatic has a similar issue. Look how its done there and unify the ways of doing this

semantic_digital_twin/src/semantic_digital_twin/adapters/ros/ros2_node.py
tomsch420: it do be like that

tomsch420 (changes requested):

PR #601, opened 2026-09-04
Speed up Polytope inner/outer box approximation

tomsch420:
## Summary

`Polytope.outer_box_approximation()` and `Polytope.inner_box_approximation()` (`random_events/src/random_events/polytope.py`) were spending almost all their time in incidental overhead of the underlying `polytope` library rather than in the actual geometric algorithm:

- The `polytope` library falls back to `scipy.optimize.linprog` (HiGHS) for its internal LP solves unless `cvxopt`+GLPK is installed, and profiling showed the HiGHS Python wrapper's `check_option` alone accounted for ~60% of runtime on a representative case. `cvxopt` was already listed as an optional dev dependency but not actually installed.
- `outer_box_approximation()` decided whether to keep splitting a polytope via `bounding_box <= current_polytope`, which the `polytope` library answers with a full LP-based set-difference + volume computation (`is_subset` → `diff` → `region_diff` → `reduce`) — used just to ask "is this polytope already box-shaped?", a question a single matrix multiply can answer.
- Every box built via `from_box()` (in `maximum_inner_box` and `as_box_polytope`) recomputed its bounding box via `2 * n_dimensions` LP solves and its volume via randomized Monte-Carlo sampling on first access, even though both are already known exactly at construction time — this also made `inner_box_approximation`'s stopping criterion (`inner_box.volume < minimum_volume`) nondeterministic.
- `from_2d_points` hand-rolled a 2D-only convex-hull-to-inequalities loop, when `scipy.spatial.ConvexHull` already exposes exact facet equations (`hull.equations`) for any dimension.

## Changes

- Use the already-declared `cvxopt`/GLPK optional dependency (no code change, just ensure it's installed) instead of falling back to scipy's slower LP backend.
- Add `Polytope.is_box()`: a cheap 2**n_dimensions corner check against `A x <= b`, replacing the LP-based subset check in `outer_box_approximation`.
- Add `Polytope._box_polytope_from_bounds()`: pre-populates the `bounding_box`/`volume` caches on boxes built from known bounds, used in `maximum_inner_box` and `as_box_polytope`.
- Rebuild `maximum_inner_box`'s LP with OR-Tools' low-level `Constraint`/`Objective` API (`SetCoefficient`) instead of its operator-overloaded `LinearExpr` API (`sum(a * dimension_variables) + ...`), which spent most of its time in Python-level object construction rather than the LP solve. Same solver, no new dependency.
- Rename `from_2d_points` → `from_points`, generalized to any dimension via `ConvexHull.equations`; added a 3D test.
- Added `experiments/random_events_experiments/polytope_approximation_benchmark.py` (`PolytopeApproximationBenchmarkResult`, built on `ExperimentResult`/`ExperimentsTable`), a benchmark of 2D/3D polytopes (convex hulls of random point clouds, point counts up to 800 in 2D / 200 in 3D) used to measure `inner_box_approximation`/`outer_box_approximation`. Reports, per polytope: its facet count and volume, the minimum-volume threshold passed to both algorithms, and each algorithm's box count, volume diff (`inner_volume_diff`/`outer_volume_diff`: the gap between the polytope's volume and the volume covered by its box approximation — inner is a subset so it can only under-cover, outer is a superset so it can only over-cover) and duration. The reference volume these diffs are measured against is the *exact* volume from scipy's `ConvexHull` (deterministic simplicial decomposition) rather than `Polytope.volume` (a randomized Monte-Carlo estimate), so the diffs are attributable entirely to the box approximation's own error, not to noise in the reference.

## Benchmark

Baseline (scipy/HiGHS backend, before any of these changes):

```
dim  n_pts facets   volume min_vol_frac  max_inner_box[s]  inner_approx[s]  #boxes  outer_approx[s]  #boxes
-----------------------------------------------------------------------------------------------------------
  2     10      6    1.668         0.10            0.0023           0.1874       8           1.2233      13
  2     25     11    1.821         0.10            0.0009           0.1280       5           1.6916      12
  2     50     11    2.284         0.10            0.0011           0.1237       5           1.6575      12
  2    100     19    2.646         0.10            0.0016           0.1487       5           2.4187      12
  3     10     14    0.574         0.20            0.0014           0.0110       1           4.0186      15
  3     20     28    0.952         0.20            0.0023           0.3458       7           7.0212      15
  3     30     38    1.275         0.20            0.0030           0.4052       7           9.4168      14
```

After this PR, widened to more point counts per dimension, trimmed to only the columns that characterize the approximation itself (facet count, volume, the threshold actually passed to both algorithms, and each algorithm's box count / volume diff / duration), and with the volume-diff columns measured against the exact `ConvexHull` volume instead of `Polytope.volume`'s Monte-Carlo estimate (2D rows first, then 3D):

```
facets  volume  min_volume  inner_#boxes  inner_vol_diff  inner[s]  outer_#boxes  outer_vol_diff  outer[s]
-------------------------------------------------------------------------------------------------------
     6  1.5491      0.1549             8          0.3106    0.0169            17          0.3509    0.0106
    11  1.7563      0.1756             8          0.2336    0.0138            12          0.2348    0.0098
    11  2.3226      0.2323             5          0.2821    0.0071            12          0.2264    0.0097
    19  2.6237      0.2624             5          0.4945    0.0090            12          0.3885    0.0110
    16  2.8659      0.2866             5          0.4624    0.0085            12          0.4029    0.0105
    23  2.9755      0.2976             5          0.4599    0.0099            12          0.3797    0.0116
    32  3.0236      0.3024             5          0.4536    0.0129            12          0.4044    0.0132
    14  0.5550      0.1110             1          0.4447    0.0008            15          0.4250    0.0121
    28  0.9377      0.1875             7          0.5235    0.0200            15          0.6703    0.0146
    38  1.2379      0.2476             7          0.7053    0.0257            15          1.0018    0.0165
    46  1.9477      0.3895             7          1.0759    0.0312            16          1.3869    0.0193
    52  2.2731      0.4546             7          1.0441    0.0355            16          1.5063    0.0204
    90  2.3790      0.4758             7          1.1809    0.0835            16          1.5887    0.0275
   100  2.4774      0.4955             7          1.1202    0.0841            16          1.9265    0.0294
    92  2.8538      0.5708             7          1.2627    0.0736            16          2.1052    0.0280
```

`outer_box_approximation` is ~100-590x faster than the pre-PR scipy/HiGHS baseline (up to 9.4s → 0.03s on the largest 3D case now tested, which is larger than the original baseline's largest case); `inner_box_approximation` is ~5-12x faster, including a further 15-25% cut from the OR-Tools low-level rewrite on top of the earlier fixes (most visible on the small/cheap cases where `maximum_inner_box` is a bigger share of the total, e.g. 14 facets: 0.0015s → 0.0008s). `inner_vol_diff`/`outer_vol_diff` are consistently positive as expected (inner under-covers, outer over-covers), and — now that they're measured against an exact rather than a sampled reference volume — track facet/point count smoothly instead of jumping around with Monte-Carlo sampling noise, giving a reliable baseline for how tight these two approximations are at these `min_volume` thresholds, independent of this PR's performance changes.

Note: `inner_box_approximation`'s remaining cost is now dominated by `polytope.mldivide`'s facet-by-facet redundancy removal (`region_diff`/`reduce`) -- a further ~4-5x speedup is achievable there by reimplementing that operation on top of `pplpy` (Parma Polyhedra Library bindings), prototyped separately but not included here: it would add a hard GPL-3.0 runtime dependency (heavier than the optional `cvxopt`/GLPK backend above) plus float-to-exact-rational conversion machinery, for a further ~1.5-2x on `inner_box_approximation` specifically (`outer_box_approximation` wouldn't benefit at all). Not pursued in this PR pending a decision on whether that trade-off is worth it.

## Test plan

- [x] `pytest test/random_events_test/test_polytope.py -v` — 5/5 pass (4 existing + new `test_from_points_3d`)
- [x] `Polytope.is_box()` verified against the original LP-based subset check on 58 random 2D/3D polytopes (0 mismatches), including full `outer_box_approximation` box-count comparisons
- [x] Cached `bounding_box`/`volume` on boxes verified against independently-recomputed reference values on 20 random boxes
- [x] `python3 -m experiments.random_events_experiments.polytope_approximation_benchmark` runs and renders the Typst table, including the widened sweep and exact-volume diffs above
- [x] `maximum_inner_box`'s OR-Tools low-level rewrite checked against 40 random 2D/3D polytopes for box-in-polytope containment and positive box volume (0 mismatches)

Note: `cvxopt` (used for the optional GLPK LP backend) is GPL-3.0-licensed; `random_events` is LGPL-3.0. This is compatible as an optional dependency (the same pattern SciPy itself uses for optional GPL solvers), but flagging since it's a new copyleft dependency in the dependency graph when installed.

## Review feedback addressed

- Benchmark script moved into `experiments/random_events_experiments/` and rebuilt on `ExperimentResult`/`ExperimentsTable` (`TypstRenderer`) instead of an ad hoc dataclass with manual table formatting
- Array parameters use `numpy.typing.NDArray[np.float64]` instead of bare `np.ndarray`
- `Polytope.is_box()`'s independently-optional `lower`/`upper` bundled into a single `bounds` tuple parameter, matching `bounding_box`'s own return shape, so a caller can no longer pass just one and have it silently discarded
- Abbreviated names (`dim`, `n_points`, `n_facets`, `rng`) spelled out; the sweep config is now a `PolytopeApproximationBenchmarkSweep` dataclass instead of plain tuples
- Volume-diff reference switched from `Polytope.volume` (Monte-Carlo estimate) to an exact `ConvexHull`-based volume, and the sweep widened to more point counts per dimension

https://claude.ai/code/session_01KbE8FSNcU7G6XMxCYVjcNK


random_events/scripts/benchmark_polytope_approximation.py
tomsch420, line 1: put this into experiments

tomsch420: also use the proper epxeriments table subclass

tomsch420: Moved to `experiments/random_events_experiments/polytope_approximation_benchmark.py` and rebuilt on `ExperimentResult`/`ExperimentsTable` (rendered via `TypstRenderer`), matching the sibling scripts in that package.

random_events/src/random_events/polytope.py
```diff
@@ -37,31 +39,17 @@ def from_polytope(cls, polytope_: polytope.Polytope) -> Self:
         return cls(polytope_.A, polytope_.b)
 
     @classmethod
-    def from_2d_points(cls, points: np.ndarray) -> Self:
+    def from_points(cls, points: np.ndarray) -> Self:
```
LucaKro: numpy typing

random_events/src/random_events/polytope.py
```diff
@@ -97,12 +85,65 @@ def inner_box_approximation(self, minimum_volume: float = 0.1) -> Event:
...
+        machinery, which dominates the runtime of `outer_box_approximation`.
+
+        :param lower: The lower bounds of the bounding box. Computed from
+            `self.bounding_box` if not given.
+        :param upper: The upper bounds of the bounding box. Computed from
+            `self.bounding_box` if not given.
+        :param tolerance: The numerical tolerance for the inequality check.
+        :return: Whether this polytope equals its own bounding box.
+        """
+        if lower is None or upper is None:
```
LucaKro: i think this is sus, if they are dependent put them in a datastructure. why would one of them be none and not the other?

LucaKro (changes requested):
cool cool cool

experiments/src/experiments/random_events_experiments/polytope_approximation_benchmark.py
tomsch420, line 1: no abbrivations

experiments/src/experiments/random_events_experiments/polytope_approximation_benchmark.py
```diff
@@ -0,0 +1,203 @@
...
+        inner_volume_diff=volume - inner_event.size,
+        outer_box_approximation_duration=outer_box_approximation_duration,
+        outer_box_count=len(outer_event.simple_sets),
+        outer_volume_diff=outer_event.size - volume,
+    )
+
+
+def main():
+    configs = [
+        # (dim, [n_points...], min_volume_fraction)
```
tomsch420: this config should be a class

random_events/src/random_events/polytope.py
tomsch420: agree

random_events/src/random_events/polytope.py
tomsch420: I see. Claude, check if this can even happen. If yes, check if making a class here will slow down things

random_events/src/random_events/polytope.py
tomsch420: Fixed: array parameters (`from_points`, `_box_polytope_from_bounds`, `is_box`) now use `numpy.typing.NDArray[np.float64]` instead of bare `np.ndarray`.

random_events/src/random_events/polytope.py
tomsch420: Confirmed it couldn't currently happen (the only call site always passes both, straight from `self.bounding_box`), but the API allowed a caller to pass just one, which was silently discarded rather than raising. Bundled `lower`/`upper` into a single `bounds: Optional[Tuple[...]]` parameter matching `self.bounding_box`'s own `(lower, upper)` return shape, so that can no longer happen.

On the class-vs-tuple question: measured it directly. Two positional args vs. one tuple arg is ~76ns vs ~78ns per call in isolation (noise), and the full `is_box` call costs ~17us either way (dominated by the corner enumeration + matmul, not argument passing). Went with a plain tuple rather than a new class since it already matches `bounding_box`'s return type exactly, so it adds no new allocation or dependency.

experiments/src/experiments/random_events_experiments/polytope_approximation_benchmark.py
tomsch420: Renamed dim/n_points/n_facets/n/rng to number_of_dimensions/number_of_points/number_of_facets/random_number_generator throughout, matching the naming already used in sibling scripts (e.g. complement_worst_case_experiment.py's number_of_dimensions).

experiments/src/experiments/random_events_experiments/polytope_approximation_benchmark.py
tomsch420: Replaced the (dim, [n_points...], min_volume_fraction) tuples with a PolytopeApproximationBenchmarkSweep dataclass (number_of_dimensions, point_counts, min_volume_fraction).

random_events/src/random_events/polytope.py
```diff
@@ -97,12 +86,71 @@ def inner_box_approximation(self, minimum_volume: float = 0.1) -> Event:
...
         """
         :return: The polytope as box polytope.
         """
         lower, upper = self.bounding_box
-        return self.from_box([(a[0], b[0]) for a, b in zip(lower, upper)])
+        return self._box_polytope_from_bounds(lower, upper)
+
+    def is_box(
+        self,
+        bounds: Optional[
```
tomsch420: No funky tuples, move the is None check to the while loop and make it two arguments again

tomsch420: Reverted: `is_box(self, lower, upper, tolerance=1e-7)` is back to two required positional arguments, no tuple, no `Optional`/default. The only call site (`outer_box_approximation`'s while loop) already computes both unconditionally before calling it, so there's nothing to check there either.

random_events/src/random_events/polytope.py
```diff
@@ -185,33 +220,44 @@ def maximum_inner_box(self) -> Self:
...
+        objective = solver.Objective()
+        objective.SetCoefficient(scale, 1.0)
+        objective.SetMaximization()
 
         # create the guess for the r vector
         scale_of_box = maxima - minima
 
-        # create the matrix A^+
+        # create the matrix A^+ and, per facet, the scalar coefficient on `scale`
+        # (a_positive_row . scale_of_box)
```
tomsch420: if this is semantically hard to understand, wrap this part in a method and document the method (call it create constraints or so)

tomsch420: Extracted into `_create_maximum_inner_box_constraints()`, documented with the constraint it builds (Proposition 2) and the reason for the low-level Constraint API.

LucaKro (approved):

AbdelrhmanBassiouny (approved):

PR #603, opened 2026-09-07
Thread safe symbolic math

ichumuh:
for my casadi friend @AbdelrhmanBassiouny 

Makes symbolic math safe to use from more than one thread, after a native crash in concurrent forward-kinematics composition.

- **CasADi gets its own lock.** CasADi reference-counts shared expression nodes without atomics, so two threads building or copying *different* expressions still corrupt them and segfault. `CasadiLock` (`krrood/symbolic_math/symbolic_math.py`) is held by `SpatialType.__deepcopy__` and by FK expression composition.
- **`CompiledFunction.__call__` is serialised**, so binding arguments into and evaluating from the shared buffers is atomic to concurrent callers. `bind_args_to_memory_view()`/`evaluate()` stay lock-free for the single-threaded control loop.
- **Call sites that aliased `casadi_sx` now copy it** (`to_pose`, `to_generic_vector`, `from_iterable`, …), so two wrappers no longer share one container that `_verify_type()` writes into in place.
- **Memoization is rewritten as a `MemoizationCache`** (`krrood/patterns/caching.py`), replacing the implicit `__memo__` attribute: one documented object per owner, found in `vars(owner)` so a class no longer shares a cache with its subclasses, copying and pickling as empty, and locking per cache key rather than per owner. `memoize`/`copy_memoize`/`clear_memoization_cache` move out of `utils.py`.
- **Tests**: the segfault is reproduced in a subprocess (`-m slow`), plus per-key concurrency, cross-owner nesting, copy/pickle and cache-isolation tests. Full `krrood_test` (2294) and `semantic_digital_twin_test` (1407) pass.


krrood/src/krrood/utils.py
tomsch420, line 1: hmm there is alot of implicit interface assumptions in here with __memo__ and __memo__dict__. Is there some better way to do this?

tomsch420 (changes requested):

krrood/src/krrood/symbolic_math/symbolic_math.py
```diff
@@ -243,6 +244,20 @@ class CompiledFunction:
     Used to memorize if the result must be recomputed every time.
     """
 
+    _call_lock: threading.Lock = field(
```
LucaKro, line 290: reason for Lock over RLock?

krrood/src/krrood/utils.py
```diff
@@ -682,47 +689,87 @@ def _handle_import_from_node(
 
 TCallable = TypeVar("TCallable", bound=Callable[..., Any])
 
+_memo_lock_creation_lock = threading.Lock()
```
LucaKro: same here, why this over RLock?

LucaKro (changes requested):

krrood/src/krrood/utils.py
ichumuh: claude says you are very smart

krrood/src/krrood/symbolic_math/symbolic_math.py
ichumuh: claude says that there are no calls that reach out and require an RLock. if something actually does then it would be some weird fuckup that is worse to debug than a deadlock. and it can't think of a way to catch this that is better than a deadlock

krrood/src/krrood/utils.py
ichumuh: same as above

tomsch420 (approved):

LucaKro (approved):

PR #606, opened 2026-09-08
added tests and fixes for soft connection serialization

maltehue:
PiecewiseConstantCurvatureConnection and CosseratRodConnection did not survive serialization. Therefore I added:

  - New SoftConnection(Connection, ABC) owns segment_length and implements copy_for_world,
    copy_with_new_parent, to_json and _from_json, all driven by a deformation property each
    model extends with its own degrees of freedom.
  - SoftTrunk now ends both builders with world.add_semantic_annotation_recursively(trunk), as
    AbstractRobot does, registering the arm and end effector in dependency order. The arm was
    previously referenced but never registered, so that reference could not be resolved on
    the way back in.

### DIFF
--- a/semantic_digital_twin/src/semantic_digital_twin/datastructures/soft_trunk.py
+++ b/semantic_digital_twin/src/semantic_digital_twin/datastructures/soft_trunk.py
@@ -43,7 +43,9 @@ class SoftTrunkSection:
     """The rest length of the section in meters."""
 
     radius: float
-    """The radius of the cylinder representing the section's volume."""
+    """
+    The radius of the cylinder representing the section's volume.
+    """
 
     resolution: int
     """The number of discrete rigid segments used to approximate the continuous curve."""
@@ -51,7 +53,9 @@ class SoftTrunkSection:
 
 @dataclass(eq=False, kw_only=True)
 class SoftEndEffector(EndEffector):
-    """Concrete implementation of EndEffector for soft robots."""
+    """
+    Concrete implementation of EndEffector for soft robots.
+    """
 
     @classmethod
     def setup_default_configuration_in_world_below_robot_root(cls, robot_root):
@@ -66,7 +70,9 @@ class SoftEndEffector(EndEffector):
 
 @dataclass(eq=False, kw_only=True)
 class SoftArm(Arm):
-    """Concrete implementation of Arm for soft robots."""
+    """
+    Concrete implementation of Arm for soft robots.
+    """
 
     @classmethod
     def setup_default_configuration_in_world_below_robot_root(cls, robot_root):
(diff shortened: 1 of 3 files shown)


semantic_digital_twin/src/semantic_digital_twin/world_description/soft_connections.py
```diff
@@ -14,10 +19,101 @@
...
+    and serializing a connection both read.
+    """
+
+    segment_length: float = field(kw_only=True)
+    """
+    The length of backbone this connection spans while undeformed.
+    """
+
+    @property
+    def deformation(self) -> dict[str, Any]:
```
ichumuh: is this a to_json?

maltehue: yeah almost. Should i move it into a to_jsom directly instead of this abstraction?

ichumuh: if claude thinks that makes sense, or at least rename it to something json like, so ppl don't complain about the hardcoded str and stuff

semantic_digital_twin/src/semantic_digital_twin/datastructures/soft_trunk.py
tomsch420, line 1: there is a lot of "no abbrivations" and "obsessed with primitives" in here

semantic_digital_twin/src/semantic_digital_twin/world_description/soft_connections.py
```diff
@@ -14,10 +19,101 @@
     HomogeneousTransformationMatrix,
 )
 from semantic_digital_twin.world import World
+from typing_extensions import Any
+
+# %% what every soft connection has to say about itself
```
tomsch420: weird comment

semantic_digital_twin/src/semantic_digital_twin/world_description/soft_connections.py
tomsch420: i dont think we need these one liner methods, they can just be used in to/from_json

semantic_digital_twin/src/semantic_digital_twin/world_description/soft_connections.py
tomsch420, line 1: check if this is done by the to/from_json stuff without the explicit mentioning here

tomsch420 (changes requested):

semantic_digital_twin/src/semantic_digital_twin/datastructures/soft_trunk.py
ichumuh: there are some abbriviations, like in kappa_dof_id, but they are not newly introduced. where do you see obsessed with primitives? its creating json. also the scope of this pr was intended to fix the json serialization, not refactoring soft bodies. i would argue that once he uses proper to/from_json this pr is fine.

tomsch420: list[tuple[DegreeOfFreedom, ...]] is used twice and seems to be semantically important to me

tomsch420 (approved):

tomsch420 (changes requested):
i accidantly approved the wrong PR lol

semantic_digital_twin/src/semantic_digital_twin/datastructures/soft_trunk.py
ichumuh: fair but i don't think a PR for fixing json serialization should be forced to refactor soft bodies. the code is not getting worse if we keep this for now.

semantic_digital_twin/src/semantic_digital_twin/world_description/soft_connections.py
LucaKro: yea i think basically all of the added lines in this file can just be the normal to and from json

semantic_digital_twin/src/semantic_digital_twin/datastructures/soft_trunk.py
LucaKro: "leave each file you touch a bit cleaner than you found it" :D I think its just gonna be one datastructure no? in the same breath you tell claude to fix these comments it can also add the one datastructure i think

LucaKro: I think cleaning up a little bit of code surrounding the PR is good, there will almost never be dedicated cleanup PRs.

LucaKro (changes requested):

LucaKro (approved):

ichumuh (approved):

tomsch420 (approved):
very good , very clean

PR #609, opened 2026-09-09
orm-build auto now detects a failed import instead of outdated

Tigul:
# Describe 
Quick 5 min adventure that changes the auto behaviour of --orm-build to try to import the interfaces and only if they fail they are re-generated

### DIFF
--- a/cognitive_robot_abstract_machine/orm_import.py
+++ b/cognitive_robot_abstract_machine/orm_import.py
@@ -9,37 +9,77 @@ follow a failed import without leaving a stale interface behind in it.
 
 Started as ``python -P -m cognitive_robot_abstract_machine.orm_import <module>...`` by
 :mod:`cognitive_robot_abstract_machine.orm_interfaces`, which reads the report below back
-from this run's output.
+from this run's output. The report is part of that output rather than of this run's
+logging, which an interface can reconfigure while it is imported.
 """
 
 from __future__ import annotations
 
 import argparse
 import importlib
+import json
 import sys
 from dataclasses import dataclass
-from enum import IntEnum
+from enum import Enum, IntEnum, StrEnum
 
 from typing_extensions import Optional, Sequence, Tuple, Type
 
-LINE_PREFIX = "cram-orm-interface-stale "
-"""
-What marks a line of output as this run's report rather than as something an interface
-wrote while it was imported.
-"""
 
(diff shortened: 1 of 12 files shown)


cognitive_robot_abstract_machine/orm_import.py
```diff
@@ -0,0 +1,184 @@
...
+
+import argparse
+import importlib
+import sys
+from dataclasses import dataclass
+from enum import IntEnum
+
+from typing_extensions import Optional, Sequence, Tuple, Type
+
+LINE_PREFIX = "cram-orm-interface-stale "
```
tomsch420: grrr module level variables

cognitive_robot_abstract_machine/orm_import.py
```diff
@@ -0,0 +1,184 @@
...
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument(
+        "interfaces",
+        nargs="+",
+        help="The interface modules to import, in dependency order.",
+    )
+    stale = import_interfaces(parser.parse_args().interfaces)
+    if stale is None:
+        return
+    print(stale.to_line(), flush=True)
```
LucaKro, line 224: logging

cognitive_robot_abstract_machine/orm_interfaces.py
```diff
@@ -32,24 +38,19 @@
...
 """
-What a module of a package is named like.
+Name of the file holding it.
 """
 
-MAPPING_ENGINE_SOURCE_FOLDERS: Tuple[Path, ...] = (
-    Path("krrood") / "src" / "krrood",
-    Path("cognitive_robot_abstract_machine"),
-)
+INTERFACE_FOLDER_NAME = "orm"
```
LucaKro: make these strenum please

LucaKro (changes requested):

tomsch420 (changes requested):
this whole things seems to be a weird mix of module level variables. Isnt there any other way to do this with less shitty stuff?

Tigul:
In my defence the module level variables were there before

cognitive_robot_abstract_machine/orm_import.py
LucaKro: please use the logger

LucaKro (changes requested):

PR #610, opened 2026-09-09
[Giskard] Added client presence and watchdog

MideDay:
- Adds `client_presence.py`: a heartbeat mechanism where a client periodically announces it is still alive, plus a `ClientWatchdog` that Giskard uses to notice when the client that sent the running goal has gone quiet
- `MotionServer` now starts watching the goal's client when a goal begins and stops watching when it ends; if that client disconnects while its required world update is still being awaited, execution is stopped early with a new `ClientDisconnectedError` instead of running to completion or timing out for no one
- Adds `NoWatchedClientError` (`exceptions.py`) for the case where a presence check is queried before it has started watching a client
- Wires the watchdog through `control_loop`, `motion_goal`, `giskard`, and `python_interface` so clients can send heartbeats and Giskard can react to their absence
- Adds `test_client_presence.py` and extends `test_motion_server.py`, `test_motion_goal.py`, `test_integration_pr2.py`, and `test_world_updates.py` to cover the new behaviour

### DIFF
--- a/giskardpy/src/giskardpy/data_types/exceptions.py
+++ b/giskardpy/src/giskardpy/data_types/exceptions.py
@@ -5,7 +5,7 @@ from dataclasses import dataclass
 from krrood.exceptions import DataclassException
 
 
-class DontPrintStackTrace:
+class DoesntPrintStackTrace:
     """
     Marker mixin for exceptions whose stack trace should not be printed.
     """
--- a/giskardpy/src/giskardpy/middleware/ros2/client_presence.py
+++ b/giskardpy/src/giskardpy/middleware/ros2/client_presence.py
@@ -4,13 +4,15 @@ import json
 import time
 from abc import ABC, abstractmethod
 from dataclasses import dataclass, field
-from typing import Callable, ClassVar, Dict, List, Optional, Set
+from datetime import timedelta
+from typing import Callable, Dict, List, Optional
 
 import std_msgs.msg
 from rclpy.node import Node
 from rclpy.publisher import Publisher
 from rclpy.subscription import Subscription
 from rclpy.timer import Timer
+from sortedcontainers import SortedSet
 
 from giskardpy.middleware.ros2.exceptions import NoWatchedClientError
 from krrood.adapters.json_serializer import from_json, to_json
@@ -41,14 +43,9 @@ class ClientHeartbeatPublisher:
     Node name of the Giskard the heartbeat is meant for.
     """
 
-    period: float = 0.1
+    period: timedelta = timedelta(milliseconds=100)
     """
-    Seconds between two heartbeats.
-    """
-
-    topic_suffix: ClassVar[str] = "client_heartbeat"
-    """
-    What the heartbeat topic of a Giskard is called, relative to its node name.
+    Time between two heartbeats.
     """
 
     publisher: Publisher = field(init=False)
@@ -73,14 +70,14 @@ class ClientHeartbeatPublisher:
             topic=self.topic_name(self.giskard_node_name),
             qos_profile=10,
         )
-        self.timer = self.node.create_timer(self.period, self.publish)
+        self.timer = self.node.create_timer(self.period.total_seconds(), self.publish)
 
-    @classmethod
-    def topic_name(cls, giskard_node_name: str) -> str:
+    @staticmethod
+    def topic_name(giskard_node_name: str) -> str:
         """
         The topic the clients of the given Giskard announce themselves on.
         """
-        return f"{giskard_node_name}/{cls.topic_suffix}"
+        return f"{giskard_node_name}/client_heartbeat"
 
     def publish(self) -> None:
         """
@@ -160,9 +157,9 @@ class HeartbeatPresence(ClientPresence):
     Node of Giskard, which the heartbeat topic is named after.
     """
 
-    timeout: float = 1.0
+    timeout: timedelta = timedelta(seconds=1)
     """
-    Seconds without a heartbeat after which a client counts as gone.
+    Time without a heartbeat after which a client counts as gone.
 
     Heartbeats are received on a ros executor thread while the control loop runs on its
     own, so this has to stay well above the heartbeat period: a value close to it would
@@ -206,7 +203,7 @@ class HeartbeatPresence(ClientPresence):
         last_heartbeat = self.last_heartbeat.get(client)
         if last_heartbeat is None:
             return False
-        return self.clock() - last_heartbeat <= self.timeout
+        return self.clock() - last_heartbeat <= self.timeout.total_seconds()
 
     def start_watching(self, client: MetaData) -> bool:
         if not self.has_recent_heartbeat(client):
@@ -239,7 +236,7 @@ class GraphPresence(ClientPresence):
     Name of the action whose clients are watched.
     """
 
-    watched_endpoints: Set[bytes] = field(init=False, default_factory=set)
+    watched_endpoints: SortedSet[bytes] = field(init=False, default_factory=SortedSet)
     """
     The subscriptions the watched client had when its goal started.
 
@@ -254,18 +251,18 @@ class GraphPresence(ClientPresence):
         """
         return f"{self.action_name}/_action/feedback"
 
-    def subscribed_endpoints_of(self, client: MetaData) -> Set[bytes]:
+    def subscribed_endpoints_of(self, client: MetaData) -> SortedSet[bytes]:
         """
         The subscriptions the given client currently keeps on the feedback of the
         action.
         """
-        return {
+        return SortedSet(
             bytes(endpoint.endpoint_gid)
             for endpoint in self.node.get_subscriptions_info_by_topic(
                 self.feedback_topic
             )
             if endpoint.node_name == client.node_name
-        }
+        )
 
     def start_watching(self, client: MetaData) -> bool:
         endpoints = self.subscribed_endpoints_of(client)
@@ -277,7 +274,7 @@ class GraphPresence(ClientPresence):
 
     def stop_watching(self) -> None:
         super().stop_watching()
-        self.watched_endpoints = set()
+        self.watched_endpoints = SortedSet()
 
     def is_client_present(self) -> bool:
         return bool(self.watched_endpoints & self.subscribed_endpoints_of(self.client))
--- a/giskardpy/src/giskardpy/middleware/ros2/client_presence.py
+++ b/giskardpy/src/giskardpy/middleware/ros2/client_presence.py
@@ -1,0 +1,347 @@
+from __future__ import annotations
+
+import json
+import time
+from abc import ABC, abstractmethod
+from dataclasses import dataclass, field
+from typing import Callable, ClassVar, Dict, List, Optional, Set
+
+import std_msgs.msg
+from rclpy.node import Node
+from rclpy.publisher import Publisher
+from rclpy.subscription import Subscription
+from rclpy.timer import Timer
+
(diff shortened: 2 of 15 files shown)


giskardpy/src/giskardpy/middleware/ros2/client_presence.py
```diff
@@ -0,0 +1,347 @@
...
+    """
+
+    giskard_node_name: str
+    """
+    Node name of the Giskard the heartbeat is meant for.
+    """
+
+    period: float = 0.1
+    """
+    Seconds between two heartbeats.
```
tomsch420: timedelta

giskardpy/src/giskardpy/middleware/ros2/client_presence.py
```diff
@@ -0,0 +1,347 @@
...
+    """
+
+    topic_suffix: ClassVar[str] = "client_heartbeat"
```
tomsch420: this should not be a classvar

giskardpy/src/giskardpy/middleware/ros2/client_presence.py
```diff
@@ -0,0 +1,347 @@
...
+    observation. It also notices a client that is still running but no longer gets
+    around to announcing itself.
+    """
+
+    node: Node
+    """
+    Node of Giskard, which the heartbeat topic is named after.
+    """
+
+    timeout: float = 1.0
```
tomsch420: timedelta

giskardpy/src/giskardpy/middleware/ros2/client_presence.py
```diff
@@ -0,0 +1,347 @@
...
+    """
+    Node of Giskard, which sees the graph.
+    """
+
+    action_name: str
+    """
+    Name of the action whose clients are watched.
+    """
+
+    watched_endpoints: Set[bytes] = field(init=False, default_factory=set)
```
tomsch420: sets have the funny property that everytime you iterate over them they are in a different order. Is this something that may makes this flaky?

tomsch420 (changes requested):

giskardpy/src/giskardpy/middleware/ros2/client_presence.py
ichumuh: and i think 1 s default is enough

giskardpy/src/giskardpy/middleware/ros2/client_presence.py
ichumuh: haram

giskardpy/src/giskardpy/middleware/ros2/client_presence.py
MideDay: changed

giskardpy/src/giskardpy/middleware/ros2/client_presence.py
MideDay: changed

giskardpy/src/giskardpy/middleware/ros2/client_presence.py
MideDay: removed the topic_suffix outright and moved suffix topic_name method like in other publishers within cram

giskardpy/src/giskardpy/middleware/ros2/client_presence.py
MideDay: Shouldn't be a problem currently for bool() and & intersect calls, but changed to sortedSet in case of future changes that might cause problems with flakyness

giskardpy/src/giskardpy/middleware/ros2/client_presence.py
```diff
@@ -0,0 +1,344 @@
...
+    """
+    Identity of that client, the same one its goals name.
+    """
+
+    giskard_node_name: str
+    """
+    Node name of the Giskard the heartbeat is meant for.
+    """
+
+    period: timedelta = timedelta(milliseconds=100)
```
ichumuh: i think 1s i fine

giskardpy/src/giskardpy/middleware/ros2/exceptions.py
```diff
@@ -84,6 +85,48 @@ def suggest_correction(self) -> str:
         return ""
 
 
+@dataclass
+class ClientDisconnectedError(ExecutionException, DontPrintStackTrace):
```
AbdelrhmanBassiouny: The class name `DontPrintStackTrace` should be `DoesntPrintStackTrace`. I know you didn't create this class but since you used it, why not also fix it :D

AbdelrhmanBassiouny (changes requested):

giskardpy/src/giskardpy/middleware/ros2/exceptions.py
MideDay: Changed

giskardpy/src/giskardpy/data_types/exceptions.py
```diff
@@ -5,7 +5,7 @@
 from krrood.exceptions import DataclassException
 
 
-class DontPrintStackTrace:
+class DoesntPrintStackTrace:
```
tomsch420, line 8: this looks like it belongs in krrood as a bool field of dataclass exception. Mixins only for markers are 👃

giskardpy/src/giskardpy/middleware/ros2/client_presence.py
```diff
@@ -0,0 +1,344 @@
...
+    timeout: timedelta = timedelta(seconds=1)
+    """
+    Time without a heartbeat after which a client counts as gone.
+
+    Heartbeats are received on a ros executor thread while the control loop runs on its
+    own, so this has to stay well above the heartbeat period: a value close to it would
+    turn a busy executor into a stopped robot.
+    """
+
+    clock: Callable[[], float] = time.monotonic
```
tomsch420: is there a usecase for this being a different thing?

giskardpy/src/giskardpy/middleware/ros2/client_presence.py
```diff
@@ -0,0 +1,344 @@
...
+    """
+    Node of Giskard, which sees the graph.
+    """
+
+    action_name: str
+    """
+    Name of the action whose clients are watched.
+    """
+
+    watched_endpoints: SortedSet[bytes] = field(init=False, default_factory=SortedSet)
```
tomsch420: why is this bytes btw?

tomsch420 (changes requested):

giskardpy/src/giskardpy/middleware/ros2/client_presence.py
MideDay: The tests use it to use a SteppingClock, that allows for deterministic tests.

giskardpy/src/giskardpy/middleware/ros2/client_presence.py
MideDay: The endpoint.endpoint_gid returns a list[int], which is turned into bytes to make it work with set. Set is used here to make the condition in is_client_present simpler. Otherwise List[list[int]] could be used, but the logic in is_client_present would have to change, since lists aren't hashable. As such & operator can't be used.

giskardpy/src/giskardpy/data_types/exceptions.py
MideDay: That seems like something out of scope for this PR.

giskardpy/src/giskardpy/middleware/ros2/client_presence.py
MideDay: Changed the docstring to make the reason more obvious

giskardpy/src/giskardpy/middleware/ros2/client_presence.py
MideDay: Added to the docstring for clarity

giskardpy/src/giskardpy/data_types/exceptions.py
tomsch420: its only one prompt away

giskardpy/src/giskardpy/middleware/ros2/client_presence.py
```diff
@@ -0,0 +1,351 @@
...
+    """
+    The subscriptions the watched client had when its goal started.
+
+    A restarted client subscribes anew, so comparing the endpoints rather than the node
+    name keeps a namesake of the dead client from passing as the client that is gone.
+
+    ``rclpy`` reports each endpoint's id as a raw ``list[int]``, which cannot be a set
```
tomsch420: hmm i dont think set[bytes] is correct here. Should the rclpy lists not just be cast into tuples? This looks very unsemantic with bytes

tomsch420 (changes requested):

giskardpy/src/giskardpy/middleware/ros2/client_presence.py
```diff
@@ -0,0 +1,351 @@
...
+            return False
+        self.watched_client = client
+        return True
+
+    def is_client_present(self) -> bool:
+        return self.has_recent_heartbeat(self.client)
+
+
+@dataclass
+class GraphPresence(ClientPresence):
```
ichumuh: yagni

ichumuh (changes requested):

giskardpy/src/giskardpy/middleware/ros2/client_presence.py
```diff
@@ -0,0 +1,351 @@
...
+
+    @abstractmethod
+    def is_client_present(self) -> bool:
+        """
+        Whether the watched client is still connected.
+        """
+
+
+@dataclass
+class HeartbeatPresence(ClientPresence):
```
Tigul: Is there another use-case for ClientPresence? If not I don't see the need for an abstract base class

giskardpy/src/giskardpy/middleware/ros2/client_presence.py
Tigul: Shouldn't this be a field?

Tigul (changes requested):

giskardpy/src/giskardpy/data_types/exceptions.py
MideDay: will be done in a future PR

giskardpy/src/giskardpy/middleware/ros2/client_presence.py
MideDay: Removing the whole GraphPresence as it was deemed redundant

giskardpy/src/giskardpy/middleware/ros2/client_presence.py
MideDay: yep

giskardpy/src/giskardpy/middleware/ros2/client_presence.py
MideDay: Was also used for GraphPresence. As GraphPresence is going to be removed, HeartbeatPresence and ClientPresence will be merged into one class, removing the abstract base class

giskardpy/src/giskardpy/middleware/ros2/client_presence.py
MideDay: Changed to 1s and timeout to 3s for now

giskardpy/src/giskardpy/middleware/ros2/client_presence.py
MideDay: Changed to field with default=time.monotonic.

tomsch420 (approved):

ichumuh (approved):

PR #612, opened 2026-09-09
Fix duplicated goal children on motion statechart JSON round trip

MarvinDufner:
`MotionStatechart._from_json` could return a chart where a goal's children
  appear more than once, multiplying by nesting depth. On deserialization the
  motions then run several times.

  The cause: `Sequence` and `Parallel` declare their `nodes` field `init=True`,
  so they serialize their children inline, while the base `Goal` keeps `nodes`
  `init=False` and builds its children in `expand()`. A registered goal therefore
  carries each child both inline and in the flat node list, and `_from_json`
  re-nested from the flat list on top of the inline copies.

  `_from_json` now drops a goal's inline children the first time one of its flat
  children is re-nested, so each child is attached once. A goal whose children
  only live inline, an unexpanded nested motion chart that no flat node names as
  its parent, is left untouched and keeps them.

  ### Tests

  Added `test_goal_children_survive_json_round_trip_exactly_once`, covering both a
  registered goal (children in the flat list and inline) and an inline-only goal
  built through the constructor. It fails without the fix.

### DIFF
--- a/giskardpy/src/giskardpy/motion_statechart/motion_statechart.py
+++ b/giskardpy/src/giskardpy/motion_statechart/motion_statechart.py
@@ -1199,6 +1199,12 @@ class MotionStatechart(SubclassJSONSerializer):
         :meth:`to_json`: first all nodes, then their transition conditions, then
         goal/child parent links.
 
+        A registered goal serializes each child both in the flat node list and inline
+        (Sequence, Parallel), so its inline copies are dropped the first time a flat
+        child is re-nested; keeping them would list each child twice. A goal whose
+        children only live inline (an unexpanded nested motion chart named by no flat
+        node) is untouched.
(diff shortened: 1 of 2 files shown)


ichumuh (changes requested):
modifying to/from json is not the right fix for that, it only solves the symptom doesn't fixes the cause. you need to make `add_node` in Goal private and give Sequence/Parallel and `add_node` that only does `self.nodes.append(node)`

ichumuh:
and tell claude to figure out a better name for `_add_node`

AbdelrhmanBassiouny (approved):

ichumuh:
fixed by https://github.com/cram2/cognitive_robot_abstract_machine/pull/615

PR #618, opened 2026-09-11
Smaller statechart jsons

ichumuh:
- motion statechart jsons are now much smaller, because they only hold a body id instead of a json copy
- Connection is now a WorldEntityWithID, so that we can generalize this to connections
- replaces https://github.com/cram2/cognitive_robot_abstract_machine/pull/548


LucaKro (changes requested):
labels?

tomsch420 (approved):

LucaKro (approved):

PR #625, opened 2026-09-11
feat(semantic_digital_twin): add GraspClutter6D dataset adapter

tomsch420:
https://github.com/user-attachments/assets/09c84251-5e3e-4e7d-bfe5-fad06976e205

## Summary
- Adds `semantic_digital_twin.adapters.grasp_clutter_6d_dataset`, a loader and BOP-format scene parser for the [GraspClutter6D](https://huggingface.co/datasets/GraspClutter6D/GraspClutter6D) dataset (1000 cluttered bin/shelf/table scenes, 200 objects + YCB-Video objects, per-frame 6D ground-truth poses), mirroring the existing `sage_10k_dataset`/`artvip_dataset` adapters.
- `GraspClutter6DDatasetLoader` downloads/extracts the dataset's 7z archives from Hugging Face (`split_info`, the model variants, and the combined ~203GB `scenes` archive - which cannot be fetched per-scene, unlike this package's other adapters, since GraspClutter6D packs all 1000 scenes into one multi-volume archive).
- `GraspClutter6DScene.create_world()` builds a `World` for one frame: a camera body, and one `Connection6DoF`-attached body per object at its ground-truth pose, with an optional `map` root when the frame carries a world-to-camera transform. Objects are attached directly to the world root (not the camera body) so the resulting world can be mirrored into a MuJoCo simulation - a free joint must be a direct child of the world root, which is covered by a MuJoCo-backed regression test.
- Adds `py7zr` as a dependency.

Verified against real downloaded dataset files (not just synthetic fixtures): `split_info.7z`, `models_eval.7z`, and (after downloading and extracting the full ~221GB scenes archive) a real annotated scene's `scene_camera.json`/`scene_gt.json`, rendered end-to-end through `MujocoVideoRecorder`.

## Test plan
- [x] `pytest test/semantic_digital_twin_test/test_grasp_clutter_6d` - 13 passed (network-backed tests skip gracefully if the dataset is unreachable; the MuJoCo-backed test only runs in CI)
- [x] Full `semantic_digital_twin` suite - no regressions (1672 passed, 10 skipped, 1 xfailed)

### DIFF
--- a/semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/exceptions.py
+++ b/semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/exceptions.py
@@ -49,7 +49,10 @@ class GraspClutter6DImageNotFoundError(DataclassException, LookupError):
         return f"Frame '{self.image_id}' not found in GraspClutter6D scene '{self.scene_id}'."
 
     def suggest_correction(self) -> str:
-        return "Check `scene.frames.keys()` for the frame ids available in this scene."
+        return (
+            "Check the image_id of each frame in `scene.frames`, or call "
+            "`scene.frame(image_id)` to look one up."
+        )
 
 
 @dataclass
--- a/semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/exceptions.py
+++ b/semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/exceptions.py
@@ -1,0 +1,132 @@
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from pathlib import Path
+from typing import Tuple
+
+from krrood.exceptions import DataclassException
+
+
+@dataclass
+class GraspClutter6DSceneFilesMissingError(DataclassException, LookupError):
+    """
+    Raised when a scene's directory does not contain both `scene_camera.json` and
+    `scene_gt.json`.
+    """
+
+    scene_id: str
+    """The scene id that was being loaded."""
+
+    directory: Path
+    """The directory that was searched."""
+
+    def error_message(self) -> str:
+        return (
+            f"GraspClutter6D scene '{self.scene_id}' at '{self.directory}' is missing "
+            f"'scene_camera.json' and/or 'scene_gt.json'."
+        )
+
+    def suggest_correction(self) -> str:
+        return (
+            "Make sure the scenes archive was fully extracted and that this directory "
+            "is the scene's own folder, not one of its parents."
+        )
+
+
+@dataclass
+class GraspClutter6DImageNotFoundError(DataclassException, LookupError):
+    """
+    Raised when a requested frame id is not part of a scene.
+    """
+
+    scene_id: str
+    """The scene that was searched."""
+
+    image_id: str
+    """The frame id that could not be found."""
+
+    def error_message(self) -> str:
+        return f"Frame '{self.image_id}' not found in GraspClutter6D scene '{self.scene_id}'."
+
+    def suggest_correction(self) -> str:
+        return "Check `scene.frames.keys()` for the frame ids available in this scene."
+
+
+@dataclass
+class GraspClutter6DObjectModelNotFoundError(DataclassException, LookupError):
+    """
+    Raised when an object's mesh file is not found in the given models directory.
+    """
+
+    obj_id: int
+    """The object id that was searched for."""
+
+    models_directory: Path
+    """The directory that was searched."""
+
+    def error_message(self) -> str:
+        return (
+            f"No mesh file for object id {self.obj_id} found in "
+            f"'{self.models_directory}' (expected 'obj_{self.obj_id:06d}.ply')."
+        )
+
+    def suggest_correction(self) -> str:
+        return (
+            "Download and extract one of the dataset's model archives with "
+            "GraspClutter6DDatasetLoader.download_models()."
+        )
+
+
+@dataclass
+class GraspClutter6DArchiveLayoutError(DataclassException, LookupError):
+    """
+    Raised when extracting one of the dataset's archives does not produce the top-level
+    folder its filename is expected to create.
+    """
+
+    archive_path: Path
+    """The archive that was extracted."""
+
+    expected_directory: Path
+    """The folder that was expected to exist afterward, but does not."""
+
+    def error_message(self) -> str:
+        return (
+            f"Extracting '{self.archive_path}' did not create the expected "
+            f"'{self.expected_directory}'."
+        )
+
+    def suggest_correction(self) -> str:
+        return (
+            "Inspect the archive's actual top-level contents and adjust the extraction "
+            "logic - this dataset's archive-name/folder-name convention may not hold for "
+            "this archive."
+        )
+
+
+@dataclass
+class GraspClutter6DSceneNotFoundError(DataclassException, LookupError):
+    """
+    Raised when a scene id does not resolve to exactly one folder under the extracted
+    scenes directory.
+    """
+
+    scene_id: str
+    """The scene id that was searched for."""
+
+    scenes_directory: Path
+    """The extracted scenes root that was searched."""
+
+    candidates: Tuple[Path, ...] = field(default_factory=tuple)
+    """Every folder that matched, if more than one did."""
+
+    def error_message(self) -> str:
+        if self.candidates:
+            return (
+                f"Scene id '{self.scene_id}' matched {len(self.candidates)} folders "
+                f"under '{self.scenes_directory}', not exactly one: {self.candidates}."
+            )
+        return f"Scene id '{self.scene_id}' not found under '{self.scenes_directory}'."
+
+    def suggest_correction(self) -> str:
+        return "Call GraspClutter6DDatasetLoader.available_scene_ids() for valid scene ids."
--- a/semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/exceptions.py
+++ b/semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/exceptions.py
@@ -58,7 +58,7 @@ class GraspClutter6DObjectModelNotFoundError(DataclassException, LookupError):
     Raised when an object's mesh file is not found in the given models directory.
     """
 
-    obj_id: int
+    object_id: int
     """The object id that was searched for."""
 
     models_directory: Path
@@ -66,8 +66,8 @@ class GraspClutter6DObjectModelNotFoundError(DataclassException, LookupError):
 
     def error_message(self) -> str:
         return (
-            f"No mesh file for object id {self.obj_id} found in "
-            f"'{self.models_directory}' (expected 'obj_{self.obj_id:06d}.ply')."
+            f"No mesh file for object id {self.object_id} found in "
+            f"'{self.models_directory}' (expected 'obj_{self.object_id:06d}.ply')."
         )
 
     def suggest_correction(self) -> str:
--- a/semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/loader.py
+++ b/semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/loader.py
@@ -7,7 +7,8 @@ import shutil
 from dataclasses import dataclass, field
 from enum import StrEnum
 from pathlib import Path
-from typing import ClassVar, Optional, Tuple
+from types import ModuleType
+from typing import Optional, Tuple
 
 logger = logging.getLogger(__name__)
 
@@ -26,7 +27,7 @@ except ImportError:
         "huggingface_hub not installed. `GraspClutter6DDatasetLoader` downloads will not "
         "work. Install it with `pip install huggingface_hub`."
     )
-    huggingface_hub = None
+    huggingface_hub: Optional[ModuleType] = None
 
 try:
     import py7zr
@@ -35,7 +36,7 @@ except ImportError:
         "py7zr not installed. `GraspClutter6DDatasetLoader` archive extraction will not "
         "work. Install it with `pip install py7zr`."
     )
-    py7zr = None
+    py7zr: Optional[ModuleType] = None
 
 
 class GraspClutter6DModelVariant(StrEnum):
@@ -92,12 +93,16 @@ class GraspClutter6DDatasetLoader:
     repository_id: str = "GraspClutter6D/GraspClutter6D"
     """The Hugging Face dataset repository ID."""
 
-    _SCENES_ARCHIVE_PARTS: ClassVar[Tuple[str, ...]] = (
-        "scenes.7z.001",
-        "scenes.7z.002",
-        "scenes.7z.003",
-        "scenes.7z.004",
-        "scenes.7z.005",
+    _scenes_archive_parts: Tuple[str, ...] = field(
+        default=(
+            "scenes.7z.001",
+            "scenes.7z.002",
+            "scenes.7z.003",
+            "scenes.7z.004",
+            "scenes.7z.005",
+        ),
+        init=False,
+        repr=False,
     )
     """
     The 5 volumes of the dataset's one combined scenes archive (~203 GB total), split
@@ -204,7 +209,7 @@ class GraspClutter6DDatasetLoader:
         if expected_directory.exists():
             return expected_directory
         part_paths = [
-            self._download_archive(name) for name in self._SCENES_ARCHIVE_PARTS
+            self._download_archive(name) for name in self._scenes_archive_parts
         ]
         return self._extract_archive(part_paths[0], self.directory, "scenes")
 
--- a/semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/loader.py
+++ b/semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/loader.py
@@ -54,6 +54,29 @@ class GraspClutter6DModelVariant(StrEnum):
     OBJ_METERS = "models_obj_m"
 
 
+class GraspClutter6DObjectSet(StrEnum):
+    """
+    The two object catalogs GraspClutter6D's scenes are built from. See
+    :meth:`GraspClutter6DDatasetLoader.available_scene_ids`.
+    """
+
+    GRASP = "grasp"
+    """The dataset's own 200 novel objects."""
+
+    YCBV = "ycbv"
+    """The standard YCB-Video objects."""
+
+
+class GraspClutter6DSplit(StrEnum):
+    """
+    The dataset's train/test split. See
+    :meth:`GraspClutter6DDatasetLoader.available_scene_ids`.
+    """
+
+    TRAIN = "train"
+    TEST = "test"
+
+
 @dataclass
 class GraspClutter6DDatasetLoader:
     """
@@ -214,18 +237,21 @@ class GraspClutter6DDatasetLoader:
         return self._extract_archive(part_paths[0], self.directory, "scenes")
 
     def available_scene_ids(
-        self, object_set: str = "grasp", split: str = "train"
+        self,
+        object_set: GraspClutter6DObjectSet = GraspClutter6DObjectSet.GRASP,
+        split: GraspClutter6DSplit = GraspClutter6DSplit.TRAIN,
     ) -> Tuple[str, ...]:
         """
         List the scene ids of one split, without downloading any scene data.
 
-        :param object_set: ``"grasp"`` (scenes built from the dataset's own 200 novel
-            objects) or ``"ycbv"`` (scenes built from the standard YCB-Video objects).
-        :param split: ``"train"`` or ``"test"``.
+        :param object_set: Which object catalog the scenes are built from.
+        :param split: Which split to list.
         :return: The scene ids of that split, e.g. ``("000005", "000009", ...)``.
         """
         split_info_directory = self.download_split_info()
-        file_path = split_info_directory / f"{object_set}_{split}_scene_ids.json"
+        file_path = (
+            split_info_directory / f"{object_set.value}_{split.value}_scene_ids.json"
+        )
         return tuple(json.loads(file_path.read_text()))
 
     def object_ids_for_scene(self, scene_id: str) -> Tuple[int, ...]:
--- a/semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/loader.py
+++ b/semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/loader.py
@@ -176,6 +176,11 @@ class GraspClutter6DDatasetLoader:
         expected_directory = extraction_root / expected_name
         if expected_directory.exists():
             return expected_directory
+        if py7zr is None:
+            raise ImportError(
+                "py7zr is required to extract GraspClutter6D archives. Install it with "
+                "`pip install py7zr`, or the semantic_digital_twin 'datasets' extra."
+            )
         extraction_root.mkdir(parents=True, exist_ok=True)
         try:
             with py7zr.SevenZipFile(str(archive_path), mode="r") as archive:
--- a/semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/loader.py
+++ b/semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/loader.py
@@ -1,0 +1,276 @@
+from __future__ import annotations
+
+import json
+import logging
+import os
+import shutil
+from dataclasses import dataclass, field
+from pathlib import Path
+from typing import Optional, Tuple
+
+logger = logging.getLogger(__name__)
+
+from semantic_digital_twin.adapters.grasp_clutter_6d_dataset.exceptions import (
+    GraspClutter6DArchiveLayoutError,
+    GraspClutter6DSceneNotFoundError,
+)
+from semantic_digital_twin.adapters.grasp_clutter_6d_dataset.schema import (
+    GraspClutter6DScene,
+)
+
+try:
+    import huggingface_hub
+except ImportError:
+    logger.warning(
+        "huggingface_hub not installed. `GraspClutter6DDatasetLoader` downloads will not "
+        "work. Install it with `pip install huggingface_hub`."
+    )
+    huggingface_hub = None
+
+try:
+    import py7zr
+except ImportError:
+    logger.warning(
+        "py7zr not installed. `GraspClutter6DDatasetLoader` archive extraction will not "
+        "work. Install it with `pip install py7zr`."
+    )
+    py7zr = None
+
+
+MODEL_ARCHIVE_VARIANTS = (
+    "models",
+    "models_eval",
+    "models_m",
+    "models_obj",
+    "models_obj_eval",
+    "models_obj_m",
+)
+"""
+The dataset's model archives: millimeter- and meter-unit point-cloud (`.ply`) and mesh
+(`.obj`) variants, plus a simplified/watertight `_eval` variant used for pose-error
+metrics. See :meth:`GraspClutter6DDatasetLoader.download_models`.
+"""
+
+_SCENES_ARCHIVE_PARTS = (
+    "scenes.7z.001",
+    "scenes.7z.002",
+    "scenes.7z.003",
+    "scenes.7z.004",
+    "scenes.7z.005",
+)
+"""
+The 5 volumes of the dataset's one combined scenes archive (~203 GB total), split because
+Hugging Face Hub caps individual file size.
+"""
+
+
+@dataclass
+class GraspClutter6DDatasetLoader:
+    """
+    Loader for scenes of the GraspClutter6D dataset
+    (https://huggingface.co/datasets/GraspClutter6D/GraspClutter6D), a dataset of 1000
+    densely cluttered bin/shelf/table scenes (~14 objects/scene, 200 object models plus
+    the standard YCB-Video objects), annotated with per-frame 6D object poses in the BOP
+    dataset format (https://github.com/thodan/bop_toolkit).
+
+    .. important::
+        Unlike this package's other dataset loaders (e.g.
+        :class:`~semantic_digital_twin.adapters.sage_10k_dataset.loader.Sage10kDatasetLoader`),
+        GraspClutter6D does not store scenes as separate repository files - all 1000
+        scenes are packed into one combined, 5-volume, ~203 GB `scenes.7z` archive, so
+        there is no way to fetch a single scene without downloading and extracting that
+        entire archive first. :meth:`download_split_info` and :meth:`download_models` are
+        comparatively small (KB-GB) and safe to call freely; :meth:`download_scenes` is
+        not.
+
+    .. note::
+        Requires the ``huggingface_hub`` and ``py7zr`` packages.
+    """
+
+    directory: Path = field(
+        default_factory=lambda: Path.home() / "graspclutter6d-dataset"
+    )
+    """The directory archives are downloaded to and extracted in."""
+
+    token: Optional[str] = field(default_factory=lambda: os.environ.get("HF_TOKEN"))
+    """
+    The Hugging Face access token used to download the dataset.
+
+    The dataset is public, so this is only needed to raise Hugging Face's anonymous-
+    access rate limit.
+    """
+
+    repository_id: str = "GraspClutter6D/GraspClutter6D"
+    """The Hugging Face dataset repository ID."""
+
+    def _download_archive(self, filename: str) -> Path:
+        """
+        :param filename: The archive's path within the repository, e.g.
+            ``"models_eval.7z"``.
+        :return: The path the archive was downloaded to. Returns early without a network
+            call if it is already there.
+        """
+        return Path(
+            huggingface_hub.hf_hub_download(
+                repo_id=self.repository_id,
+                repo_type="dataset",
+                filename=filename,
+                token=self.token,
+                local_dir=self.directory,
+            )
+        )
+
+    @staticmethod
+    def _extract_archive(
+        archive_path: Path, extraction_root: Path, expected_name: str
+    ) -> Path:
+        """
+        Extract a (possibly multi-volume) 7z archive into `extraction_root`, returning
+        the subdirectory the archive itself creates.
+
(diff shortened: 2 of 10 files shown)


semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/loader.py
```diff
@@ -0,0 +1,276 @@
...
+    import py7zr
+except ImportError:
+    logger.warning(
+        "py7zr not installed. `GraspClutter6DDatasetLoader` archive extraction will not "
+        "work. Install it with `pip install py7zr`."
+    )
+    py7zr = None
+
+
+MODEL_ARCHIVE_VARIANTS = (
```
tomsch420: no module level varaibles, move this to and enum or the loader

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/schema.py
```diff
@@ -0,0 +1,402 @@
...
+from semantic_digital_twin.world import World
+from semantic_digital_twin.world_description.connections import (
+    Connection6DoF,
+    FixedConnection,
+)
+from semantic_digital_twin.world_description.geometry import Mesh
+from semantic_digital_twin.world_description.shape_collection import ShapeCollection
+from semantic_digital_twin.world_description.world_entity import Body
+
+MILLIMETERS_TO_METERS = 1e-3
```
tomsch420: no module level variables

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/schema.py
```diff
@@ -0,0 +1,402 @@
...
+
+@dataclass
+class GraspClutter6DCameraInfo:
+    """
+    Camera parameters of one frame, parsed from one entry of a scene's
+    `scene_camera.json`, following the BOP dataset format
+    (https://github.com/thodan/bop_toolkit).
+    """
+
+    intrinsics: np.ndarray
```
tomsch420: we have a camera class already. Perhaps use this

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/schema.py
```diff
@@ -0,0 +1,402 @@
...
+
+
+@dataclass
+class GraspClutter6DObjectPose:
+    """
+    One object's ground-truth pose in one frame, parsed from one entry of a scene's
+    `scene_gt.json`, following the BOP dataset format.
+    """
+
+    obj_id: int
```
tomsch420: no abbrivations

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/schema.py
```diff
@@ -0,0 +1,402 @@
...
+    """
+    1-indexed object id, matching the `obj_%06d` numbering of the dataset's model files
+    and `models_info.json` keys.
+    """
+
+    camera_R_object: np.ndarray
```
tomsch420: if its some form of rotation matrix, use the correct spatial type

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/loader.py
tomsch420: Moved to a GraspClutter6DModelVariant StrEnum and a class-level constant on the loader, in 911e0d0a3.

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/schema.py
tomsch420: Moved to a ClassVar on GraspClutter6DCameraInfo, in 911e0d0a3.

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/schema.py
tomsch420: Added a GraspClutter6DCamera(Camera) subclass and used it for the frame's camera body instead of a bare Body, with a field of view estimated from cam_K. In 911e0d0a3.

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/schema.py
tomsch420: Renamed obj_id to object_id throughout (this field and the matching exceptions.py field). In 911e0d0a3.

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/schema.py
tomsch420: Changed camera_R_object to RotationMatrix and camera_t_object to Point3 (and the matching world_R_camera/world_t_camera fields on GraspClutter6DCameraInfo), converting mm to m at parse time instead of at each point of use. In 911e0d0a3.

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/schema.py
```diff
@@ -0,0 +1,481 @@
...
+    `scene_gt.json`, following the BOP dataset format.
+    """
+
+    object_id: int
+    """
+    1-indexed object id, matching the `obj_%06d` numbering of the dataset's model files
+    and `models_info.json` keys.
+    """
+
+    camera_R_object: RotationMatrix
```
tomsch420: put these two together in a real pose (spatial type) and dont keep them seperated

tomsch420: or homogenous transformation matrix if its a better fit

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/schema.py
```diff
@@ -0,0 +1,481 @@
...
+
+    image_id: str
+    """
+    The frame's id, exactly as given by `scene_camera.json`/`scene_gt.json` (e.g.
+    ``"1"``). GraspClutter6D numbers its four physical cameras (RealSense D415/D435,
+    Azure Kinect, Zivid) and their viewpoints into one flat sequence of frames per scene,
+    so this id does not by itself say which physical camera captured it.
+    """
+
+    camera: GraspClutter6DCameraInfo
```
tomsch420, line 159: docstrings

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/schema.py
```diff
@@ -0,0 +1,481 @@
...
+    dataset is converted with this factor.
+    """
+
+    intrinsics: np.ndarray
+    """The 3x3 camera intrinsic matrix (`cam_K`), in pixels."""
+
+    depth_scale: float
+    """Multiply this frame's raw depth image values by this factor to get millimeters."""
+
+    world_R_camera: Optional[RotationMatrix] = None
```
tomsch420: also here bundle it to a homegenous transformation matrix

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/schema.py
```diff
@@ -0,0 +1,481 @@
...
+    """
+
+    MILLIMETERS_TO_METERS: ClassVar[float] = 1e-3
+    """
+    GraspClutter6D's BOP-format JSON files give every translation in millimeters; the
+    rest of this package works in meters (SI), so every translation parsed from the
+    dataset is converted with this factor.
+    """
+
+    intrinsics: np.ndarray
```
tomsch420: do we have some special type for intrinsics in semdt already?

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/schema.py
```diff
@@ -0,0 +1,481 @@
...
+        focal_x, focal_y = self.intrinsics[0, 0], self.intrinsics[1, 1]
+        principal_x, principal_y = self.intrinsics[0, 2], self.intrinsics[1, 2]
+        return FieldOfView(
+            horizontal_angle=2 * np.arctan(principal_x / focal_x),
+            vertical_angle=2 * np.arctan(principal_y / focal_y),
+        )
+
+
+@dataclass(eq=False)
+class GraspClutter6DCamera(Camera):
```
tomsch420: this is then not a Liskov Substitutable class, remove it again

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/schema.py
```diff
@@ -0,0 +1,481 @@
...
+        annotations are wanted.
+    """
+
+    scene_id: str
+    """The scene's id, e.g. ``"000005"``."""
+
+    directory: Path
+    """The scene's own extracted folder, directly containing `scene_camera.json`/`scene_gt.json`."""
+
+    frames: Dict[str, GraspClutter6DFrame] = field(default_factory=dict)
```
tomsch420: why do we need the key?

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/schema.py
tomsch420: Bundled into a single frameless HomogeneousTransformationMatrix field (camera_T_object), in 7cd8d13f9.

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/schema.py
tomsch420: Same treatment: bundled into camera_T_world (a frameless HomogeneousTransformationMatrix), in 7cd8d13f9.

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/schema.py
tomsch420: Added docstrings to the camera and object_poses fields, in 7cd8d13f9.

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/schema.py
tomsch420: Checked - the only camera-related type in semdt is CameraResolution (plain pixel width/height), nothing for a projection/intrinsics matrix. Left it as np.ndarray and documented why in the field's docstring, in 7cd8d13f9.

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/schema.py
tomsch420: Agreed, removed - overriding setup_default_configuration_in_world_below_robot_root to raise was exactly the LSP violation. Back to a bare Body for the camera. In 7cd8d13f9.

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/schema.py
tomsch420: No good reason - GraspClutter6DFrame already carries its own image_id, so the dict key was a redundant duplicate. Replaced with a plain List[GraspClutter6DFrame] plus a frame(image_id) lookup method, in 7cd8d13f9.

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/loader.py
```diff
@@ -0,0 +1,273 @@
...
+    huggingface_hub = None
+
+try:
+    import py7zr
+except ImportError:
+    logger.warning(
+        "py7zr not installed. `GraspClutter6DDatasetLoader` archive extraction will not "
+        "work. Install it with `pip install py7zr`."
+    )
+    py7zr = None
```
tomsch420: make this nonetype so its at least a type

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/loader.py
```diff
@@ -0,0 +1,273 @@
...
+    The Hugging Face access token used to download the dataset.
+
+    The dataset is public, so this is only needed to raise Hugging Face's anonymous-
+    access rate limit.
+    """
+
+    repository_id: str = "GraspClutter6D/GraspClutter6D"
+    """The Hugging Face dataset repository ID."""
+
+    _SCENES_ARCHIVE_PARTS: ClassVar[Tuple[str, ...]] = (
```
tomsch420: make this a field

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/schema.py
tomsch420: how do we store intrinsics in other cases?

tomsch420: Checked further - nobody stores raw intrinsics anywhere in this codebase. Every other Camera (PR2KinectV1, TracyCamera, HSRB*, etc.) only ever carries a FieldOfView (literal calibrated angles), never a K matrix. Fixing this now: dropping the intrinsics field entirely and storing field_of_view directly, matching that pattern - cam_K is only used transiently at parse time to derive it.

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/loader.py
tomsch420: Added an explicit Optional[ModuleType] annotation to both fallbacks, in afdfe020c.

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/loader.py
tomsch420: Turned into a proper (init=False) dataclass field, _scenes_archive_parts, in afdfe020c.

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/schema.py
tomsch420: Done: dropped the intrinsics field entirely, field_of_view is now computed at parse time and stored directly (cam_K is only used transiently), in afdfe020c.

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/loader.py
Narenvasant, line 1: could you add texts about this graspclutter6d dataset in the dataset.md file as well?

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/schema.py
```diff
@@ -0,0 +1,417 @@
...
+    every other camera in this package (e.g.
+    :class:`~semantic_digital_twin.robots.pr2.PR2KinectV1`) only ever stores a
+    `FieldOfView`, never a raw intrinsics matrix, so `cam_K` is used only transiently, to
+    derive this, rather than kept as a field of its own. `scene_camera.json` does not
+    give the image resolution directly, so this assumes the principal point sits at the
+    image center (``width = 2 * cx``, ``height = 2 * cy``) - a common approximation, not
+    a value read from the dataset itself.
+    """
+
+    depth_scale: float
```
Narenvasant, line 72: are we loading depth image too?

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/exceptions.py
```diff
@@ -0,0 +1,132 @@
...
+    """The scene that was searched."""
+
+    image_id: str
+    """The frame id that could not be found."""
+
+    def error_message(self) -> str:
+        return f"Frame '{self.image_id}' not found in GraspClutter6D scene '{self.scene_id}'."
+
+    def suggest_correction(self) -> str:
+        return "Check `scene.frames.keys()` for the frame ids available in this scene."
```
Narenvasant: shouldnt this be scene.frames(image_id)?

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/schema.py
```diff
@@ -0,0 +1,417 @@
...
+                child=body,
+                parent_T_connection_expression=root_T_object,
+            )
+            world.add_body(body)
+            world.add_connection(connection)
+
+        with world.modify_world():
+            world.add_semantic_annotation(
+                NaturalLanguageWithTypeDescription(
+                    root=body, description=name, type_description="object"
```
Narenvasant: why type_description="object", shouldnt this also be called as name

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/loader.py
```diff
@@ -0,0 +1,278 @@
...
+    (https://huggingface.co/datasets/GraspClutter6D/GraspClutter6D), a dataset of 1000
+    densely cluttered bin/shelf/table scenes (~14 objects/scene, 200 object models plus
+    the standard YCB-Video objects), annotated with per-frame 6D object poses in the BOP
+    dataset format (https://github.com/thodan/bop_toolkit).
+
+    .. important::
+        Unlike this package's other dataset loaders (e.g.
+        :class:`~semantic_digital_twin.adapters.sage_10k_dataset.loader.Sage10kDatasetLoader`),
+        GraspClutter6D does not store scenes as separate repository files - all 1000
+        scenes are packed into one combined, 5-volume, ~203 GB `scenes.7z` archive, so
```
Narenvasant, line 93: 200 g XD memory gaming

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/schema.py
```diff
@@ -0,0 +1,417 @@
...
+    ) -> Body:
+        """
+        Create one object's body, mesh, connection to `root_body`, and semantic
+        annotation in `world`.
+
+        :param camera_body: The frame's camera body - always used to compute the
+            object's pose from its ground-truth `cam_R_m2c`/`cam_t_m2c`, since that is
+            always available and unambiguous.
+        :param root_body: The body the object's :class:`Connection6DoF` is actually
+            attached to. Must be `world.root` (`camera_body` itself when there is no
```
Narenvasant, line 365: gucci

Narenvasant (changes requested):
Nice work 🤝

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/loader.py
tomsch420: yes

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/schema.py
tomsch420: not really, the schema defines every piece of information defined in the dataset. We can then just extract what we need from it, but this is just for reference

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/exceptions.py
tomsch420: actually, this looks indeed like a wrong suggestion. Claude fix it

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/schema.py
tomsch420: true, if the type description is "object" is pretty useless. In the case there is no useful annotation to extract use NaturalLanguageDescription instead Claude

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/loader.py
tomsch420: yeah im currently creating a database for this, so you dont have to download it :D

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/loader.py
tomsch420: Added a GraspClutter6D section to doc/datasets.md, in 4db818be7.

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/exceptions.py
tomsch420: Fixed - it was a stale reference to frames as a dict (from before frames became a List[GraspClutter6DFrame]). Now points at scene.frame(image_id), in 4db818be7.

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/schema.py
tomsch420: Switched to NaturalLanguageDescription (no type_description) - agreed, "object" carried no real information and there's nothing more specific to put there (no semantic names in this dataset, only numeric ids). In 4db818be7.

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/loader.py
```diff
@@ -0,0 +1,278 @@
...
+        expected_directory = self.directory / "scenes"
+        if expected_directory.exists():
+            return expected_directory
+        part_paths = [
+            self._download_archive(name) for name in self._scenes_archive_parts
+        ]
+        return self._extract_archive(part_paths[0], self.directory, "scenes")
+
+    def available_scene_ids(
+        self, object_set: str = "grasp", split: str = "train"
```
tomsch420: can the object_set and split parameters be string enums as well? or is there no finite set of them

tomsch420: Yes, both are genuinely finite - confirmed from the real dataset's split_info files (grasp/ycbv x train/test, no others). Added GraspClutter6DObjectSet and GraspClutter6DSplit enums, in 8876b7ae5.

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/schema.py
```diff
@@ -0,0 +1,415 @@
...
+
+@dataclass
+class GraspClutter6DCameraInfo:
+    """
+    Camera parameters of one frame, parsed from one entry of a scene's
+    `scene_camera.json`, following the BOP dataset format
+    (https://github.com/thodan/bop_toolkit).
+    """
+
+    MILLIMETERS_TO_METERS: ClassVar[float] = 1e-3
```
LucaKro: classvars haram

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/schema.py
```diff
@@ -0,0 +1,415 @@
...
+    derive this, rather than kept as a field of its own. `scene_camera.json` does not
+    give the image resolution directly, so this assumes the principal point sits at the
+    image center (``width = 2 * cx``, ``height = 2 * cy``) - a common approximation, not
+    a value read from the dataset itself.
+    """
+
+    depth_scale: float
+    """Multiply this frame's raw depth image values by this factor to get millimeters."""
+
+    camera_T_world: Optional[HomogeneousTransformationMatrix] = None
```
LucaKro, line 75: why not default factory?

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/schema.py
```diff
@@ -0,0 +1,415 @@
...
+        :raises GraspClutter6DImageNotFoundError: if `image_id` is not in `self.frames`.
+        :raises GraspClutter6DObjectModelNotFoundError: if an object's mesh file is not
+            found in `models_directory`.
+        :return: The built world.
+        """
+        frame = self.frame(image_id)
+
+        world = World()
+        camera_body = Body()
+        camera_body.name = PrefixedName(
```
LucaKro: why not pass it directly?

LucaKro: or actually use World.create_with_root_body directly

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/schema.py
```diff
@@ -0,0 +1,415 @@
...
+        Add a `map` root body and place `camera_body` under it via the inverse of the
+        frame's world-to-camera transform.
+
+        :param world: The world `camera_body` was already added to.
+        :param camera_info: The frame's camera parameters, with `camera_T_world` present.
+        :param camera_body: The already-added camera body.
+        :return: The new `map` body.
+        """
+        map_body = Body()
+        map_body.name = PrefixedName(name="map", prefix=camera_body.name.prefix)
```
LucaKro: why are you doing this everywhere?

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/schema.py
```diff
@@ -0,0 +1,415 @@
...
+
+    camera: GraspClutter6DCameraInfo
+    """This frame's camera parameters."""
+
+    object_poses: List[GraspClutter6DObjectPose] = field(default_factory=list)
+    """The ground-truth pose of every object visible in this frame."""
+
+
+@dataclass
+class GraspClutter6DScene:
```
LucaKro, line 167: i think this class needs to be cleaned up, i think its weird that you first create the camera world and then potentially "add" the world frame afterwards...this is a really confusing workflow. 
also i feel like the body construction and everything is overly complicated as well

LucaKro (changes requested):
noice looks interesting💪
 
please ensure that one can run the tests without needing to add an optional dependency here. I had this happen twice recently and its a bit annoying

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/schema.py
tomsch420: i actually dislike this a lot too. But i didnt find a better way. The problem is that the numbers in the dataset are always millimeters, and i am unsure if its the better solution to put this multiplication with raw floats in the code instead, as it gives me magic number vibes. Claude try to centralize the millimiters to meters function and use it always, then we can remove this constant here

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/schema.py
tomsch420: read the docstring 🤓

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/schema.py
tomsch420: sounds good, claude do it!

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/schema.py
tomsch420: agreed, claude do it

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/schema.py
tomsch420: i will refactor the body creation. However, funny dataset format time, the camera always exists, and the world frame only sometimes 🙃 So its easiest to construct it this way

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/schema.py
tomsch420: Centralized into a single _millimeters_to_meters function (module-level, not a variable/ClassVar) and used it everywhere the 1e-3 factor was needed, including create_world's mesh_unit_scale default. In 49094f202.

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/schema.py
tomsch420: Done - create_world now builds via World.create_with_root_body(root_body_name="camera", prefix=...) instead of World() + a separate add_body call, and camera_body/map_body/the object body all get their name passed directly to Body(name=...) instead of being assigned afterwards. In 49094f202.

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/schema.py
tomsch420: Same fix as the sibling thread on line 281 - name is now passed directly to Body(name=...) for camera_body, map_body, and the object body, instead of being assigned as a separate statement afterwards. In 49094f202.

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/schema.py
LucaKro: maybe we should either have a unit subpackage in krrood or semdt that handles this, or we use something like https://pint.readthedocs.io/en/stable/

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/schema.py
```diff
@@ -0,0 +1,411 @@
...
+        """
+        frame = self.frame(image_id)
+
+        world = World.create_with_root_body(
+            root_body_name="camera", prefix=f"{self.scene_id}_{image_id}"
+        )
+        camera_body = world.root
+
+        root_body = camera_body
+        if with_world_frame and frame.camera.camera_T_world is not None:
```
LucaKro: in what cases would you pass with_world_frame=True and frame.camera.camera_T_world = None or the other way around? this feels like duplicated information, and its sus that it may contradict

test/semantic_digital_twin_test/test_grasp_clutter_6d/test_loader.py
```diff
@@ -1,6 +1,10 @@
 import re
 
 import pytest
+
+pytest.importorskip(
```
LucaKro, line 5: i like

LucaKro (changes requested):

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/schema.py
tomsch420: as discussed in zulib, we can read into it, but in a later PR :D

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/schema.py
tomsch420: in the case  that `with_world_frame=False` and there is a transform means the data is always camera relative.  `with_world_frame=True` and no transform is actually a bug. Claude, raise a dedicated exception for when this is tried to be done

tomsch420: Agreed - added a dedicated GraspClutter6DMissingWorldFrameError, raised in create_world when with_world_frame=True but the frame's camera has no camera_T_world, instead of silently falling back to camera-as-root. with_world_frame=False + a transform present stays as-is (data is always camera-relative). In ea6203eb3.

semantic_digital_twin/src/semantic_digital_twin/adapters/grasp_clutter_6d_dataset/schema.py
```diff
@@ -279,7 +283,11 @@ def create_world(
         camera_body = world.root
 
         root_body = camera_body
-        if with_world_frame and frame.camera.camera_T_world is not None:
+        if with_world_frame:
+            if frame.camera.camera_T_world is None:
```
tomsch420: use an and here

tomsch420: Done - flattened into a single `if with_world_frame and frame.camera.camera_T_world is None: raise ...` guard clause, with the _add_world_frame call kept in its own `if with_world_frame:` right after. In b8aa88d97.

Narenvasant (approved):

LucaKro (approved):

PR #631, opened 2026-09-15
Underspecified generic now behaves like superclass in ORMatic

tomsch420:
Fixes a bug where a referenced underspecified generic in a field type annotation would not behave like a superclass mention: ORMatic's `parse_field` unconditionally skipped a field typed to a bare, unparametrized generic class, even when that class already has a mapped polymorphic root table (because concrete parametrizations of it are mapped elsewhere). Such a field is now mapped as an ordinary relationship to that root, so it round trips to whatever concrete subclass it actually held.

Also fixes two pre-existing bugs that were failing CI here. Neither was caused by the change above, but both were exposed by it — fixing the first cleared a cascade of ~75 unrelated-looking failures that were hiding the second underneath, and CI had never run against either, since this branch's base commit was pushed straight to `main` without going through a pull request.

- `test_complicated_equal` asserted a scalar from `EQLTranslator.evaluate()` for a `the()` query, but a prior commit already correctly changed that method to match its own `List[Any]` contract; the test was never updated to match.
- A shared generic instance resolved to a DAO class based on whichever field reached it first. `_get_or_queue_dao` keeps the DAO it resolves for an object the first time it sees it, so an object reached first through a bare-typed field got pinned to the ambiguous base DAO, silently dropping everything the concrete table holds. It now resolves through the object's own parametrization (`__orig_class__`), which is independent of traversal order. An earlier attempt ordered relationships by type specificity instead; that only ordered within a single owner, so it left the cross-owner case broken and has been reverted in favour of the above.

Finally, `test_annotation_specification_survives_orm_round_trip` in `semantic_digital_twin` was marked `xfail(strict=True)` specifically because ORMatic used to map no specification-valued field at all — its own message said to remove the marker once the generator maps those fields, which this PR's core fix now does, so the marker (and the strict-xpass failure it caused) is removed.


ichumuh (approved):

krrood/src/krrood/ormatic/data_access_objects/dao.py
```diff
@@ -707,23 +720,39 @@ def _fill_relationships_from_plan(
...
-                setattr(self, relationship.key, None)
-            else:
-                setattr(
-                    self,
-                    relationship.key,
-                    self._get_or_queue_dao(value, state, relationship.domain_type),
-                )
+        ordered_relationships = sorted(
+            [(True, relationship) for relationship in single_relationships]
+            + [(False, relationship) for relationship in collection_relationships],
```
tomsch420: this is a very hardly readable way of ding this. Refactor this to be easier to read

tomsch420: Refactored in e08940cc59: split into `_fill_relationships_from_plan`, `_fill_single_relationship`, and `_fill_collection_relationship`, each doing one thing, and swapped the boolean-tagged `(True/False, relationship)` tuples for `isinstance` dispatch on the existing `SingleRelationship`/`CollectionRelationship` types. Same behavior (verified: full `krrood` and `semantic_digital_twin` test suites still pass), just easier to follow.

tomsch420: Correction, and it goes further than readability: you were right to be suspicious of this code, because the approach itself was wrong.

Ordering by specificity only orders relationships *within one owner*. A shared object reachable from two different owners is still pinned by whichever owner is converted first, and two different parametrizations of the same object still tie on the sort key. I wrote the cross-owner case as a test and it fails with the ordering in place — so it made `test_generic_class` pass without fixing the problem.

The premise was also just wrong. I claimed the parametrization is erased at runtime so the referencing field's static type is the only signal. It isn't: constructing through a parametrized alias leaves `__orig_class__` on the instance, so the object carries its own type argument. Reading that directly makes resolution independent of traversal order entirely, and the ordering is reverted in a360d59440. A bare construction carries no alias and still resolves to the ambiguous base, which stays correct since nothing records which parametrization it meant.

One thing worth flagging: `_parametrization_of` uses `getattr(obj, "__orig_class__", None)`, which brushes against the "always access attributes via ." rule. The attribute is genuinely optional — CPython sets it only for parametrized construction — and try/except is ruled out too, so a defaulted lookup seemed like the honest option. Happy to change it if you'd rather it went elsewhere.

LucaKro (approved):

PR #628, opened 2026-09-13
Add layered numpy probabilistic circuits with full rustworkx query support

tomsch420:
## Summary

The JAX layered probabilistic circuit implementation trades the structural inferences (truncation, conditioning, marginalization) for hardware acceleration. This adds a numpy implementation with the same layout — sum-unit weights and product-unit edges grouped per layer — that keeps every structural query the rustworkx implementation supports, in `probabilistic_model/src/probabilistic_model/probabilistic_circuit/tensorized/` (the `LayeredProbabilisticCircuit` class).

- Sparse COO representation that keeps explicit zeros (product layers store child-node indices as values, and index `0` is a real edge).
- Bottom-up queries memoized by layer identity for the duration of one pass, so a layer shared by several parents is evaluated once.
- Structural passes never write into the layers they read, which lets a composite truncation work off a single copy of the circuit.
- Each concrete layer declares the `rx` unit type it represents through a `Generic[RustworkxUnitType]` parameter on its base class (e.g. `class ProductLayer(InnerLayer[ProductUnit])`) rather than a separate method. The rustworkx conversion lives in `rustworkx_conversion.py`; the layers themselves are in the `inner_layer/` package, with `base.py` for what every layer shares and a module per layer type.

Truncating to an event with many simple sets — the bottleneck of the target application — is handled in one batched pass instead of once per simple set. Every layer's parameter block is replicated once per simple set and the dead replicas are pruned afterward, so the result stays at a constant number of layers instead of spreading over hundreds of them. A layer that cannot be batched (its simple sets disagree on the resulting layer type) reports this, and the circuit falls back to truncating once per simple set.

## Measurements

Reproduce with `experiments/src/experiments/probabilistic_model_experiments/layered_circuit_speed.py`. Every figure is the mean ± standard deviation over repeated runs of one measurement; the circuit is a 319-node joint probability tree over 4 correlated normal variables.

> Measured on a 15 GB developer laptop, so the absolute numbers are not comparable with figures taken on other machines — the ratios are the point. Two independent runs agreed on every row's direction.

**Truncating to a staircase of disjoint boxes.** This is what the implementation exists for. The layered result stays at 10 layers no matter how many simple sets it was truncated to, where truncating once per simple set and mixing would produce one set of layers per set.

| simple sets | rustworkx | layered | speedup | nodes in result | layers in result |
| --- | --- | --- | --- | --- | --- |
[...]

PR #632, opened 2026-09-15
Add SortedSet support to DataclassJSONSerializer

tomsch420:
## Summary
- `DataclassJSONSerializer.to_json` walked `list`/`set` fields item-by-item but skipped `sortedcontainers.SortedSet`, since it is not a subclass of `set` — a `SortedSet` field fell through to the generic object branch instead.
- Extracted as a standalone, minimal fix from #628 (the layered numpy circuit PR), where a `SortedSet`-typed field surfaced this gap.

## Changes
- `krrood/src/krrood/adapters/json_serializer.py`: extend the `isinstance` check to include `SortedSet`; document the list/set/SortedSet round-trip behavior in the class docstring (deserializes as a plain `list`; a class needing the type back coerces it in `__post_init__`, same convention as other container fields).
- `test/krrood_test/test_utils/test_json_serializer.py`: add `ClassWithSortedSet` fixture and `test_dataclass_sorted_set`, mirroring the existing `ClassWithDict`/`test_dataclass_dict` pair.

## Test plan
- [x] `pytest test/krrood_test/test_utils/test_json_serializer.py` — 27 passed
- [x] `pytest test/krrood_test/test_utils/` — 57 passed


test/krrood_test/test_utils/test_json_serializer.py
```diff
@@ -166,6 +167,15 @@ class ClassWithDict(DataclassJSONSerializer):
     a: Dict[str, int]
 
 
+@dataclass
+class ClassWithSortedSet(DataclassJSONSerializer):
+    a: SortedSet
+
+    def __post_init__(self):
```
tomsch420: remove the post init

krrood/src/krrood/adapters/json_serializer.py
```diff
@@ -614,7 +619,7 @@ def to_json(cls, obj, **kwargs) -> Dict[str, Any]:
         for field_ in introspector.discover(obj.__class__):
             value = getattr(obj, field_.public_name)
 
-            if isinstance(value, (list, set)):
+            if isinstance(value, (list, set, SortedSet)):
```
tomsch420: save the collection type that was used alongside the collection field itself and adjust the from_json method such that it uses the type mentioned there to reconstruct this

test/krrood_test/test_utils/test_json_serializer.py
tomsch420: Removed — from_json now reconstructs the SortedSet directly (see the reply on the other thread), so the constructor always receives one already and the coercion was redundant.

krrood/src/krrood/adapters/json_serializer.py
tomsch420: Done. to_json now tags a set/SortedSet field with {"collection_type": <type name>, "items": [...]} instead of a bare array, and from_json looks up that name in a small type map (DATACLASS_SERIALIZER_COLLECTION_TYPES) to reconstruct the same type instead of always producing a list. Left plain list fields as a bare array on the wire (no tag needed there, since list is already what from_json produces by default) to keep the change minimal. This is a general fix, not SortedSet-specific — a plain set field had the exact same bug, so I added test_dataclass_set alongside test_dataclass_sorted_set to cover both. Full krrood suite: 2387 passed, 5 skipped.

tomsch420: Follow-up: now always tags the collection type, including plain list, instead of leaving list as a bare array — so the wire format for any list/set/SortedSet field no longer depends on list happening to be the deserialization default. Added test_dataclass_list alongside the set/SortedSet tests. Full krrood suite: 2388 passed, 5 skipped; probabilistic_model (excluding test_gui, an unrelated Qt/X11 container issue): 400 passed, 1 skipped.

tomsch420: Follow-up: replaced DATACLASS_SERIALIZER_COLLECTION_TYPES (a plain dict) with an enum, DataclassSerializerCollectionType (LIST/SET/SORTED_SET, each valued as the actual type). to_json writes DataclassSerializerCollectionType.of(value).name as the tag, from_json looks the tag up as a member and uses .value to reconstruct — ties the write and read sides to the same enum instead of a dict whose keys had no compile-time link to what to_json actually wrote. Updated the three tests wire-format assertions accordingly (tags are now "LIST"/"SET"/"SORTED_SET"). Full krrood suite: 2388 passed, 5 skipped.

tomsch420: CI caught a real regression from tagging plain list uniformly: WorldModelModificationBlock.apply_from_json (semantic_digital_twin) hand-parsed its own generically-serialized modifications field as a bare JSON array, so it broke once list started carrying a collection_type tag too — surfaced as AttributeError across giskardpy/experiments/coraplex integration tests. Fixed by having apply_from_json delegate to DataclassJSONSerializer.from_json(data, clazz=cls, **kwargs) instead of re-implementing the reconstruction by hand, so it stays correct regardless of the tag format. Verified with a standalone repro (serialize a block, apply_from_json it against a fake world, confirm the modification is reconstructed and applied) since this container cannot run the semantic_digital_twin/giskardpy suites directly (missing ROS/USD/etc. deps) — CI is the real check here, pushed as 66d7db82f, waiting on it now.

krrood/src/krrood/adapters/json_serializer.py
```diff
@@ -598,13 +599,37 @@ def from_json(
         return np.array(data["data"], dtype=data["type"])
 
 
+class DataclassSerializerCollectionType(enum.Enum):
```
LucaKro: i feel like 1. this is missing tuples, and 2. this is a bit duplicated with list_like_classes. nothing dramatic but maybe this can be unified somehow?

krrood/src/krrood/adapters/json_serializer.py
```diff
@@ -614,8 +639,11 @@ def to_json(cls, obj, **kwargs) -> Dict[str, Any]:
         for field_ in introspector.discover(obj.__class__):
             value = getattr(obj, field_.public_name)
 
-            if isinstance(value, (list, set)):
-                current_result = [to_json(item, **kwargs) for item in value]
+            if isinstance(value, (list, set, SortedSet)):
```
LucaKro: yea this sounds like list_like_classes would fit here as well again

semantic_digital_twin/src/semantic_digital_twin/world_description/world_modification.py
```diff
@@ -523,10 +524,10 @@ def apply_from_json(cls, world: World, data: Dict[str, Any], **kwargs) -> Self:
         """
         Apply the modifications in the given JSON data to the given world.
         """
-        data = data["modifications"]
+        block = DataclassJSONSerializer.from_json(data, clazz=cls, **kwargs)
```
LucaKro: why not just "from_json"?

LucaKro (changes requested):

ichumuh (approved):
just fix what luca said

semantic_digital_twin/src/semantic_digital_twin/world_description/world_modification.py
tomsch420: Agree, claude check if that works.

krrood/src/krrood/adapters/json_serializer.py
tomsch420: Hmmm, i think i have a cooler ormatic like idea. Claude, but the module and the class from the module in the json, then in the from_json load that module and call the class constructor.

semantic_digital_twin/src/semantic_digital_twin/world_description/world_modification.py
tomsch420: Checked — this works. Added a plain `from_json` classmethod to `WorldModelModificationBlock` that delegates to `DataclassJSONSerializer.from_json(data, clazz=cls, **kwargs)`, and removed `apply_from_json` entirely. The one call site (`messages.py`'s `apply_to_json_snapshot_to_world`) now reads `WorldModelModificationBlock.from_json(modification, **kwargs).apply(world)`.

Bonus: this also fixes a latent bug — `WorldModelSnapshot._from_json` already called `WorldModelModificationBlock.from_json(...)`, which didn't exist before this change (would have raised `AttributeError`). That path isn't exercised by the test suite in this environment (needs `geometry_msgs`), so I verified the mechanism manually: built a real `World`, applied modifications, serialized the block, used `WorldEntityWithIDKwargsTracker` to rebuild kwargs, then `from_json(...).apply()`'d it onto a fresh world and confirmed bodies/connections reconstruct correctly. Full krrood suite: 2389 passed, 5 skipped.

krrood/src/krrood/adapters/json_serializer.py
tomsch420: Done — replaced `DataclassSerializerCollectionType` (the LIST/SET/SORTED_SET enum) with your idea: `to_json` now tags a collection field with its fully qualified class name (`get_full_class_name(type(value))`, e.g. `"builtins.tuple"`), and `from_json` imports that module and resolves the class via a new `resolve_class_from_full_name` helper. That helper is factored out of the module+class resolution logic that `SubclassJSONSerializer.from_json` already had inline (it now calls the shared helper too), so it's no longer duplicated there either.

Also switched the isinstance check for which fields get tagged from a hand-written `(list, set, SortedSet)` to `list_like_classes + (SortedSet,)` — literally reusing `list_like_classes`, per your suggestion on the sibling thread — so `tuple` is now covered automatically instead of being missed.

Added `test_dataclass_tuple` and updated the wire-format assertions in the three existing collection tests to expect the fully qualified name instead of the enum member name. Full krrood suite: 2389 passed, 5 skipped.

krrood/src/krrood/adapters/json_serializer.py
```diff
@@ -68,6 +68,31 @@ class JSONData:
         """
 
 
+def resolve_class_from_full_name(fully_qualified_class_name: str) -> Type:
```
tomsch420, line 66: this feels abit duplicated with the TypeType of ORMatic. Check if this can be deduplicated

tomsch420: Done — added a shared `resolve_class_from_full_name` helper to `krrood.utils` (alongside `get_full_class_name`/`module_and_class_name`, whose inverse it is) that does the plain `rsplit(".", 1)` + `importlib.import_module` + `getattr` resolution. `TypeType.process_result_value` now just calls it directly. `json_serializer.py`'s own `resolve_class_from_full_name` still exists but is now a thin wrapper around the shared one that only adds the JSON-specific exception translation (`InvalidTypeFormatError`/`UnknownModuleError`/`ClassNotFoundError`) that `SubclassJSONSerializer.from_json` and the tests depend on — so the actual resolution logic lives in exactly one place now.

krrood test suite (json_serializer + ormatic): 167+30 passed, 0 failed.

semantic_digital_twin/src/semantic_digital_twin/world_description/world_modification.py
```diff
@@ -519,14 +520,8 @@ def update_references_for_world_and_apply(self, world: World):
...
-    def apply_from_json(cls, world: World, data: Dict[str, Any], **kwargs) -> Self:
-        """
-        Apply the modifications in the given JSON data to the given world.
-        """
-        data = data["modifications"]
-
-        for modification in data:
-            from_json(modification, **kwargs).apply(world)
+    def from_json(cls, data: Dict[str, Any], **kwargs) -> Self:
+        return DataclassJSONSerializer.from_json(data, clazz=cls, **kwargs)
```
tomsch420: this change looks dangerours as its semantically different

krrood/src/krrood/adapters/json_serializer.py
```diff
@@ -603,8 +616,18 @@ class DataclassJSONSerializer(ExternalClassJSONSerializer[None]):
...
-    this is not enough, you still need to implement a custom serializer.
+    It creates a dict where all fields are serialized using the to_json function. A
+    ``list``-like (see :data:`list_like_classes`) or :class:`~sortedcontainers.SortedSet`
+    field always serializes as a JSON object that also records its collection type as a
+    fully qualified class name, so ``from_json`` restores that same type on the way back
+    rather than guessing or defaulting to ``list``. If this is not enough, you still need
+    to implement a custom serializer.
+    """
+
+    collection_field_types = list_like_classes + (SortedSet,)
```
tomsch420: just update the list like classes inplace and dont do that here

krrood/src/krrood/adapters/json_serializer.py
```diff
@@ -614,8 +637,11 @@ def to_json(cls, obj, **kwargs) -> Dict[str, Any]:
         for field_ in introspector.discover(obj.__class__):
             value = getattr(obj, field_.public_name)
 
-            if isinstance(value, (list, set)):
-                current_result = [to_json(item, **kwargs) for item in value]
+            if isinstance(value, cls.collection_field_types):
+                current_result = {
+                    "collection_type": get_full_class_name(type(value)),
```
tomsch420: make special double under fields for those and put them to the definitions where the other JSON serializer specific files are mentioned

semantic_digital_twin/src/semantic_digital_twin/world_description/world_modification.py
tomsch420: You are right, and the fix is to drop the method rather than keep it.

It looked like the frameworks `from_json` but was not one: `SubclassJSONSerializer.from_json` reads `__json_type__` out of the payload and dispatches on it, while this one hard-forced `clazz=cls` and ignored the tag, so a payload of some other class would have been read as a block without complaint.

`WorldModelModificationBlock` is a plain dataclass and its `to_json` already writes `__json_type__`, so the free `from_json` resolves it to `DataclassJSONSerializer` through the registry by itself. Removed the classmethod (and the now unused `DataclassJSONSerializer` import); both call sites in `messages.py` now just call `from_json(...)`. That is also the idiom `test_world_modifications.py` already uses on blocks (`from_json(to_json(modifications), **kwargs)`).

Verified against a real world in the container (the sdt suite still cannot collect here, pre-existing `CouldNotResolveType: MetaData` in its ORM regeneration conftest): serialize a block, `from_json` it back to a `WorldModelModificationBlock` with all three modifications, `.apply()` onto a fresh world, bodies and connections reconstruct. Full krrood suite: 2389 passed, 5 skipped.

krrood/src/krrood/adapters/json_serializer.py
tomsch420: Done — `SortedSet` is now in `list_like_classes` itself and `collection_field_types` is gone; the check is plain `isinstance(value, list_like_classes)`.

Also fixed the trailing comment on the tuple, which said these are "classes that can be serialized by the built-in JSON module" — not true of `SortedSet`, so it now reads "classes that are serialized as a JSON array".

Worth knowing this reaches past this file: `list_like_classes` is imported in about 15 places across krrood, sdt and coraplex. I went through them; every one is an "iterate and rebuild the same type" pattern (`value.__class__([...])` in `world_entity.py`, `type(obj)(... for ...)` in `world.py`, and so on), which is what you want for a `SortedSet` too — before this it fell into the scalar `else` branch there and was silently mishandled. The one real behaviour change is module-level `to_json(SortedSet(...))`, which used to raise `ClassNotSerializableError` and now writes a JSON array like `set` and `tuple` already did, so nothing could have depended on the old path.

Full krrood suite: 2389 passed, 5 skipped. probabilistic_model, which uses `SortedSet` the most (excluding test_gui, the unrelated Qt/X11 container issue): 400 passed, 1 skipped.

krrood/src/krrood/adapters/json_serializer.py
tomsch420: Done — added

```python
JSON_COLLECTION_TYPE = "__collection_type__"
JSON_COLLECTION_ITEMS = "__items__"
```

right next to `JSON_IS_CLASS` at the top of `json_serializer.py`, where `JSON_TYPE_NAME` is imported, each with a docstring saying what it holds. `to_json` and `from_json` use the constants on both sides now, as do the tests and `test_get_payload_as_json` in `test_world_fetcher.py`.

Besides being consistent with the other field names, this closes a real hole: the read side recognises a collection by `"collection_type" in current_data and "items" in current_data`, so a dataclass that happened to have fields named `collection_type` and `items` would have been reconstructed as a collection. Dunders make that impossible.

One left alone: the `dict` branch four lines below still writes plain `"keys"` and `"values"` and has exactly the same hazard. It predates this PR and renaming it would move the wire format for every dict field, so I did not touch it — say the word and I will convert it in the same style.

Full krrood suite: 2389 passed, 5 skipped.

tomsch420: Follow-up: all of them are in a `StrEnum` now — `JSONField` in the new `krrood/adapters/json_field.py`.

It holds the 16 keys this module writes: the four that were constants (`TYPE`, `IS_CLASS`, `COLLECTION_TYPE`, `ITEMS`) plus the twelve that were bare literals respelled at each read and write site — `KEYS`/`VALUES` for a `dict` field, `ATTRIBUTE_NAME`/`ADDED_VALUES`/`REMOVED_VALUES` for an attribute diff, `VALUE`, `DAYS`/`SECONDS`/`MICROSECONDS`, `MEMBER_NAME`, and `ELEMENT_TYPE`/`ELEMENTS`.

Two notes on how it landed:

- It needed its own module rather than living in `json_serializer.py`. `JSON_TYPE_NAME` sat in `adapters/exceptions.py` because `MissingTypeError` interpolates it into its message, and `json_serializer` imports the exceptions — so neither of the two can own the enum without a cycle. A leaf module both import solves it, and gives the wire format an honest home instead of an exceptions module.
- Three members are not spelled like the key they carry, because the plain name was taken or said nothing: `ELEMENT_TYPE = "type"` and `ELEMENTS = "data"` for the numpy array, whose class already goes under `TYPE`, and `MEMBER_NAME = "name"` for the name of an enum member. The values are untouched, so the JSON of an object is unchanged — verified by round-tripping a dataclass, a numpy array, a UUID and a duration and diffing the payloads, and all 16 values are distinct so nothing aliased into another member.

`KEYS` and `VALUES` still carry `"keys"` and `"values"` rather than dunders, since that rename is still the open question above.

Out of scope but worth knowing: `ros_msg_serializer.py`, `cas_annotation_codecs.py` and `sage_10k_dataset/schema.py` each still spell out their own payload keys. Those are separate wire formats and would each want their own enum, the way `SymbolicMathJSONKey` already does in `symbolic_math.py`, rather than joining `JSONField`. Happy to do them if you want.

Full krrood suite: 2389 passed, 5 skipped.

tomsch420: Right — `json_field.py` now holds only the six keys krrood itself puts into the JSON of any object, whatever its class: `TYPE`, `IS_CLASS`, `COLLECTION_TYPE`, `ITEMS`, `KEYS`, `VALUES`.

The ten that describe one particular class moved out to the serializer that writes them, each in its own enum next to it, the way `SymbolicMathJSONKey` already does in `symbolic_math.py`:

| enum | keys |
| --- | --- |
| `AttributeDiffJSONKey` | `attribute_name`, `added_values`, `removed_values` |
| `UUIDJSONKey` | `value` |
| `TimedeltaJSONKey` | `days`, `seconds`, `microseconds` |
| `EnumJSONKey` | `name` |
| `ExceptionJSONKey` | `value` |
| `NumpyArrayJSONKey` | `type`, `data` |
| `NumpyFloatJSONKey` | `value` |

Two things got better out of the split. The three serializers that each wrote `"value"` no longer share one member, so the exception reads `ExceptionJSONKey.MESSAGE` now — which is what `str(obj)` is and what the class is raised again with — while the UUID and the numpy float keep `VALUE` in their own. And `ELEMENT_TYPE`/`ELEMENTS` no longer have to dodge the class tag from a different enum; they sit in `NumpyArrayJSONKey` where `type` and `data` mean the array.

They use `enum.StrEnum` rather than a second `from enum import StrEnum`, since the module already imports `enum` and spells `enum.Enum`.

The strings are untouched, so the JSON of an object is the same — checked per serializer that the payload for a dataclass, a numpy array and float, a UUID, a duration, an exception and an attribute diff is byte for byte what it was. Full krrood suite: 2389 passed, 5 skipped.

ichumuh (approved):

Sanic (approved):

LucaKro (approved):

PR #633, opened 2026-09-16
Replace synthetic data with real robocasa objects

ganeshyadavyatham:
Replaces the synthetic EQL-generated training data with real robocasa objects, keeping the merged RSPN architecture unchanged (per-class models, same interface/flow — only data source and features differ).

Points I'd like your input on:

- **Mass excluded**: robocasa has no real per-object mass — the density is a hardcoded build-tool default (traced to `MJCFObject` `density=100` / objaverse `1000`). So I used real geometric features instead.
- **Geometry read directly, not via FeatureExtractor**: `FeatureExtractor` extracts DAO attributes (mass, center_of_mass), not geometry-derived features (volume, aspect_ratio), so I read those from the collision shapes directly. Open to extending FeatureExtractor for geometry .
- **Per-class thresholds**: kept per-class thresholds
- **Instances capped (~10/class)**: robocasa's per-object loading leaks memory on the full catalog; some classes are thin (Cup=4) or unusable (Bottle/Pan=0, SDT parser limitation). Can run full-scale on a bigger machine .

Familiar objects are accepted, anomalous ones flagged. Tests skip without robocasa assets.


experiments/scripts/generate_orm.py
```diff
@@ -18,6 +19,11 @@
 ignored_classes |= set(
     classes_of_module(experiments.control_loop_experiments.control_loop_profiler)
 )
+# robocasa loading results describe a one-off load attempt (including the raw
```
tomsch420: why?

experiments/src/experiments/confidence_aware_eql/feature_pipeline.py
```diff
@@ -36,39 +50,59 @@ class Feature(enum.StrEnum):
...
     """The object's semantic-annotation class."""
 
+    VOLUME = "volume"
+    """The object's root body collision geometry's total watertight mesh volume, in cubic meters."""
+
+    ASPECT_RATIO = "aspect_ratio"
+    """The object's height over its widest horizontal extent."""
+
+
+def _collision_volume(instance: Any) -> float:
```
tomsch420: looks like these should be classes with the enum together

experiments/src/experiments/confidence_aware_eql/robocasa_data.py
```diff
@@ -0,0 +1,119 @@
...
+from semantic_digital_twin.adapters.robocasa_dataset.semantics import (
+    RoboCasaObjectCategory,
+    RoboCasaObjectResolver,
+)
+from semantic_digital_twin.world_description.world_entity import SemanticAnnotation
+from typing_extensions import Dict, List, Type
+
+
+@dataclass
+class RoboCasaInstanceLoadFailure:
```
tomsch420: is this supposed to be an exception?

tomsch420 (changes requested):

experiments/scripts/generate_orm.py
ganeshyadavyatham: I added two small result classes in robocasa_data.py and ormatic tried to map them and first ci checks were failed. so to exclude them I have added there. those classes were not used anywhere.so I removed them . now the generate_orm.py is unchanged.

experiments/src/experiments/confidence_aware_eql/robocasa_data.py
ganeshyadavyatham: while loading instances , pan had an issue and it failed with BrokenWorldModificationHistoryError  . to avoid stopping the run , added that to make failed load to become part of return value . removed it because if an asset fails to load , it is an issue at parser level and it should raise an exception

experiments/src/experiments/confidence_aware_eql/feature_pipeline.py
ganeshyadavyatham: done updated those to Classes and feature.extract makes matching class to get dispatched

experiments/src/experiments/confidence_aware_eql/feature_pipeline.py
```diff
@@ -36,39 +57,106 @@ class Feature(enum.StrEnum):
...
+        features: Dict[Feature, Type[ObjectFeature]] = {
+            Feature.CLASS: SemanticClass,
+            Feature.VOLUME: CollisionVolume,
+            Feature.ASPECT_RATIO: AspectRatio,
+        }
+        return features[self](instance).value()
+
+
+@dataclass
+class ObjectFeature(ABC):
```
tomsch420: this should be unified with what the aggregations class is already doing.  See if you can reuse https://cram2.github.io/cognitive_robot_abstract_machine/krrood/autoapi/krrood/parametrization/feature_extraction/aggregations/index.html#krrood.parametrization.feature_extraction.aggregations.AggregationStatistic here

experiments/src/experiments/confidence_aware_eql/robocasa_data.py
```diff
@@ -0,0 +1,82 @@
...
+from semantic_digital_twin.adapters.robocasa_dataset.semantics import (
+    RoboCasaObjectCategory,
+    RoboCasaObjectResolver,
+)
+from semantic_digital_twin.world_description.world_entity import SemanticAnnotation
+from typing_extensions import Dict, List, Type
+
+
+def downloaded_instance_count(
+    loader: RoboCasaDatasetLoader, category: RoboCasaObjectCategory
```
tomsch420: looks like a method of RoboCasaDatasetLoader

experiments/src/experiments/confidence_aware_eql/robocasa_data.py
```diff
@@ -0,0 +1,82 @@
...
+    """
+    objects_directory = loader.directory / "objects"
+    return sum(
+        1
+        for group in loader.self_contained_object_groups
+        for _ in (objects_directory / group).glob(f"{category.value}/**/model.xml")
+    )
+
+
+def load_all_robocasa_objects(
```
tomsch420: sounds like a method of RoboCasaDatasetLoader

tomsch420 (changes requested):

PR #635, opened 2026-09-16
Do()'s relationship gets complicated 

Narenvasant:
This PR lets a relational circuit be fitted so that every value of a cause ends up under its own branch, on the class level and on exchangeable parts alike, which is what causal queries on it need.

- How a circuit is fitted is now a `LearningMethod`. `JointProbabilityTree` is one, taking its variables at fit time, and `StratifiedLearning` partitions the rows by the joint value of some variables and fits every partition with another learning method, combined under a sum weighted by frequency. That makes the fit support-deterministic over those variables.
- `RelationalProbabilisticCircuit` takes a `learning_method` for its class circuit and `part_learning_methods` per exchangeable part, so a stratified fit is plain configuration instead of a callback reaching into the fit.
- `ProbabilisticCircuit.log_conditional_in_place` can keep the circuit's structure while conditioning, so strata that were kept apart on purpose are not merged into one sum. Grounding a class circuit and its parts uses that.
- When grounding, a mounting node that gives no weight to the shared latent samples draws its own, and instances with the same latent values are mounted once and reused.
- A causal circuit no longer leaves an empty mixture or a second root behind when an adjustment or cause region is skipped.
- Enum columns stay enum members when a joint probability tree is fitted, so conditioning on a member matches its leaf again, and the JSON serializer keeps the enum type of a member that is also a string or an int.
- The mutagenesis experiment fits its model with `StratifiedLearning`, and the causal reasoning tests now run in CI.


probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/causal.py
```diff
@@ -44,6 +45,28 @@
 
 logger = logging.getLogger(__name__)
 
+@dataclass(frozen=True)
```
tomsch420: i think stratification should be like a learning algorithm. So this should become its own LearningMethod, that has a JPT (or any other learning methd) as a member. Then the code is better structured and you dont have to intervene with the RSPNs methods directly

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/rspn.py
```diff
@@ -605,6 +678,26 @@ class RelationalProbabilisticCircuit:
...
+    """
+    The fewest training rows a leaf of the class-level ``JointProbabilityTree`` may
+    hold, or, below one, that number as a fraction of the training rows; forwarded to
+    every exchangeable part's template fit as well.
+
+    The default lets the tree split down to one row per leaf, which pins every
+    continuous attribute to the training values it saw.
+    """
+
+    part_circuit_builders: dict[
```
tomsch420: yeah if we make a class for learning methid these fields become way better

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/rspn.py
```diff
@@ -804,8 +902,14 @@ def _condition_class_circuit(
             with the grounded exchangeable distribution.
         """
         if aggregation_statistics:
+            # Structure is preserved so a determined statistic's own leaves stay inside
```
tomsch420: remove the comment block

tomsch420 (changes requested):

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/causal.py
Narenvasant: done

probabilistic_model/src/probabilistic_model/probabilistic_circuit/relational/rspn.py
Narenvasant: done too

probabilistic_model/src/probabilistic_model/learning/learning_method.py
```diff
@@ -0,0 +1,104 @@
...
+    method, and combines them under a sum weighted by each value's relative frequency.
+
+    Every row of one partition shares the same value of those columns, so within its
+    circuit their distribution is a single point by construction, however the wrapped
+    method splits on the remaining columns: the fitted circuit is support-deterministic
+    over the columns, the precondition a causal query on them needs. A single
+    unconstrained fit could instead spread rows sharing a value across sibling leaves.
+    """
+
+    columns: Sequence[str]
```
tomsch420: name this variables

probabilistic_model/src/probabilistic_model/learning/learning_method.py
```diff
@@ -0,0 +1,104 @@
...
+        """
+
+
+@dataclass
+class JointProbabilityTreeLearning(LearningMethod):
+    """
+    Fits a :class:`~probabilistic_model.learning.jpt.jpt.JointProbabilityTree`.
+    """
+
+    min_samples_per_leaf: float = 1
```
tomsch420: JPT itself is the subclass of learning method you should use here

tomsch420 (changes requested):

probabilistic_model/src/probabilistic_model/learning/learning_method.py
Narenvasant: done

probabilistic_model/src/probabilistic_model/learning/learning_method.py
Narenvasant: done

probabilistic_model/src/probabilistic_model/learning/jpt/jpt.py
```diff
@@ -243,13 +246,27 @@ def preprocess_data(self, data: pd.DataFrame) -> np.ndarray:
...
-        Fit the model to the data.
+        Fit the model to the data, into a circuit of its own.
 
         :param data: The data to fit the model to.
+        :param variables: The annotated variables to fit over, replacing the ones given
+            at initialization, with all of them as targets and features. ``None``
+            keeps the initialized ones.
         :return: The fitted model.
         """
+        if variables is not None:
```
tomsch420, line 271: in the case that there are no variables, can you infer them from the dataframe?

tomsch420 (changes requested):

krrood/src/krrood/parametrization/feature_extraction/feature_extractor.py
```diff
@@ -275,24 +275,26 @@ def create_dataframe(self, instances: list[DataAccessObject]) -> pd.DataFrame:
...
 
         :param df: The dataframe to preprocess.
         :return: The dataframe in a JPT compatible format.
+        :raises UnsupportedFeatureTypeError: If a column's type cannot be fitted on.
         """
         feature_map = dict(zip(df.columns, self.features))
         for column in df.columns:
             feature = feature_map[column]
             if isinstance(feature._type_, enum.EnumType):
-                df[column] = df[column].apply(lambda x: hash(x))
```
davidprueser, line 294: yes

probabilistic_model/src/probabilistic_model/learning/jpt/jpt.py
```diff
@@ -32,15 +33,17 @@
...
 
 @dataclass
-class JointProbabilityTree(SubclassJSONSerializer):
+class JointProbabilityTree(LearningMethod, SubclassJSONSerializer):
     """
     Class that implements the JPT learning algorithm for probabilistic circuits.
     """
 
-    annotated_variables: Iterable[AnnotatedVariable]
+    annotated_variables: Iterable[AnnotatedVariable] = field(default=())
```
davidprueser: I think using field(default_factory=tuple) would be more consistent, as we use default_factory for every other collection type. Also, if it is always tuple, you could do tuple[AnnotatedVariable, ...]

probabilistic_model/src/probabilistic_model/learning/jpt/jpt.py
Narenvasant: done

tomsch420 (approved):

.github/workflows/ci.yml
```diff
@@ -39,6 +39,7 @@ jobs:
           - lib: robokudo
           - lib: segmind
           - lib: experiments
+          - lib: causal_reasoning
```
davidprueser: Are we sure we want an independent  package for causal reasoning? It is still knowledge reasoning, which is part of what krrood does in theory

davidprueser (changes requested):

probabilistic_model/src/probabilistic_model/learning/jpt/jpt.py
Narenvasant: done

.github/workflows/ci.yml
Narenvasant: It is not a package. causal_reasoning here is only the name of the test folder test/causal_reasoning_test, which was already on main and holds the tests of the experiments under experiments/causal_reasoning. This line just makes CI run them. They have their own small conftest because the experiments_test conftest needs ROS and builds the whole ORM chain, which these tests do not need. I added a comment next to the line saying that. The code itself lives in krrood and probabilistic_model, nothing new was split off.

.github/workflows/ci.yml
```diff
@@ -39,6 +39,9 @@ jobs:
           - lib: robokudo
           - lib: segmind
           - lib: experiments
+          # test/causal_reasoning_test: the tests of the experiments under
+          # experiments/causal_reasoning, kept out of test/experiments_test because
+          # its conftest needs ROS and the whole ORM chain, which these do not
```
davidprueser: Oh I see, it is about the tests. I still don't like that there is a causal_reasoning test folder at the tests root, even though the tests are experiment-related. The issue is that most tests are directly at the root of the experiments_test directory and @LucaKro defined fixtures needed for his own tests in the conftest file. I think it would be better if each experiment had its tests in a dedicated test directory within the experiments_test package, with each directory defining its own conftest if needed. This way, ROS-dependent and ROS-independent experiment tests can co-exist and the causal reasoning tests can move back into the experiments_test

davidprueser (changes requested):

.github/workflows/ci.yml
Narenvasant: ah i see what you mean. yeah fixed them

davidprueser (approved):

PR #637, opened 2026-09-17
Read meshes from a dataset server instead of only from local files

tomsch420:
The flexible mesh backend using neem4 server as discussed on tuesday. This can hardly be unit tested but i included mocks as far as possible. I tested it on my laptop without having the meshes, and it correctly fetched them from the server. Even works with RVIZ as they are cached locally


semantic_digital_twin/src/semantic_digital_twin/adapters/dataset_server.py
```diff
@@ -0,0 +1,206 @@
...
+    The address serving :attr:`dataset_root`.
+    """
+
+    dataset_root: PurePosixPath
+    """
+    The dataset's location on the machine holding it, which is what the references
+    recorded against it are written relative to.
+    """
+
+    cache: Path
```
tomsch420: i think we have a cache default path somewhere in semdt already. Check if you can default to that too

semantic_digital_twin/src/semantic_digital_twin/adapters/dataset_server.py
```diff
@@ -0,0 +1,206 @@
...
+    The directory the dataset's files are copied into on this machine.
+    """
+
+    session: requests.Session = field(default_factory=requests.Session)
+    """
+    The connection the requests are made on, reused so a directory and its files cost
+    one handshake between them rather than one each.
+    """
+
+    timeout: float = 30.0
```
tomsch420: use timedelta

semantic_digital_twin/src/semantic_digital_twin/adapters/dataset_server.py
```diff
@@ -0,0 +1,206 @@
...
+    """
+    Seconds to wait for a single request.
+    """
+
+    completion_marker: ClassVar[str] = ".entry-complete"
```
tomsch420: make this an ordinary field

semantic_digital_twin/src/semantic_digital_twin/world_description/mesh_file_storage.py
```diff
@@ -10,7 +10,70 @@
...
 import psutil
 from krrood.singleton import SingletonMeta
-from typing_extensions import ClassVar
+from typing_extensions import ClassVar, List, Protocol, runtime_checkable
+
+# %% where a mesh file is read from
+
+
+@runtime_checkable
+class MeshFileSource(Protocol):
```
tomsch420: dont make any protocols.

test/semantic_digital_twin_test/test_adapters/dataset_server_fixture/mesh_store/ab/ab5f1c3d9e2b4a6c8d0f1e2a3b4c5d6e7f809a1b/material.mtl
tomsch420, line 1: can you use existing files in the repo instead of these?

semantic_digital_twin/src/semantic_digital_twin/adapters/dataset_server.py
tomsch420: There was one: `create_cache_dir` in `semantic_digital_twin/utils.py`, which `pybullet_collision_detector` already uses for `convex_decompositions` and `log`. The field now defaults to `create_cache_dir(CACHE_FOLDER_NAME)`, so the copied entries land beside everything else the package downloads instead of under a second hand-written path. `from_environment` only overrides it when `SEMANTIC_DIGITAL_TWIN_MESH_CACHE` is set, so the default has one definition.

semantic_digital_twin/src/semantic_digital_twin/adapters/dataset_server.py
tomsch420: Done: `timeout: timedelta = timedelta(seconds=30)`, converted with `total_seconds()` at the single `requests` call.

semantic_digital_twin/src/semantic_digital_twin/adapters/dataset_server.py
tomsch420: Done, it is a plain field now. The two tests that read it off the class read it off the instance.

semantic_digital_twin/src/semantic_digital_twin/world_description/mesh_file_storage.py
tomsch420: Dropped. `MeshFileSources.sources` is now `List[PathResolver]` — the ABC in `adapters/package_resolver.py`, which is the abstraction this repository already has for answering a reference with a local path.

I had used a protocol to keep `world_description` from importing `adapters`, but that direction costs nothing: `package_resolver` imports only `exceptions`, and `adapters/__init__.py` is empty. I checked imports from all four entry points, no cycle. `DatasetServer` now satisfies one abstraction rather than structurally matching a second one saying the same thing.

test/semantic_digital_twin_test/test_adapters/dataset_server_fixture/mesh_store/ab/ab5f1c3d9e2b4a6c8d0f1e2a3b4c5d6e7f809a1b/material.mtl
tomsch420: Yes. The served entry is now built at test time from `resources/ply/chair.ply` and `chair_texture.png` through the repository's own `Mesh.from_ply_file`, which exports exactly the shape the test needs: `chair.obj` with `mtllib material.mtl` and `usemtl chair_texture`, and those two files beside it. The several-files-per-mesh case is still what is exercised, and nothing new is checked in.

It also takes the negated `.gitignore` rules back out, which the fixture needed because `*.png` and `*.obj` are both ignored repository-wide. That was the part I liked least about it: the fixture would have been silently absent in CI while passing locally.

semantic_digital_twin/doc/datasets.md
```diff
@@ -135,3 +135,47 @@ camera (and, with `with_world_frame=True`, a `map` root body placing the camera
...
+```nginx
+location /datasets/ {
+    alias /path/to/datasets/;
+    autoindex on;
+    autoindex_format json;
+    add_header Cache-Control "public, max-age=31536000, immutable";
+}
+```
+
+The mount is read-only and nothing is ingested or indexed, so adding to the dataset stays
```
tomsch420: remove this last paragraph

semantic_digital_twin/src/semantic_digital_twin/adapters/dataset_server.py
```diff
@@ -0,0 +1,209 @@
...
+        """
+        :param data: One element of a listing.
+        :return: The item it describes.
+        """
+        return cls(name=data["name"], kind=ListedItemKind(data["type"]))
+
+
+# %% reaching a dataset that is served over http
+
+CACHE_FOLDER_NAME = "dataset_server"
```
tomsch420: dont declare module level variables. Just write this in the default factory directly

test/semantic_digital_twin_test/test_adapters/listing_file_server.py
```diff
@@ -0,0 +1,109 @@
...
+) -> Type[SimpleHTTPRequestHandler]:
+    """
+    Build a handler serving a directory tree, answering a directory with a json listing.
+
+    :param root: The directory tree to serve.
+    :param requested_paths: The list every requested path is appended to.
+    :return: The handler class.
+    """
+
+    class Handler(SimpleHTTPRequestHandler):
```
tomsch420: can this be a dataclass?

tomsch420: also its weird to declare this in a function

test/semantic_digital_twin_test/test_adapters/listing_file_server.py
```diff
@@ -0,0 +1,109 @@
...
+            """
+
+    return Handler
+
+
+# %% the server
+
+
+@dataclass
+class ListingFileServer:
```
tomsch420: in the name say that these are mocks, otherwise they may be confused

semantic_digital_twin/doc/datasets.md
tomsch420: Removed.

semantic_digital_twin/src/semantic_digital_twin/adapters/dataset_server.py
tomsch420: Inlined: `field(default_factory=lambda: create_cache_dir("dataset_server"))`.

The one consequence is that the test asserting the default now spells `"dataset_server"` a second time, so the two can drift. Say if you would rather have that test drop the exact path and assert only that the default is the directory `create_cache_dir` hands out.

test/semantic_digital_twin_test/test_adapters/listing_file_server.py
tomsch420: Not as a dataclass: the server constructs a handler per request with the three arguments `BaseRequestHandler` fixes, and that constructor is what serves the request, so a generated `__init__` would replace the thing doing the work.

What it can be is an ordinary module level class, which is what it is now. The directory and the recording list reach it through the `partial` the server is built with, so nothing is captured from an enclosing scope.

tomsch420: Agreed, and it no longer is. `MockDatasetFileRequestHandler` sits at module level; the factory function is gone.

test/semantic_digital_twin_test/test_adapters/listing_file_server.py
tomsch420: Renamed: `MockDatasetFileServer` and `MockDatasetFileRequestHandler`, in `mock_dataset_file_server.py`.

I kept "file" in the name so it does not read as a mock of `DatasetServer`, which is the client. This one stands for the machine serving the files.

semantic_digital_twin/doc/datasets.md
```diff
@@ -135,3 +135,42 @@ camera (and, with `with_world_frame=True`, a `map` root body placing the camera
...
+
+```python
+WorldSpecification.from_urdf(path, path_resolver=CompositePathResolver([server]))
+```
+
+Anything that serves a directory tree and answers a directory with a json listing can be
+the server. nginx does both without code:
+
+```nginx
+location /datasets/ {
```
tomsch420: dont include the nginx command here. Thats better server by their doc. Just say that nginx is good for that.

tomsch420: Done. It now names `autoindex` and `autoindex_format` so there is something to look up, and leaves the configuration to nginx's own documentation.

tomsch420: Added: both directives now link to their entries in nginx's `ngx_http_autoindex_module` documentation.

semantic_digital_twin/src/semantic_digital_twin/exceptions.py
```diff
@@ -1334,6 +1335,35 @@ def error_message(self) -> str:
...
+    def suggest_correction(self) -> str:
+        if self.status_code == HTTPStatus.NOT_FOUND:
+            return (
+                "check that the dataset server serves the directory this path is under."
+            )
+        return (
+            "check that the dataset server is reachable and serving the dataset root."
+        )
+
     def suggest_correction(self) -> str:
```
Narenvasant: I think this should be removed

Narenvasant (changes requested):
Nice

semantic_digital_twin/src/semantic_digital_twin/exceptions.py
tomsch420: its an abstract mehtod that i have to implement. There is just not a suggestable correction

Narenvasant: there is already def suggest_correction just above this, i felt this overrides the other

tomsch420: You are right, and it was a real bug — thank you for pushing on it.

`DatasetServerError` had been inserted between `PathResolutionError`'s `error_message` and its `suggest_correction`, so that `suggest_correction` ended up inside the new class. Being the later definition, it shadowed the one the class declares itself, and `DatasetServerError.suggest_correction()` returned `""`. So the advice above it was never reachable, exactly as you said.

Fixed in 1961b4ee: the method is back on `PathResolutionError`, and the diff against main now adds `DatasetServerError` whole rather than reaching into the class before it.

Narenvasant (approved):

LucaKro (approved):

PR #638, opened 2026-09-18
Bug fix on causal circuit

Narenvasant:
A bug in the support determinism check of `CausalCircuit`, with tests that fail without the fix. The string bug in random_events is handled in #640.

- The check only noticed a sum unit when at least one pair of its children was fully disjoint on the cause variable. Children that overlap without being equal, like `{a, b}` and `{a, b, c}`, were never looked at, so the circuit passed and `backdoor_adjustment` returned a number that was not an interventional probability.
- A sum unit now counts as splitting on the cause as soon as its children differ, and overlapping children are then rejected as they should be.
- The overlap check also treated two continuous leaves that only touch at an endpoint, like `[0, 1]` and `[1, 2]`, as overlapping. Every fitted tree produces neighbours like that, so valid circuits were rejected. An intersection now only counts as overlap when it is more than single points.
- A sum unit with fewer than two children has nothing to split on and is reported as not splitting.
- One existing krrood test only passed through the first hole. It is rewritten so every partition holds both object types and the circuit is genuinely support deterministic.
- New tests cover the overlapping children, the touching intervals, and the units with fewer than two children.


probabilistic_model/src/probabilistic_model/probabilistic_circuit/causal/causal_circuit.py
```diff
@@ -399,32 +399,59 @@ def _child_marginals_split_on_variable(
...
+        :param query_variable: The variable the marginals are restricted to.
         :returns: True if any pair overlaps.
         """
         return any(
-            not child_marginals[i].intersection_with(child_marginals[j]).is_empty()
+            cls._has_extent(
+                child_marginals[i].intersection_with(child_marginals[j]),
+                query_variable,
+            )
             for i, j in itertools.combinations(range(len(child_marginals)), 2)
```
davidprueser: Abbreviations spotted

probabilistic_model/src/probabilistic_model/probabilistic_circuit/causal/causal_circuit.py
```diff
@@ -399,32 +399,59 @@ def _child_marginals_split_on_variable(
...
 
         :param child_marginals: Marginal support events, one per SumUnit child.
-        :returns: True if any pair of marginals is disjoint.
+        :returns: True if any two marginals differ.
         """
-        return any(
-            child_marginals[i].intersection_with(child_marginals[j]).is_empty()
-            for i, j in itertools.combinations(range(len(child_marginals)), 2)
-        )
+        first, *others = child_marginals
```
davidprueser, line 418: Can child_marginals be empty? If so, it would raise a ValueError here

random_events/src/random_events/variable.py
```diff
@@ -145,7 +145,8 @@ def _from_cpp(cls, cpp_object):
         )
 
     def make_value(self, value) -> Set:
-        if not isinstance(value, Iterable):
+        # a string is one symbol, not the sequence of its characters
+        if not isinstance(value, Iterable) or isinstance(value, str):
```
davidprueser: I don't get it, a string is an Iterable?

davidprueser (changes requested):

probabilistic_model/src/probabilistic_model/probabilistic_circuit/causal/causal_circuit.py
Narenvasant: Replaced "e.g." with "for example".

probabilistic_model/src/probabilistic_model/probabilistic_circuit/causal/causal_circuit.py
Narenvasant: The only caller skips sum units with fewer than two children, so it could not be empty in practice, but the unpacking would indeed raise on an empty list. It now returns False for zero or one marginals, since such a unit has nothing to split on, with a test for both cases and a sentence in the docstring.

random_events/src/random_events/variable.py
Narenvasant: Yes, it is. A string iterates over its characters, so `isinstance("absent", Iterable)` is True and the old code turned `make_value("absent")` into the six symbols `a`, `b`, `s`, `e`, `n`, `t`. That also hits every StrEnum member. The check now tests for str first and the comment says why.

probabilistic_model/src/probabilistic_model/probabilistic_circuit/causal/causal_circuit.py
```diff
@@ -399,32 +399,59 @@ def _child_marginals_split_on_variable(
...
+            do, share no extent.
+        """
+        for simple_event in event.simple_sets:
+            value = simple_event[query_variable]
+            if isinstance(value, Interval):
+                if any(
+                    interval.lower < interval.upper for interval in value.simple_sets
+                ):
+                    return True
+            elif not value.is_empty():
```
tomsch420: value.is_empty() cannot be true per construction

probabilistic_model/src/probabilistic_model/probabilistic_circuit/causal/causal_circuit.py
```diff
@@ -399,32 +399,59 @@ def _child_marginals_split_on_variable(
...
+        :param query_variable: The variable the event restricts.
+        :return: Whether the event carries more than a boundary: a set of symbols that
+            is not empty, or an interval of positive length. Two continuous regions
+            that only touch at an endpoint, as neighbouring leaves of a fitted tree
+            do, share no extent.
+        """
+        for simple_event in event.simple_sets:
+            value = simple_event[query_variable]
+            if isinstance(value, Interval):
+                if any(
```
tomsch420: i think there is a property like is_singleton or so

tomsch420: or you could just generally check the volum of an event, that makes it way easier

tomsch420 (changes requested):

probabilistic_model/src/probabilistic_model/probabilistic_circuit/causal/causal_circuit.py
davidprueser: I meant i, j

probabilistic_model/src/probabilistic_model/probabilistic_circuit/causal/causal_circuit.py
Narenvasant: you're right. removed

probabilistic_model/src/probabilistic_model/probabilistic_circuit/causal/causal_circuit.py
Narenvasant: Used is_singleton on each simple interval.

probabilistic_model/src/probabilistic_model/probabilistic_circuit/causal/causal_circuit.py
```diff
@@ -405,15 +405,15 @@ def _child_marginals_split_on_variable(
...
+        example a sibling SumUnit in a ProductUnit that has no relationship to this
+        variable) are not split nodes and must be skipped to avoid false positives. A
+        SumUnit with fewer than two children has nothing to split on.
 
         :param child_marginals: Marginal support events, one per SumUnit child.
         :returns: True if any two marginals differ.
         """
-        first, *others = child_marginals
-        return any(other != first for other in others)
+        return any(marginal != child_marginals[0] for marginal in child_marginals[1:])
```
davidprueser: This would still fail with an Indexerror if child_marginals is empty

random_events/src/random_events/variable.py
```diff
@@ -145,8 +145,8 @@ def _from_cpp(cls, cpp_object):
         )
 
     def make_value(self, value) -> Set:
-        # a string is one symbol, not the sequence of its characters
-        if not isinstance(value, Iterable) or isinstance(value, str):
+        # a string is iterable over its characters, but it names one symbol
```
davidprueser: Why is this needed?

davidprueser (changes requested):

probabilistic_model/src/probabilistic_model/probabilistic_circuit/causal/causal_circuit.py
```diff
@@ -399,32 +399,57 @@ def _child_marginals_split_on_variable(
...
         """
-        return any(
-            child_marginals[i].intersection_with(child_marginals[j]).is_empty()
-            for i, j in itertools.combinations(range(len(child_marginals)), 2)
-        )
+        return any(marginal != child_marginals[0] for marginal in child_marginals[1:])
 
     @staticmethod
-    def _overlapping_pair_exists(child_marginals: List[Any]) -> bool:
+    def _has_extent(event: Event, query_variable: Variable) -> bool:
```
tomsch420: remove this method. You can check the exact same by asking `event.size > 0`  in _overlapping_pair_exists

tomsch420 (changes requested):

random_events/src/random_events/variable.py
Narenvasant: this is bug is taken over by Bass in his other pr (640)

probabilistic_model/src/probabilistic_model/probabilistic_circuit/causal/causal_circuit.py
Narenvasant: pls check now

probabilistic_model/src/probabilistic_model/probabilistic_circuit/causal/causal_circuit.py
Narenvasant: it is now an if len(child_marginals) < 2: return False before the unpacking. but tbh it didnt fail in practice though

probabilistic_model/src/probabilistic_model/probabilistic_circuit/causal/causal_circuit.py
Narenvasant: done

tomsch420 (approved):

davidprueser (approved):

PR #639, opened 2026-09-18
[Ormatic] An alternative mapping is converted after the mappings it holds

AbdelrhmanBassiouny:
## What was wrong

Full detail: https://github.com/AbdelrhmanBassiouny/cognitive_robot_abstract_machine/pull/374


tomsch420 (changes requested):
No this is not doable. I thought about doing it this way, but that means a toplogical sort of the dependencies must exist, which is not the case. Hence the solution with the alternative mappings specifying their dependencies. Just declare at the PlanMapping that this requires world

tomsch420 (approved):

Narenvasant (approved):

PR #642, opened 2026-09-18
[Ormatic] Keep the concrete enum of a value stored in a polymorphic enum column

AbdelrhmanBassiouny:
Handing a concrete entity that carries an enum field to a query the `ProbabilisticBackend` resolves fails with

Full detail: https://github.com/AbdelrhmanBassiouny/cognitive_robot_abstract_machine/pull/357


test/krrood_test/conftest.py
```diff
@@ -110,7 +112,12 @@ def generate_sqlalchemy_interface():
     instance = ORMatic(
         class_dependency_graph=class_diagram,
         interface_information=OrmaticInterfaceInformation(
-            type_mappings=TypeDict({KRROODPhysicalObject: ConceptType}),
+            type_mappings=TypeDict(
+                {
```
tomsch420: I am unsure if this is a true problem. PolymorphicEnumType should always be used for enums in ORM anyways

tomsch420 (changes requested):
I am unsure if this is a true problem. PolymorphicEnumType should always be used for enums in ORM anyways

test/krrood_test/conftest.py
AbdelrhmanBassiouny: Well yes that's the case, but we need the more specific enum type and that was the issue that happened in the feature extractor if I understand correctly, it was able to only get it by checking the value type of the column instead of just getting the super Enum class.

AbdelrhmanBassiouny: Ok I am also insure why claude adjusted the type mappings here, this should have just been a feature extractor fix if I understand correctlym

AbdelrhmanBassiouny: Yes you are right, this is unnecessary, removed it.

tomsch420 (approved):

LucaKro (approved):

PR #645, opened 2026-09-18
[RDR] Say why an RDR model failed to load, and load one from anywhere

AbdelrhmanBassiouny:
Two faults in the RDR model load path, both found while root-causing the red `test_each_lib (krrood)` check. **The test-ordering half of that failure is #251's, not this one's** — see below.

Full detail: https://github.com/AbdelrhmanBassiouny/cognitive_robot_abstract_machine/pull/269


ichumuh (approved):

krrood/src/krrood/utils.py
```diff
@@ -169,29 +169,49 @@ def is_builtin_type(type_object: Any):
     )
 
 
+def get_import_root_from_path(path: str) -> str:
+    """
```
tomsch420, line 173: Use path object

krrood/src/krrood/utils.py
```diff
@@ -169,29 +169,49 @@ def is_builtin_type(type_object: Any):
...
+    :return: The Python import path, or None if the path is not inside a package.
+    """
+    absolute_path = os.path.abspath(path)
+    root = get_import_root_from_path(absolute_path)
+    if root == absolute_path:
+        return None
+    return os.path.relpath(absolute_path, root).replace(os.path.sep, ".")
+
+
+def make_path_importable(path: str) -> None:
```
tomsch420: Use paths everywhere

tomsch420 (changes requested):

LucaKro (approved):
just toms comments

krrood/src/krrood/utils.py
AbdelrhmanBassiouny: Done

krrood/src/krrood/utils.py
AbdelrhmanBassiouny: Done

tomsch420 (approved):

PR #636, opened 2026-09-16
[Experiments] Shelf schema and sage10k preprocessing pipeline

davidprueser:
## Summary

This PR is the foundation for follow-up PRs that will contain the actual thesis experiments. It introduces the general shelf schema (`shelf_schema.py`, `shelf_schema_aggregations.py`) used throughout those experiments, and the preprocessing pipeline (`preprocessing/preprocess_sage10k.py`) that turns the raw sage10k dataset into the shape that schema expects. No experiment logic itself lives here yet.

### Preprocessing pipeline

`Sage10kPreprocessingRun` reads the raw sage10k dataset once and writes a processed, fitting-ready copy of it, applying every correction shelf-layout fitting used to otherwise repeat on each run:

- **Extracting shelf-relevant objects** – `ShelfMembershipClassifier` decides which raw furniture names are shelf-like (bookcase, cabinet, rack, ...), and `ShelfContents` then keeps only those shelves and whatever is placed on them, streaming every other object straight through instead of holding the whole dataset in memory.
- **Unifying object categories** – `ObjectTypeClassifier` maps the dataset's free-form, near-instance-specific type strings (e.g. `"bookchair8eba7fdc"`) onto a fixed `ObjectType` enum, so downstream code reasons about a closed set of categories instead of arbitrary strings.
- **Correcting object positions** – a sage10k object's recorded position is its mesh's local origin, not necessarily its centre. `MeshMeasurements` loads and measures each shelf-relevant mesh's bounding box and corrects the recorded position onto the mesh's true horizontal centre, rotated into world axes by the object's own yaw.
- **Deriving a shelf's use-case** – a shelf is no longer classified by its furniture name; `theme_dominant_type` is derived instead from the most frequent `ObjectType` actually standing on it, which better reflects what the shelf is really used for.
- **Extracting layers** – objects standing on a shelf are clustered by height into horizontal layers, ordered bottom to top, each recording where it sits between the shelf's base and top and how much vertical clearance it has above it.
- **Translating object positions into local shelf space** – each object's pose is re-expressed relative to its shelf, in the shelf's own content frame (its yaw plus a fixed content-frame offset), so a shelf's contents can be learned and reasoned about independently of where the shelf itself stands in the room.

Mesh measurement and the read-convert-write object pass each run across multiple worker processes for throughput; only the summary counts are gathered back into the main process.

PR #651, opened 2026-09-18
[KRROOD] Fix DAO lookup for domain classes reloaded as a spawn worker's entry point

davidprueser:
## Problem

`get_dao_class()` (and `to_dao()`, which relies on it) matched a domain class to its
registered DAO by plain identity (`subclass.original_class() == original_clazz`).

That breaks down for any class defined in a module that gets re-executed under one of
Python's two synthetic entry-point module names:

- `__main__`, when the module is run as a script or via `python -m`;
- `__mp_main__`, which the `spawn` multiprocessing start method uses to re-execute that
  same script inside every worker it starts (so a worker's own re-import doesn't collide
  with a genuinely separate `__main__` there).

Concretely, take a script `build_widgets.py`:

```python
# build_widgets.py
from concurrent.futures import ProcessPoolExecutor
from multiprocessing import get_context

class Widget:
    pass

class WidgetDAO(DataAccessObject[Widget]):
    ...  # registered once, at import time

def build_and_convert() -> DataAccessObject:
    return to_dao(Widget())

if __name__ == "__main__":
    with ProcessPoolExecutor(mp_context=get_context("spawn")) as pool:
        pool.submit(build_and_convert).result()
```

Running `python -m build_widgets` re-executes the whole file under the name `__main__`
in the parent process, so `Widget` there has `__module__ == "__main__"`. `spawn` then
re-executes the *same file* in the worker, under `__mp_main__` this time, producing a
**second, distinct** `Widget` class object — same class body, different object, because
the module was loaded a second time under a different name.

`build_and_convert()` runs inside that worker, so the `Widget()` it builds is an instance
of the worker's own `Widget` — not the one `WidgetDAO` was registered against when the
parent process first imported the module. `get_dao_class(type(widget))` compared them by
identity, found no match, and returned `None`; `to_dao()` then raised `NoDAOFoundError`
for every single object any spawned worker built, regardless of whether a DAO was in fact
registered for its class.

This is not a hypothetical: it's exactly what broke
`preprocess_sage10k_for_training`'s worker pool in practice, and is the reason that
pipeline now has to be run through a slim wrapper module rather than `-m` on the library
module directly.

## Fix

`_same_domain_class()` falls back to matching by defining file and qualified name when a
straight identity comparison fails *and* one of the two classes was loaded under
`__main__`/`__mp_main__`. Two classes that merely happen to share a name in genuinely
different, normally imported modules are never conflated, because the entry-point check
gates the fallback.

## Tests

Added to `test/krrood_test/test_ormatic/test_dao_conversion.py`:

- `test_dao_lookup_matches_domain_class_reexecuted_as_spawn_worker_entry_point` —
  reproduces the bug directly: builds a second class object standing in for the
  `__mp_main__` reload and asserts `get_dao_class()` still resolves it to the DAO
  registered against the normally imported class. Fails with `AssertionError: assert None
  is _SpawnWorkerDomainClassDAO` on the pre-fix code.
- `test_dao_lookup_does_not_conflate_unrelated_same_named_classes` — guards the negative
  case the docstring calls out: two same-named classes in genuinely different, normally
  imported modules must not be conflated.

Full `test/krrood_test/test_ormatic/` suite (142 tests) passes.

## Relation to other branches

This fix currently ships bundled inside `shelf_schema_preprocessing` (PR #636), which
only needed it incidentally while extracting an unrelated pipeline. Once this lands,
that PR will be rebased to depend on it instead of duplicating the change.

Made with the help of Claude.


AbdelrhmanBassiouny (approved):

LucaKro (approved):
whats __mp_main__?

tomsch420 (approved):

tomsch420:
> whats **mp_main**?

name of main process in multiprocessing stuff

PR #670, opened 2026-09-22
Added SegMind to the Bullet World demo

sorinar329:
- is_supported_by reworked: support now comes from an actual near-contact point (within a 5mm resting tolerance) below the supported body's own center of mass, read from the real collision detector, instead of a center-of-mass comparison plus bounding-box overlap. Fixes false positives where a container's or wall's bounding box reached past its actual walls, and contact() now searches out to its given threshold instead of ignoring it.
- SegMind added as a live event segmenter: Segmind/LiveSegmenter tick a set of detectors against the world while a plan runs it rather than after the fact, resolving each requested detector's own dependencies automatically (asking for pick-ups brings support and translation detectors along)
- Context gained a segment_events flag so a plan states whether it wants to be watched this way, instead of every run paying for it.
- The bullet-world demo now runs Segmind live across its milk/bowl/spoon transport plan, watching for pick-up, placing, support, and containment events as the plan executes.
- GraspEvent/LossOfGraspEvent added: an agent's own hand taking hold of a body and letting go of it again, read as its own relation rather than folded into support or contact.

Made with the help of Claude


coraplex/demos/coraplex_bullet_world_demo/demo.py
```diff
@@ -120,5 +136,15 @@
     context=context,
 ).plan
 
-with simulated_robot:
+segmentation = (
+    Segmind.watching_bodies_named(
```
LucaKro, line 140: I dont the name should be the key here. I think it should either be the type of the semantic annotation of the semantic annotation instance (i feel like type would for better, but also both could be supported)

coraplex/demos/coraplex_bullet_world_demo/demo.py
```diff
@@ -120,5 +136,15 @@
...
+        world,
+        ("milk.stl", "bowl.stl", "spoon.stl"),
+        detectors=(PickUpDetector, PlacingDetector, ContainmentDetector, GraspDetector),
```
LucaKro, line 143: just from reading this, this reminds me of the estimators in AICON

coraplex/src/coraplex/datastructures/dataclasses.py
```diff
@@ -84,6 +84,14 @@ class Context(PlanEntity):
     Should pre -and postconditions of actions be evaluated in this plan.
     """
 
+    segment_events: bool = field(default=True)
+    """
+    Should what happens while this plan runs be segmented into events.
```
LucaKro, line 89: > Should what happens

i dont understand

segmind/src/segmind/datastructures/events.py
```diff
@@ -131,13 +130,14 @@ class MotionEvent(EventWithTrackedObjects, ABC):
     vice versa.
     """
 
-    start_pose: Pose = field(default_factory=Pose)
+    start_pose: Pose = field(kw_only=True)
     """
-    The pose of the object at the start of the event.
+    The pose of the object at the start of the event, in the world frame.
```
LucaKro, line 135: name this variable world_T_start_pose then, so people instantly see what reference frame you expect. same with "current pose" below etc

segmind/src/segmind/datastructures/events.py
```diff
@@ -248,6 +248,23 @@ class LossOfContactEvent(AbstractContactEvent):
     ...
 
 
+@dataclass(unsafe_hash=True)
+class GraspEvent(EventWithTrackedObjects):
```
LucaKro, line 252: from the code it seems this event and LossOfGraspEvent are literally the same. How do you differentiate between them such they dont always fire at the same time?

segmind/src/segmind/detectors/agent_event_detector_nodes.py
```diff
@@ -0,0 +1,155 @@
...
+)
+from segmind.detectors.base import AbstractDetector, IndexedBodyPairs, SegmindContext
+from semantic_digital_twin.reasoning.predicates import contact
+from semantic_digital_twin.robots.robot_part_mixins import HasTwoFingers
+from semantic_digital_twin.robots.robot_parts import EndEffector
+from semantic_digital_twin.world_description.world_entity import Body
+
+
+@dataclass(eq=False, repr=False)
+class AbstractGraspDetector(AbstractDetector, ABC):
```
LucaKro, line 31: i feel like these should be predicates in semdt

segmind/src/segmind/detectors/base.py
```diff
@@ -91,19 +108,44 @@ class SegmindContext(ContextExtension):
...
     """
 
+
 @dataclass(repr=False, eq=False)
 class AbstractDetector(MotionStatechartNode, ABC):
     """
     Abstract base class for all detectors.
     """
 
+    requires: ClassVar[Tuple[Type[AbstractDetector], ...]] = ()
```
LucaKro, line 118: no classvars

segmind/src/segmind/datastructures/events.py
LucaKro: okay i guess i see in the detectors

segmind/src/segmind/detectors/atomic_event_detectors_nodes.py
LucaKro, line 72: hmm feel like LossOfContactDetector and ContactDetector are super related to the grasp detectors. shouldnt the grasp detectors maybe do their detection based on the contact detectors as well?

LucaKro, line 192: hmm maybe my feeling is wrong, but i feel like the difference between predicates that go in semdt and stuff that segmind does is that predicates are always computed on the current state, and segmind always computes stuff on a sequence of past states/events?

segmind/src/segmind/detectors/base.py
```diff
@@ -126,34 +168,123 @@ def on_tick(self, context: MotionStatechartContext) -> Optional[ObservationState
...
+            if not gone:
+                continue
+            relations -= gone
+            if not relations:
+                remembered.pop(body)
+            lost[body] = gone
+        return lost
+
+    @staticmethod
+    def bodies_of_robots(world: World) -> Set[Body]:
```
LucaKro, line 219: i think we already have sth like this in world

segmind/src/segmind/detectors/coarse_event_detector_nodes.py
```diff
@@ -29,19 +47,46 @@ class AbstractInteractionDetector(AbstractDetector):
...
+
     def _find_interaction_events(
-            self,
-            segmind_context: SegmindContext,
-            primary_event_type: type,
-            secondary_event_type: type,
-            make_event: Callable[[Any, Any], DetectionEvent],
+        self,
+        segmind_context: SegmindContext,
+        primary_event_type: type,
```
LucaKro, line 78: proper type hints

segmind/src/segmind/detectors/coarse_event_detector_nodes.py
```diff
@@ -29,19 +47,46 @@ class AbstractInteractionDetector(AbstractDetector):
...
+        secondary_event_type: type,
+        make_event: Callable[[Any, Any], DetectionEvent],
```
LucaKro, line 80: is there a way to avoid passing callables? just using a normal datastructure seems more clean to me

segmind/src/segmind/detectors/coarse_event_detector_nodes.py
```diff
@@ -56,11 +101,13 @@ def _find_interaction_events(
         :return: List of newly detected interaction events.
         """
         primary_events = [
-            e for e in segmind_context.logger.get_events()
+            e
```
LucaKro, line 104: no abbreviations

segmind/src/segmind/detectors/coarse_event_detector_nodes.py
```diff
@@ -88,16 +139,34 @@ def _find_interaction_events(
...
+    with no agent in it has to go on.
 
     This class is typically used to analyze specific event types, such as stop
     motion and support events, and identify correlations that form the basis
     of new placing events. By ensuring that placing events are uniquely paired,
     the class helps maintain consistency and prevent duplication of events.
     """
 
-    def update_context_and_events(self, context:MotionStatechartContext, segmind_context:SegmindContext, obj: List[Body]) -> List[DetectionEvent]:
+    requires = (SupportDetector, TranslationDetector)
```
LucaKro, line 154: doesnt it require more? why wouldnt it require LossOfGraspDetector for example

segmind/src/segmind/detectors/spatial_relation_detector_nodes.py
```diff
@@ -78,8 +99,13 @@ class LossOfSupportDetector(AbstractDetector):
     interactions is essential.
     """
 
+    counterpart = SupportDetector
```
LucaKro, line 102: could it make sense to say there isnt a "loss of support" detector and a "support detector", but just a "support detector" that handles both support and loss of support? because it feels like they are more connected than the bases would suggest

segmind/src/segmind/detector_selection.py
```diff
@@ -0,0 +1,83 @@
...
+        if detector_type in chosen:
+            return
+        for required in detector_type.requires:
+            cls._choose(required, chosen)
+        chosen.append(detector_type)
+        for counterpart in cls._counterparts_of(detector_type):
+            cls._choose(counterpart, chosen)
+
+    @staticmethod
+    def _counterparts_of(
```
LucaKro, line 68: the one positive and its counter part are the same class, we dont need this anymore

segmind/src/segmind/live_segmenter.py
```diff
@@ -0,0 +1,191 @@
...
+    SupportDetector,
+)
+from segmind.episode_segmenter import EpisodeSegmenterExecutor
+from segmind.event_logger import EventLogger
+from segmind.statecharts.segmind_statechart import SegmindStatechart
+from segmind.utils import PropagatingThread
+from semantic_digital_twin.world import World
+from semantic_digital_twin.world_description.world_entity import Body
+
+OBJECT_DETECTOR_TYPES: Tuple[Type[AbstractDetector], ...] = (
```
LucaKro, line 45: no global constants, does an enum make sense here?

segmind/src/segmind/live_segmenter.py
```diff
@@ -0,0 +1,191 @@
...
+    def __post_init__(self):
+        super().__post_init__()
+        self.daemon = True
+        self.executor = EpisodeSegmenterExecutor(
+            context=MotionStatechartContext(world=self.world)
+        )
+        self.executor.compile(SegmindStatechart().build_statechart(self.detectors))
+
+    @classmethod
+    def watching(
```
LucaKro, line 119: i dont like these names

semantic_digital_twin/src/semantic_digital_twin/reasoning/predicates.py
```diff
@@ -247,26 +247,65 @@ def compute_euclidean_planar_distance(
     return body1_position.euclidean_distance(body2_position)
 
 
+RESTING_CONTACT_TOLERANCE = 0.005
```
LucaKro, line 250: no global constants

segmind/src/segmind/live_segmenter.py
```diff
@@ -0,0 +1,191 @@
...
+            so asking for what is to be detected is enough. Without any, every kind of
+            :data:`OBJECT_DETECTOR_TYPES` and :data:`EVENT_COMBINING_DETECTOR_TYPES`.
+        """
+        asked_for = detectors or (
+            *OBJECT_DETECTOR_TYPES,
+            *EVENT_COMBINING_DETECTOR_TYPES,
+        )
+        chosen: List[AbstractDetector] = []
+        for detector_type in DetectorSelection.of(*asked_for).detector_types:
+            if detector_type.watches_a_body():
```
LucaKro, line 139: this seems weird...so bodies is only used if ` detector_type.watches_a_body()`?

LucaKro (changes requested):

segmind/src/segmind/detectors/spatial_relation_detector_nodes.py
```diff
@@ -257,12 +276,19 @@ class InsertionDetector(AbstractDetector):
...
     """
     The threshold for the time difference between two events to be considered an insertion.
     """
 
-    def update_context_and_events(self, context:MotionStatechartContext, segmind_context:SegmindContext, tracked_objs: List[Body]) -> List[DetectionEvent]:
+    def update_context_and_events(
+        self,
+        context: MotionStatechartContext,
+        segmind_context: SegmindContext,
+        tracked_objs: List[Body],
```
Narenvasant: no abbreviation please - 'objs'

segmind/src/segmind/live_segmenter.py
```diff
@@ -0,0 +1,191 @@
...
+    TranslationDetector,
+    StopTranslationDetector,
+    RotationDetector,
+    StopRotationDetector,
+)
+"""
+The detectors that each watch a single body.
+"""
+
+EVENT_COMBINING_DETECTOR_TYPES: Tuple[Type[AbstractDetector], ...] = (
```
Narenvasant: here as well, global constants :(

segmind/src/segmind/detectors/spatial_relation_detector_nodes.py
```diff
@@ -278,9 +304,19 @@ def update_context_and_events(self, context:MotionStatechartContext, segmind_con
         :return List of InsertionEvent objects representing detected insertions.
         """
         events = []
-        contact_events = [i for i in segmind_context.logger.get_events() if isinstance(i, ContactEvent)]
-        contact_events_with_holes = [i for i in contact_events if i.with_object in segmind_context.holes]
-        containment_event = [i for i in segmind_context.logger.get_events() if isinstance(i, ContainmentEvent)]
+        contact_events = [
```
Narenvasant: i know this isn't new from this pr, but suggesting could you maybe replace i and j with names for better readability?

test/segmind_test/test_detectors/test_several_watched_bodies.py
```diff
@@ -0,0 +1,127 @@
...
+from ..conftest import RESTING_ON_THE_TABLE
+
+TICKS_WITHOUT_ANYTHING_MOVING = 4
+"""
+How many ticks a test runs while nothing in the world moves.
+"""
+
+
+def _ticked_while_nothing_moves(
+    world, detector_types: List[Type[AbstractDetector]], watched: List[Body]
```
Narenvasant: type hint for world

test/segmind_test/test_detectors/test_pick_ups.py
```diff
@@ -0,0 +1,173 @@
...
+    """
+    :return: A context whose logger holds ``events``.
+    """
+    segmind_context = SegmindContext()
+    for event in events:
+        segmind_context.logger.log_event(event, segmind_context.tracker_registry)
+    return segmind_context
+
+
+def _translation_of(body, at: datetime) -> TranslationEvent:
```
Narenvasant: type hint for body

test/segmind_test/test_event_segmentation.py
```diff
@@ -0,0 +1,90 @@
...
+Seconds a test waits for the watching thread to have detected something.
+"""
+
+MOVED_ALONG_X = 0.2
+"""
+How far a test moves the milk while the run is watched.
+"""
+
+
+def _stand_the_milk_on_the_table(milk) -> None:
```
Narenvasant: same for milk too :)

Narenvasant (changes requested):

segmind/src/segmind/detectors/agent_event_detector_nodes.py
sorinar329: Only the grasp detection or actually all detectors?

segmind/src/segmind/detectors/coarse_event_detector_nodes.py
sorinar329: Not a hard requirment no, because for example if we look at VR we will not detect grasping. But if Grasping is in the statechart then it will depend on it.

segmind/src/segmind/detectors/atomic_event_detectors_nodes.py
sorinar329: yes but by default i exclude the robot parts from the contact detection (currently), a grasp is a specific kind of contact indeed, for this i have to make the contact detected with the robot parts be more robust which currently isn't, will try to do it until your next review :D

coraplex/demos/coraplex_bullet_world_demo/demo.py
sorinar329: the framework from the ppl that told you it will not work in cram?

PR #673, opened 2026-09-23
Let subclasses declare ProbabilisticModel.variables as a field

tomsch420:
First of a series of PRs splitting up #628 (layered probabilistic circuits) into reviewable pieces.

## What

`ProbabilisticModel.variables` was an abstract property. A dataclass subclass could not satisfy it with a plain field: the class stays abstract and cannot be instantiated. This PR declares `variables` as a type-annotated contract instead, so each subclass may either store it as a field or compute it as a property.

- `variables: Tuple[Variable, ...]` replaces the abstract property, with a note in the class docstring.
- `@dataclass` is dropped from `ProbabilisticModel`, which has no fields of its own.
- The existing property-based subclasses (rx, jax, distributions, Bayesian networks, JPT) are unchanged.

## Why

The layered circuit in #628 is a dataclass whose `variables` is a field. This is the only change it needs outside its own package, so it is split out to be reviewed on its own.

## Tests

- `test/probabilistic_model_test`: 428 passed, 1 skipped (the known `UniformLayer` nextafter skip).
- `test/krrood_test/test_eql/test_probabilistic_queries` and `test_underspecified_knowledge` (the krrood users of `ProbabilisticModel`): 43 passed.

Made with AI assistance


LucaKro (approved):

tomsch420:
Remove the test

LucaKro:
why not test the behavior you are using here?

ichumuh (approved):

